PR #10, opened 2021-04-08
Add loader from vscode packages

leiserfg:
You can call `load()` from the added module and that will do the work, I tested it with `rafamadriz/friendly-snippets` and it works fine.


L3MON4D3:
Looks pretty good. Could you add an example for using `load()` to `Examples/snippets.lua`?

L3MON4D3:
Nice, thanks a lot for contributing :)

PR #11, opened 2021-04-09
Improve identation handling

leiserfg:
I checked other implementations and shiftwidth should have higher precedence over tabstop.

### DIFF
--- a/lua/nodes/snippet.lua
+++ b/lua/nodes/snippet.lua
@@ -238,7 +238,7 @@ function Snippet:set_text(node, text)
 
 	self:enter_node(node.indx)
 	if vim.o.expandtab then
-	    local tab_string = string.rep(" ", vim.o.tabstop)
+	    local tab_string = string.rep(" ", vim.o.shiftwidth ~=0 and vim.o.shiftwidth or vim.o.tabstop)
         for i, str in ipairs(text) do
             text[i] = string.gsub(str, "\t", tab_string)
         end
@@ -335,7 +335,7 @@ function Snippet:indent(line)
 	self.indentstr = prefix
 	-- Check once here instead of inside loop.
 	if vim.o.expandtab then
-	    local tab_string = string.rep(" ", vim.o.tabstop)
+	    local tab_string = string.rep(" ", vim.o.shiftwidth ~=0 and vim.o.shiftwidth or vim.o.tabstop)
         for _, node in ipairs(self.nodes) do
             -- put prefix behind newlines.
             if node:has_static_text() then


L3MON4D3:
According to `:h tabstop`, `tabstop` is the ''Number of spaces that a \<Tab> in the file counts for."
Could you go into more detail in your reasoning or give an example where giving `shiftwidth` precedence would make more sense?

leiserfg:
I tested the current state and it was kinda misbehaving, then checked it vsnip and found 
https://github.com/hrsh7th/vim-vsnip/blob/b08d78bb7df1f1d3b279dff182746e454b8080c5/autoload/vsnip/indent.vim#L4-L6

L3MON4D3:
TBH I'd rather follow nvim's documentation on this, srry.
 What exactly causes the misbehaviour? Maybe post a screenshot with `syntax=whitespace` or a snippet and relevant settings?
I'd be open to add a setting for number of whitespaces if all else fails.

leiserfg:
The misbehavior I noticed was when using any snippet involving tabs in a python file. 
I checked the Ultisnips and they do similar (preferring `shiftwidth` over `tabstop`)
If checked the docs and `tabstop` is actually how many  spaces are used to draw a `tab`, and  `shiftwidth` is how many  spaces to use when `expandtab` is on and one hits `tab`.
You can test it easily by doing
```vim
:set listchars=tab:→\ ,trail:• 
:set noexpandtab 
```
Hit tab several times
then
```vim
:set expandtab
:set tabstop=1   
:set  shiftwidth=4 
```

And then hitting `tab`,  for each new tab,  4 spaces will appear, but the old tabs will luck like they take a single space.

L3MON4D3:
Ah okay, thanks for the demonstration.
I think when setting `expandtab` you'd want `tabstop`=`shiftwidth` as per `:h tabstop`, but when this check is done regularly in other plugins, I agree, it probably makes sense doing so aswell.

lua/nodes/snippet.lua
```diff
@@ -335,7 +335,7 @@ function Snippet:indent(line)
 	self.indentstr = prefix
 	-- Check once here instead of inside loop.
 	if vim.o.expandtab then
-	    local tab_string = string.rep(" ", vim.o.tabstop)
+	    local tab_string = string.rep(" ", vim.o.shiftwidth or vim.o.tabstop)
```
L3MON4D3, line 338: This only checks whether `shiftwidth` is set, which it always is (I think).
Maybe replace with `vim.o.shiftwidth == 0`

L3MON4D3: Also check line 240 of this file, there is another check.

L3MON4D3:
That should do the trick, thanks a lot.

PR #13, opened 2021-04-10
Improve loader for vs code

carlitux:
If language is an array process all of them

### DIFF
--- a/lua/loaders/from_vscode.lua
+++ b/lua/loaders/from_vscode.lua
@@ -41,23 +41,26 @@ local function file_read(path)
 end
 
 
-local function load_snippet_file(lang, snippet_set_path)
+local function load_snippet_file(langs, snippet_set_path)
     if not path_exists(snippet_set_path) then return end
-    local lang_snips = ls.snippets[lang] or {}
-
     local snippet_set_data = json_decode(file_read(snippet_set_path))
-    for name, parts in pairs(snippet_set_data) do
-        local body = type(parts.body) == "string" and parts.body or table.concat(parts.body, '\n')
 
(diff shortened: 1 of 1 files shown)


leiserfg (approved):
Thanks, I was planing to fix that today.

leiserfg:
@carlitux  prefixes (aka triggers) can also be repeated.

lua/loaders/from_vscode.lua
```diff
@@ -41,23 +41,26 @@ local function file_read(path)
...
+            local body = type(parts.body) == "string" and parts.body or table.concat(parts.body, '\n')
+
+            -- There are still some snippets that fail while loading
+            pcall(function()
+            table.insert(
+                lang_snips,
+                ls.parser.parse_snippet({trig=parts.prefix, name=name, wordTrig=true}, body)
+            )
+            end) 
+        end
```
L3MON4D3, line 61: First of all, thanks :D
Second, are you certain that wrapping `table.insert` into a function is necessary? Am I missing something? :sweat_smile:

leiserfg: Yes, the parser can fail if the snippet is wrong.

L3MON4D3: Yeah, I agree with the `pcall`, I just think `function() table.insert(...) end` can be simplified into `table.insert(...)`.

leiserfg: pcall needs a function as parameter

L3MON4D3: Ah I see now, my mistake, thought pcall also error-checks the args but it doesn't.

PR #14, opened 2021-04-11
Move the code inside of a luasnip folder

leiserfg:
Neovim way of managing the packages makes it easy to have naming collisions (like having two files named `utils` in two distinct packages), using a folder per plugin is a simple way to bring "namespacing" to it.


L3MON4D3:
Very nice, thank you :+1:

PR #19, opened 2021-05-14
Handle snippets with multiple prefixes

leiserfg:


leiserfg:
I just noticed that because friendly-snippets added some new snippets for markdown and that makes compe + luasnip misbehave.

L3MON4D3:
Looks good, thanks for your contribution :+1:

PR #21, opened 2021-05-24
Move loading logic out of the main loop

leiserfg:
This change improves the startuptime of the plugin when using the loader and `friendly-snippets` around ~65ms
(from ~68 to ~3) on my pc and my laptop.


L3MON4D3:
Damn, very nice, thank you :+1:

PR #23, opened 2021-05-25
Fix small event typo

leiserfg:


L3MON4D3:
Ty!

leiserfg:
That removes the error but still does not trigger the spectated behavior.

leiserfg:
Sorry the comment was about my PR
let me try yours.

PR #24, opened 2021-05-25
Move autogrup to plugins as it has to be loaded asap

leiserfg:


leiserfg:
I see you did it already.

L3MON4D3:
Thanks a lot!

plugin/luasnip.vim
```diff
@@ -9,3 +9,9 @@ snoremap <silent> <Plug>luasnip-jump-next <cmd>lua require'luasnip'.jump(1)<Cr>
 snoremap <silent> <Plug>luasnip-jump-prev <cmd>lua require'luasnip'.jump(-1)<Cr>
 
 command! LuaSnipUnlinkCurrent lua require'luasnip'.unlink_current()
+
+
+augroup luasnip
+	au!
+	autocmd InsertLeavePre * lua ls.active_update_dependents()
```
L3MON4D3, line 16: Shouldn't that be ``require'luasnip'`?

leiserfg: Yes, I moved to my laptop and the version of the code there is old, sorry :facepalm:

L3MON4D3: Ah, no worries :D

L3MON4D3:
👌

plugin/luasnip.vim
leiserfg: I guess it's time to start sharing the plugging somehow (Reddit,  HN)
I was thinking of a couple of live quality helpers that could be added and maybe a bit of improvement on the docs (like adding the dynamic node) but besides that, the plugin is almost done. WDYT?

L3MON4D3: Oh yeah, Luasnip is feature-complete, as far as I'm concerned. Docs need to be improved (added :D), after that I was thinking of posting to the neovim subreddit. As for the qol-improvements, feel free to open an issue to discuss them there, I'd very much like to hear what you have in mind.

leiserfg: I will do some PRs for those as they don't require touching internals.
Mainly something like  `expand_or_jumpable` but for going back (I currently have my own implementation cause I use <c-k> for digraphs sometimes so I can use it for jumping or that).

A helper to declare snippets without having to use `f`,`t` and `i` explicitly, as those are the most common snippets and they can be inferred "structurally" (til now only an idea, not implemented). Still considering if it's worth it.

A keybinding to define/expand small snippets on the fly,  like <https://github.com/abo-abo/auto-yasnippet> (Could also an external plugin, but the implementation should be small enough as to be included here).

L3MON4D3: Okay, sounds good.
I agree with the small parser maybe not being worth it as most cases are/will be (lsp-transformations are missing, but I have an idea) covered by LSP Syntax.
The temp-Snippets-one sounds interesting.

PR #28, opened 2021-06-11
fix: remove ls global usage

tjdevries:


L3MON4D3:
Ah, thank you for correcting this.

PR #31, opened 2021-06-11
feat: Error for users when they don't have i(0)

tjdevries:
Now when users don't put an `i(0)` in their snippet, they get an error on creation (as well as an error that explains what's happening on expansion, but that should not happen anymore)

```
|| E5113: Error while calling lua chunk: .../pack/packer/start/LuaSnip/lua/luasnip/nodes/snippet.lua:63: It is required to have an `i(0)` insert node in your snippet
|| ctx: {
||   trig = "test"
|| }
|| nodes: { {
||     markers = {},
||     static_text = { "mirrored: " },
||     type = 0
||   }, {
||     dependents = {},
||     inner_active = false,
||     markers = {},
||     pos = 1,
||     type = 1
||   }, {
||     markers = {},
||     static_text = { " // " },
||     type = 0
||   }, {
||     args = { 1 },
||     fn = <function 1>,
||     markers = {},
||     type = 2,
||     user_args = {}
||   }, {
||     markers = {},
||     static_text = { " | " },
||     type = 0
||   } }
```

### DIFF
--- a/lua/luasnip/nodes/snippet.lua
+++ b/lua/luasnip/nodes/snippet.lua
@@ -45,6 +45,19 @@ local function S(context, nodes, condition, ...)
 	if type(context) == 'string' then
 		error("Pass table containing a 'trig'-key and optionally 'dscr' and 'name' as first arg.")
 	end
+
+    local insert_nodes = vim.tbl_filter(function(n)
+      -- todo: type should be an enum
+      return n.type == 1 and n.pos == 0
+    end, nodes)
+
+    if #insert_nodes == 0 then
+      error(string.format(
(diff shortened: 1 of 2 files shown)


clason:
An alternative could be to insert a default `i(0)` at the end of the snippet if none is specified explicitly. (That's what most other snippet engines seem to do.)

L3MON4D3:
Yeah, that wouldn't hurt. The `i(0)` can already be omitted in parsed snippets, but checking for it during snippet initialization instead is definitely better.
Would you be okay with changing your PR to instead add an `i(0)`? @tjdevries

leiserfg:
@L3MON4D3  with the automatic `i(0)` it's now obsolete isn't?

L3MON4D3:
Yes! See #49.

PR #36, opened 2021-06-24
Add verbose function names as more readable alternatives

### LINKED ISSUE #34: Add more verbose node alternatives
In `init.lua`, you define

```lua
	s = snip_mod.S,
	sn = snip_mod.SN,
	t = require("luasnip.nodes.textNode").T,
	f = require("luasnip.nodes.functionNode").F,
	i = require("luasnip.nodes.insertNode").I,
	c = require("luasnip.nodes.choiceNode").C,
	d = require("luasnip.nodes.dynamicNode").D,
```
which is nice and terse, but not super user-friendly. I know that it put me off when I first checked out this plugin and I keep needing to look up what they mean every once in a while.

Since most users will alias those functions anyways (e.g. `local s = ls.s`), would you be open to adding more verbose variants as well? `local s = ls.snippet` would be much more readable for most people and barely more verbose in the context of a full config. Of course the current ones would stay, if only for backwards compatibility.

I'd be happy to submit a PR with this and adjusting the documentation if that sounds good to you.

Thanks for LuaSnip btw, it looks like a really solid plugin from what I've seen so far!

cbarrete:
Closes #34.


L3MON4D3:
Looks great, thank you!!

PR #38, opened 2021-06-25
Add stylua action to check lua style automatically

leiserfg:


L3MON4D3:
Nice!
Does the stylua action automatically find `.lua` files in sudirectories?

leiserfg:
Yes.

L3MON4D3:
Great, thank you!

leiserfg:
I think I can extendit to auto-format the code, so if someone makes an unformated pr it will get fixed without intervention.

L3MON4D3:
That would be even better. Is it possible to have the changes as a separate commit?

leiserfg:
Yes, I'll do that.

PR #39, opened 2021-06-25
Autowrap values

cbarrete:
This allows for terser syntax, making curly bracket pairs optional in quite a few places.

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -94,7 +94,7 @@ TEXTNODE                                                       *luasnip-text-nod
 The most simple kind of node; just text.
 >
 	s({trig='trigger'}, {
-		t"Wow! Text!",
+		t("Wow! Text!"),
 		i(0)
 	})
 <
@@ -142,7 +142,7 @@ user-defined function:
 >
 	s({trig = "trig"}, {
 		i(1)
-		f(function(args, user_arg_1) return {args[1][1] .. user_arg_1} end,
+		f(function(args, user_arg_1) return args[1][1] .. user_arg_1 end,
 			{1},
 			"Will be appended to text from i(0)"),
 		i(0)
@@ -162,6 +162,9 @@ The second parameter is a table of indizes of jumpable nodes whose text is
 passed to the function. The table may be empty, in this case the function is
 evaluated once upon snippet-expansion. If the table only has a single node, it
 can be passed directly without wrapping it in a table.
+
+The function shall return a string, which will be inserted as-is, or a table
+of strings for multiline snippets.
 
 Examples:
 	Use captures from the regex-trigger using a functionNode:
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -94,7 +94,7 @@ TEXTNODE                                                       *luasnip-text-nod
 The most simple kind of node; just text.
 >
 	s({trig='trigger'}, {
-		t({"Wow! Text!"}),
+		t"Wow! Text!",
 		i(0)
 	})
 <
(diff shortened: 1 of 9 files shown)


L3MON4D3:
Looks very nice, clearly an improvement over tables only👍

L3MON4D3:
I'll look into your changes in a sec.

doc/luasnip.txt
```diff
@@ -152,14 +153,15 @@ The first parameter of `f` is the function. Its parameters are
...
+evaluated once upon snippet-expansion. If the table only has a single node, it
+can be passed directly without wrapping it in a table.
 
 Examples:
 	Use captures from the regex-trigger using a functionNode:
 >
 	s({trig = "b(%d)", regTrig = true, wordTrig = true}, {
 		f(function(args) return
-			{"Captured Text: " .. args[1].captures[1] .. "."} end, {}),
+			"Captured Text: " .. args[1].captures[1] .. "." end, {}),
```
L3MON4D3, line 164: I think the example in line 138 doesn't need { } either, could you add that?

doc/luasnip.txt
```diff
@@ -87,7 +87,7 @@ TEXTNODE                                                       *luasnip-text-nod
 The most simple kind of node; just text.
 >
 	s({trig='trigger'}, {
-		t({"Wow! Text!"}),
+		t"Wow! Text!",
```
L3MON4D3, line 90: I think this could confuse newcomers to lua, especially if the ( ) are removed in only one example.

doc/luasnip.txt
L3MON4D3: Also, I didn't add anything about the functions return-value, could you add that it can be either string or table?

lua/luasnip/nodes/functionNode.lua
```diff
@@ -30,7 +30,9 @@ function FunctionNode:input_enter()
 end
 
 function FunctionNode:update()
-	self.parent:set_text(self, self.fn(self:get_args(), unpack(self.user_args)))
+	local text = util.wrap_value(
+		self.fn(self:get_args(), unpack(self.user_args))
+	)
```
L3MON4D3, line 35: You lost the `parent:set_text` in your last refactor :D

L3MON4D3:
Apart from those minor issues everything looks great.

doc/luasnip.txt
cbarrete: Kinda? We see `require'foo`' everywhere and I think it's nice to remind people that this is possible, it almost looks like a DSL there.
I'll add it back if you disagree though.

lua/luasnip/nodes/functionNode.lua
cbarrete: Uh oh...

cbarrete:
Cool, I'll address all of that asap, thanks for the review

doc/luasnip.txt
L3MON4D3: It looks nice, but I think to just have it in one example could lead newcomers to think that it is a mistake. Doing it consistently would be best, but that's something I could do as well.

leiserfg:
@L3MON4D3  this looks like one of the TODOS I told you some weeks ago :tada:

L3MON4D3:
> @L3MON4D3 this looks like one of the TODOS I told you some weeks ago 
> But doing this will probably require some minor changes on the stylua config as I guess it does not matches our current configuration.

You refer to omitting `( )`? Right, hadn't thought about that..

leiserfg:
For removing unrequited parentheses no_call_parentheses = true  will do the trick.

L3MON4D3:
Yeah, though Id prefer to keep parantheses generally, so parentheses in lua and the _cooler_, DSL-like syntax in docs.

doc/luasnip.txt
sandersantema: Fyi the formatter used by this project, stylua, will format `require'foo'` as `require('foo')` by default if I'm not mistaken. I suppose it'd be better to keep the formatting consistent between the code itself and the examples especially when trying to understand the plugin better by reading the code and comparing it to the examples.

L3MON4D3:
Looks great, thank you!!

PR #41, opened 2021-06-25
Add options to vscode loader

leiserfg:


L3MON4D3:
Looks very nice code-wise, could you add (to the short example in `snippets.lua`) that no `path` means 'search runtime' and maybe a bit about the ex+include-tables?

leiserfg:
Sure!

L3MON4D3:
Replacing '~' is a good idea👍
Great work, thank you!

Issue #40, opened 2021-06-25
Optional terser snippet syntax


cbarrete:
See https://github.com/L3MON4D3/LuaSnip/pull/39.

On top of that, I'd like to:
1. allow strings as the first argument for snippets, which would be converted to `{ trig = the_string_here }`
2. auto-insert `i(0)` at the end of snippets that don't have any

With all of those merged, a short snippet could look something like:

```lua
s('sh', t'#!/bin/bash')
```

Arguably, we could also auto-convert strings/tables of strings into text nodes and numbers/`{ number, text_node }` into insert nodes, or something like that. It would be a bit more questionable and a quite big addition, but the user-experience/snippet readability would be improved IMO.

What's your opinion on all that? :)

clason:
On a (tenuously) related note, I'd like more builtin pre-defined function nodes for simple, common tasks like `same(1)` (which just mirrors whatever the first placeholder is).

L3MON4D3:
1 and 2 are very nice additions, I wouldn't go far beyond that, as most is already covered by LSP-parsing.
As for the builtin functions, there is the (undocumented😅) `luasnip.util.functions` (which even contains copy/same :D), but it's pretty empty right now.

PR #42, opened 2021-06-25
Allow string contexts in snippet definitions 

cbarrete:
See https://github.com/L3MON4D3/LuaSnip/issues/40.


L3MON4D3:
Looks great, thank you!!

PR #43, opened 2021-06-25
Add stylua auto format action

leiserfg:


leiserfg:
Sorry for the noise but I didn't have another way to check the action.

L3MON4D3:
That's alright.
I'm not that familiar with github actions, if I were to push, say, 5 commits from my local, would the formatter run for each commit or just once?

L3MON4D3:
Also, is a format-commit created even if the style is correct?

leiserfg:
It is executed once per push. I didn't check if the commit happens even no change is made on a lua file, let me do it and we will see.

leiserfg:
Looks like no, it does not crate emtpy commits.

L3MON4D3:
Okay, that's pretty cool I'd say :+1:
TY!

PR #45, opened 2021-06-25
Fix regression caused by non existing paths

### LINKED ISSUE #44: from_vscode: invalid value (nil) at index 1 in table for 'concat'
Updating to latest version I now have this error:
image

Here is my simple config (based off stripped down version of the example) that used to work a few commits ago:
```lua
    local luasnip = require('luasnip')

    luasnip.config.set_config({
        history = true,
        updateevents = 'TextChanged,TextChangedI',
    })

    require('luasnip/loaders/from_vscode').load()
```

I've tried playing with opts and add an include but I think the issue is with handling the path. Sadly I can't help much more than this bug report as my lua and neovim is still pretty limited.

leiserfg:


leiserfg:
Fixes #44

L3MON4D3:
Thanks for acting so quick.
Could you also add the fix to path-splitting?

leiserfg:
Sure, but I guess better to have it working again and then I clean the split without presure, t's 23:21 here :smile:  and I'm sleepy.

L3MON4D3:
Ahh okay, no prob :D

L3MON4D3:
Ahh, I thought the `split()` was necessary for the fix, mb :sweat_smile:

leiserfg:
No, was just the nil, any way I did the change already.

PR #46, opened 2021-06-25
Use vim.split

leiserfg:


L3MON4D3:
Thank you for sacrificing your good night's sleep :D

PR #47, opened 2021-06-25
Show description in snippets loaded from vscode

pstuifzand:
The description of the vscode snippets is not set, even when it is available. This adds the description to those snippets. It uses the name when the description is missing.


leiserfg (approved):

L3MON4D3:
Looks good, thank you!

PR #49, opened 2021-06-27
Insert implicit `i(0)` in snippets that don't have it

cbarrete:


lua/luasnip/nodes/snippet.lua
```diff
@@ -47,6 +47,19 @@ local function S(context, nodes, condition, ...)
...
+		if node.type == 1 and node.pos == 0 then
+			has_i0 = true
+		end
+	end
+
+	if not has_i0 then
+		local i0 = iNode.I(0)
+		nodes[#nodes + 1] = iNode.I(0)
+	end
+
```
L3MON4D3, line 62: I think it is also possible to check for `snip.insert_nodes[0]` after `snip:init_nodes()` and then insert the `iNode.I(0)` in `insert_nodes[0]` and `nodes[#nodes+1]`, which would be a bit cleaner IMO.

L3MON4D3:
Everything else looks fine :+1:

lua/luasnip/nodes/snippet.lua
cbarrete: `snip:init_nodes` is called for all node types, not just snippets, which requires more type-checking, but also this has a narrower scope (it's clearly *just* for snippets here). Why would you find it cleaner in `init_nodes`? There's one more iteration here but in practice it's really short and quick.

Also line 59 should go, sorry about that.

L3MON4D3: Oh, not in `init_nodes`, I meant after the call to `init_nodes`. In there we also step through all nodes and assign them to `snip.insert_nodes`, which means that after it returns, `insert_nodes[0]` contains the `i(0)` (if of exists).

L3MON4D3: We'd then also need to set `i(0).parent` to `self` and `i(0).indx` to 0, but that's better than searching thtough `nodes` two times.

cbarrete: I'm not sure that removing a tight loop over a handful of elements is worth the overhead in complexity/readability, but your call!

lua/luasnip/nodes/snippet.lua
```diff
@@ -76,7 +64,25 @@ local function S(context, nodes, condition, ...)
...
+		i0.parent = snip
+		snip.insert_nodes[0] = i0
+
+		local last = #snip.insert_nodes
+		i0.indx = last + 1
+
+		if last ~= 0 then
+			i0.prev = snip.insert_nodes[last]
+			snip.insert_nodes[last].next = i0
+		end
```
L3MON4D3, line 83: Looks good, that part is actually not necessary, `insert_nodes[0]` is not linked to the other insert_nodes (not really a fitting name anymore, should be eg. jumpables), it is jumped to after exiting the snippet (see lines 195 and 198).

cbarrete: Ah sweet, much better then

L3MON4D3:
Nice! Thanks a lot!

lua/luasnip/nodes/snippet.lua
L3MON4D3: Whoops, the `i0.indx = 0`was wrong, srry about that. `node.indx` actually represents where the node is located, not when it is jumped to :grimacing:

PR #50, opened 2021-06-27
Auto wrap single nodes in snippets

cbarrete:
Last one of those guys :)

### DIFF
--- a/lua/luasnip/nodes/snippet.lua
+++ b/lua/luasnip/nodes/snippet.lua
@@ -44,6 +44,13 @@ local function S(context, nodes, condition, ...)
 			return true
 		end
 	end
+
+	local nodes = nodes
+	if not nodes[1] then
+		-- Then it's a node-table rather than an array of nodes: auto-wrap it
+		nodes = { nodes }
+	end
+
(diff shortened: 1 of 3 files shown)


L3MON4D3:
Ahh, that's a shame :D

lua/luasnip/nodes/snippet.lua
```diff
@@ -44,6 +44,13 @@ local function S(context, nodes, condition, ...)
 			return true
 		end
 	end
+
+	local nodes = nodes
+	if not nodes[1] then
+		-- Then it's a node-table rather than an array of nodes: auto-wrap it
+		nodes = { nodes }
+	end
```
L3MON4D3, line 52: Very minor, but I think checking for `type(nodes) == "table"` is a bit more readable.

cbarrete: That's the point, I can't. Both nodes *and* node lists are tables, so that doesn't help (hence the comment).

L3MON4D3: Ohh yeah, you're right, mb!

PR #55, opened 2021-07-02
Add lazy loading example

leiserfg:


L3MON4D3:
Nice idea, I do wonder though, does `compe` still show all snippets?

Examples/snippets.lua
```diff
@@ -269,3 +269,24 @@ ls.snippets = {
...
+		loader.load({ include = { ft } })
+	end
+end
+
+return vim.cmd([[
+    augroup my_snips
+    autocmd!
+    augroup END
+    au my_snips BufEnter * lua _G.load_snips()
+]])
```
L3MON4D3, line 292: Looks good, minor issues:
- I'd rather have the au inside the augroup it it's the same to you, more readable IMO.
- Is the return necessary?

leiserfg: I copied it from my settings and as I use fennel the last line is always a return, and I didn't notice it :man_facepalming:

leiserfg:
> Nice idea, I do wonder though, does `compe` still show all snippets?

It works for me, I have tested it even with several languages in the same instance.

Examples/snippets.lua
L3MON4D3: Ahh, okay :D

L3MON4D3:
Very cool then, thank you!

PR #58, opened 2021-07-03
Use nvim_get_runtime_file

saadparwaiz1:
Instead of searching through all runtime directories only search directories where a package.json exists. I think it might help with performance still need to run benchmarks to see if it does.

### DIFF
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -155,6 +155,15 @@ function list_to_set(list)
 	return out
 end
 
+local function get_snippets_rtp()
+	return vim.tbl_map(function(itm)
+		return vim.fn.fnamemodify(itm, ":h")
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Sounds good, if performance isn't worse than without and @leiserfg gives his OK I'm happy to merge :D

leiserfg:
Sorry for the delay, It seems good to go, I'm not sure how faster it will be (as anyway, it has to iter over the paths) but it C so, should be faster.

leiserfg:
This will be more useful for the lazy loading as I have to store the paths, so, consider extracting that bit of code out (as a small function) so I can reuse it there.

saadparwaiz1:
It's done!

leiserfg:
Thanks @saadparwaiz1

L3MON4D3:
Nice!

PR #59, opened 2021-07-04
Add lazy loading of vscode snipets

leiserfg:
It can be improved once #58 get merged

### DIFF
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -195,4 +195,45 @@ function M.load(opts)
 	end
 end
 
+local lazy_load_paths = {}
+local lazy_loaded_ft = {}
+
+local function not_contains(tbl)
+	return function(it)
+		return not vim.tbl_contains(tbl, it)
+	end
+end
+
+function _G._luasnip_vscode_lazy_load()
+	for _, ft in ipairs({ vim.bo.filetype, "all" }) do
+		if not lazy_loaded_ft[ft] then
+			lazy_loaded_ft[ft] = true
+			M.load({ paths = lazy_load_paths, include = { ft } })
+		end
+	end
+end
+
+function M.lazy_load(opts)
+	opts = opts or {}
+	-- We have to do this here too, because we have to store them in lozy_load_paths
+	if type(opts.paths) ~= "table" and opts.paths ~= nil then
+		opts.paths = vim.split(opts.paths, ",")
+	else
+		opts.paths = get_snippets_rtp()
+	end
+
+	-- Extend the snippet paths list but avoid doppelgänger
+	vim.list_extend(
+		lazy_load_paths,
+		vim.tbl_filter(not_contains(lazy_load_paths), opts.paths)
+	)
+
+	vim.cmd([[
+		augroup _luasnip_vscode_lazy_load
+		autocmd!
+		au BufEnter * lua _G._luasnip_vscode_lazy_load()
+		augroup END
+	]])
+end
+
 return M
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -183,11 +183,11 @@ function M.load(opts)
 	-- list of paths to crawl for loading (could be a table or a comma-separated-list)
 	if type(opts.paths) ~= "table" and opts.paths ~= nil then
 		opts.paths = vim.split(opts.paths, ",")
+		opts.paths = vim.tbl_map(expand_path, opts.paths) -- Expand before deduping, fake paths will become nil
 	else
 		opts.paths = get_snippets_rtp()
 	end
 
-	opts.paths = vim.tbl_map(expand_path, opts.paths)  -- Expand before deduping, fake paths will become nil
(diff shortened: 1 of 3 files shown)


L3MON4D3:
Looks good so far :+1:

leiserfg:
I noticed that sometimes the snippets get duplicated (seems like the function from @saadparwaiz1  can return the same path twice)  and my function to remove the duplicated is not working well, so I will remove the duplicated on a deeper level. Will update the pr when that gets fixed.

leiserfg:
@L3MON4D3  have a look at it again please

L3MON4D3:
> I noticed that sometimes the snippets get duplicated (seems like the function from @saadparwaiz1 can return the same path twice) and my function to remove the duplicated is not working well, so I will remove the duplicated on a deeper level. Will update the pr when that gets fixed.

 Are you certain that there are duplicate paths in the output of `get_snippets_rtp`? I'm a bit sceptical since each dir can only contain one `package.json`, therefore there shouldn't be duplicate paths?
I think not adding these would be nicer than removing after the fact.

leiserfg:
It can, I don't know why, but any way doing it after the fact, allow us to remove other duplicated entries, like for instance, someone adding by hand a path that is already in the `runtimepath`.

leiserfg:
I checked and the repetition comes from `vim.api.nvim_get_runtime_file()` itself.

L3MON4D3:
Okay weird, wouldn't have thought that.
You're right, it won't hurt to make sure paths are unique.

leiserfg:
> Okay weird, wouldn't have thought that.
> You're right, it won't hurt to make sure paths are unique.

Maybe is that packer is adding the plugin twice to the runtimepath  and `nvim_get_runtime_file` does not dedupe.

lua/luasnip/loaders/from_vscode.lua
```diff
@@ -187,12 +187,42 @@ function M.load(opts)
 		opts.paths = get_snippets_rtp()
 	end
 
+	opts.paths = vim.tbl_map(expand_path, opts.paths) -- Expand before deduping, fake paths will become nil
```
L3MON4D3, line 190: The paths from `get_snippets_rtp` don't need to be expanded anymore, right? In that case, I think this line should be inside the 'if'.

lua/luasnip/loaders/from_vscode.lua
```diff
@@ -187,12 +187,42 @@ function M.load(opts)
...
 
+function M.lazy_load(opts)
+	opts = opts or {}
+
+	-- We have to do this here too, because we have to store them in lozy_load_paths
+	if type(opts.paths) ~= "table" and opts.paths ~= nil then
+		opts.paths = vim.split(opts.paths, ",")
+	else
+		opts.paths = get_snippets_rtp()
+	end
```
L3MON4D3, line 218: The paths should also be filtered here, should they not?

lua/luasnip/loaders/from_vscode.lua
leiserfg: I'm not sure if it always brings the absolute path as `rutimepath` may contain paths with `~`, let me check.

lua/luasnip/loaders/from_vscode.lua
```diff
@@ -187,12 +187,42 @@ function M.load(opts)
...
+	vim.list_extend(lazy_load_paths, opts.paths)
+	vim.cmd([[
+		augroup _luasnip_vscode_lazy_load
+		autocmd!
+		au BufEnter * lua _G._luasnip_vscode_lazy_load()
```
L3MON4D3, line 223: tbh I'd rather not have `lazy_load` in the global namespace, the user probably won't interact with it directly. Maybe put the function in `M` and call here using `require`?

lua/luasnip/loaders/from_vscode.lua
leiserfg: Here is not that important, the only issue could be that the user calls the function too many times (like in a loop) cause that will make the cache of paths grow. But yes, let me at least dedupe them.

lua/luasnip/loaders/from_vscode.lua
leiserfg: Ok, I put it on _G as it's not exposed API, but if you prefer not to clutter the global context,  I will do it.

L3MON4D3: ty!

lua/luasnip/loaders/from_vscode.lua
L3MON4D3: Ah ofc, completely overlooked that `M.load` is called on each path.

L3MON4D3:
Nice, if you add the `M`-part I'm happy to merge :D

leiserfg:
> Nice, if you add the `M._luasnip_vscode_lazy_load`-part I'm happy to merge :D

Sure, I thought you marked that as resolved, let me do it now.

L3MON4D3:
Yeah, I thought you were going to add it in your next commit and marked it as resolved, but that's on me :sweat_smile:

L3MON4D3:
Okay looks great, thanks a lot :D

PR #62, opened 2021-07-07
Fix get active snip

leiserfg:


L3MON4D3:
Looks good, I forgot to change the function after using per-buffer snippets I think.
Thanks!

leiserfg:
I don't use it, I was just looking for how to add the post hook and found that function that is exposed and was broken.

L3MON4D3:
Ah okay, nice.
If you have questions concerning the post-hook, feel free to open it as WIP to discuss.

Issue #66, opened 2021-07-10
Add "repeat" node with optional transform


clason:
One very common task in snippets is to insert the typed input in multiple places (possibly with a transformation), for example replacing whitespace with underscores to add a label or header guard. This can be done with a function or dynamic node, of course, but it's so common that I feel having this built-in would reduce boilerplate and make writing snippets much more convenient.

So my proposal would be to add a new node type "repeat node" that would look like 
```
r(1, {match = "[^.]+$", upper = true, gsub = {"%s+", "_"}})
```
to insert the current value of the input node 1, with everything after the first `.` stripped, converted to upper case, then with whitespace replaced by underscore. (Maybe that's not the best interface; comments welcome :))

L3MON4D3:
Great suggestion, as discussed :D
What do you think of eg. 
```lua
r(1, {"match", "[^.]+$"}, "toupper", {"gsub", "%s+", "_"}, {"gsub", "_", ","})
```
That way we could have multiple transforms (of the same kind) and the order would be clear.

L3MON4D3:
We'd have some more complexity because of the special strings eg. "toupper", but that's acceptable I think.

clason:
That also works, of course. Whatever is easiest to implement ;)

L3MON4D3:
:+1:
Any more suggestions for special strings?

clason:
"tolower", of course ;)

Maybe look at http://lua-users.org/wiki/StringRecipes for inspiration?

L3MON4D3:
Oh, I didn't know about `string.upper/lower`, that's pretty nice :D
rep also sounds usable.

leiserfg:
I use `rep` for my  
```
╔════════════╗
║box snippets║
╚════════════╝
````

L3MON4D3:
Ah, the dreaded unicode-box-snippets :P
Those should also get cleaner, eg `r(1, {"gsub", ".", "═"})`

L3MON4D3:
> But I think that instead of adding more nodes we should provide more functions in `luasnip.util.functions` and document them.

Yeah exactly, that's the plan. There was some discussion in the matrix group, should've mentioned the results here too :sweat_smile:

clason:
That as well, but I _do_ think that adding an actual node for this would be good.

L3MON4D3:
Well we don't really need a new node to implement the expected behaviour as function nodes are very generic. Could you elaborate on what you dislike about just using function nodes?

[...]

PR #67, opened 2021-07-10
Add simpler way to define functions to transform strings

leiserfg:


leiserfg:
This implements #66

leiserfg:
To use it 
`local _1 = require "luasnip.util.lambda"._1`

```
    s({ trig = "upper", wordTrig = true }, {
      i(1, { "content" }),
      t { "*" },
      f( _1:match('[^.]+$'):upper():gsub("%s+", "_"), {1}),
    }),
```

lua/luasnip/nodes/functionNode.lua
```diff
@@ -1,7 +1,14 @@
 local FunctionNode = require("luasnip.nodes.node").Node:new()
 local util = require("luasnip.util.util")
+local lambda = require("luasnip.util.lambda")
 
 local function F(fn, args, ...)
+	if lambda.isPE(fn) then
+		local expr = lambda.instantiate(fn)
+		fn = function(args)
+			return expr(unpack(args[1]))
```
L3MON4D3, line 9: Wouldn't the unpack 'push back' args if the snippet had more lines than one?

L3MON4D3:
Looks good otherwise :+1:

lua/luasnip/nodes/functionNode.lua
leiserfg: If you give me an example, I can test it.

L3MON4D3: Not snippet, srry, node. Testing should work with your snippet, just <Cr> while inside the `i(1)`.

leiserfg: Yes, it works.

L3MON4D3: In what way? Does it return two transformed lines?

leiserfg: No, it keeps formating the first line only, I thought that was the expected behavior.

L3MON4D3: Yeah, that's alright. Why not `args[1][1]` instead of `unpack(args[1])`?

leiserfg: Now I `args` is a list of inputs and each input is a list of `lines`. If that's the case my implementation is incorrect as I'm passing to the function all the lines of the first input, instead of all the inputs. But managing a list of strings in a one-liner will be tricky, is fine if I replace each input with the concatenation of its lines?

L3MON4D3: Yeah, I was going for operating on the first line only, but concatenating and splitting (return from functionNode-function should be a table of strings, each one line) after applying the function should be even more intuitive.

lua/luasnip/nodes/functionNode.lua
```diff
@@ -1,7 +1,19 @@
...
 local function F(fn, args, ...)
+	if lambda.isPE(fn) then
+		local function _concat(tbl)
+			return table.concat(tbl, "\n")
+		end
+		local expr = lambda.instantiate(fn)
+		fn = function(args)
+			local inputs = vim.tbl_map(_concat, args)
+			local out = expr(unpack(inputs))
+			return vim.split(out, "\n")
```
L3MON4D3, line 14: Nice, but wouldn't a more concise `return vim.split(expr(table.concat(args[1], "\n")), "\n")` also do it? These function nodes depend on only one other node, therefore args would always be `{{"line 1", "maybe line 2"}}`.

leiserfg: It can be more than one node,  as in   `_1 .. _2` , or   `_1:gsub(_2)`

leiserfg: Anyway, I had to stop for today, but tomorrow morning I will move it to a function out of the function_node as we talk before.

L3MON4D3: Oh wow, that's even better, I didn't realize the lambdas were this powerful :0
Sounds good, thank you!!

leiserfg:
After the last changes one can do 
local l =  require'luasnip.extras'.l
```lua
 s({ trig = "upper", wordTrig = true }, {
      i(1, { "content" }),
      t { "*" },
      f( l._1:match('[^.]+$'):upper():gsub("%s+", "_"), {1}),
    }),
 ```
  and also use
  `r(1)`   passing a second argument to r (also a lambda) will work as well

File: lua/luasnip/extras/init.lua [hunk 3]
```diff
@@ -0,0 +1,50 @@
...
+	return x[1]
+end
+
+local function rep(input_id, tx)
+	if not tx then
+		return F(_idem, { input_id })
+	end
+	tx = expr_to_fn(tx)
+	return F(tx, { input_id })
+end
```
L3MON4D3, line 41: This is rather elegant, but I wouldn't include it here, as repeating a node is pretty common and doesn't really require lambdas. The benefit of adding repeat here is (only?) that transforms can be applied, but if transforms are necessary, a simple repeat wouldn't do it either way and lambda would be the way to go.
I'd just put the function I outlined in #66 into `util.functions` and be done with it, but there's no need to do that in this PR :)

leiserfg: Feel free to pull it and reshape it as you want.

L3MON4D3: Okay, or like that :D

L3MON4D3:
Thanks a lot btw, this is really nice :heart:

PR #83, opened 2021-07-24
Docs: Specify the jump order for InsertNodes

praveendhawan:
The doc was a bit misleading as it suggested that the jump nodes
are in decreasing order from `n to 0` with 0 being the exit node.

This change explicitly specifies that the order is in fact increasing
from `1 to n` with 0 being the exit node.

Also, add missing documentation that it is not possible to define
placeholder text for the `0-th` InsertNode.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -98,14 +98,34 @@ with an example:
 
 ```lua
 s("trigger", {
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Nice, thank you!

Issue #71, opened 2021-07-12
Add conditional text node


clason:
Continuing in my quest for more "convenience nodes" to make writing snippets in LuaSnip's powerful syntax more convenient and concise, I'd like to suggest adding a "conditional text" that add text if and only if a placeholder is non-empty, so you could do something like
```
t"\usepackage", cond("[",2), i(2), cond("]",2), t"{", i(1), t"}"
```
which would expand to `\usepackage[bar]{foo}` in general but `\usepackage{bar}` if `$2` was left empty.

(One could generalize this with an arbitrary regex that is matched against the insert node, but this might be overkill?)

L3MON4D3:
Ouh, regex/pattern sounds good! More general and all that :D
Yeah, I'll definitely add that :+1:

clason:
With default "empty" so I don't have to remember that regex all the time :)

L3MON4D3:
`match_insert(insert, pattern, invert)` with pattern default "^$", invert (insert if the pattern doesn't match) default false?

clason:
So in my example, I'd have `m(2,"[")`?

L3MON4D3:
> (Your signature is missing the node on which to match.)

Whoops, `match_insert(node_indx, insert, pattern, invert)` it is.
> So in my example, I'd have `m(2,"[")`?
> 
Yeah, that'd be it.

clason:
Perfect.

L3MON4D3:
I'm thinking, perhaps lambdas would be a good fit for this function. It would look a bit more convoluted (`m(_1:match("^$"), 2, "[")`, insert if the lambda does/not return nil), but we could add a simple match as a special case so we'd still cover simple cases. That would enable us to also do transforms beforehand and compensate a bit for the rather simple patterns of lua ie. by applying `match()` more than once.

clason:
Again, my goal is to make the 90%-case as easy and concise as possible to write without having to remember regexes.

I know it's a bit more work (and, from what I gather from these discussions, doesn't appeal to you aesthetically), but you could think of adding _another_ layer on top of these for the simplest, most common case (kinda like you did for `rep` on top of `lambda`).

clason:
Can lambda nodes be nested? So I could have conditional text that is generated from another insert node? Or is that something that is starting to strain the mould and should better be done using function or dynamic nodes?

L3MON4D3:
[...]

PR #84, opened 2021-07-24
Add extra match function

leiserfg:
Implements the complex version of #71

### DIFF
--- a/lua/luasnip/extras/init.lua
+++ b/lua/luasnip/extras/init.lua
@@ -28,9 +28,50 @@ end
 
 setmetatable(lambda, LM)
 
+local function to_function(val, use_re)
+	if type(val) == "function" then
+		return val
+	end
+	if type(val) == "string" and not use_re then
+		return function()
+			return val
+		end
+	end
+	if type(val) == "string" and use_re then
+		return function(arg)
+			return arg:match(val)
+		end
+	end
+	if lambda.isPE(val) then
+		return lambda.instantiate(val)
+	end
+	assert(false, "Can't convert argument to function")
+end
+
+local function match(index, _then, _match, _else)
+	assert(_then, "You have to pass at least 2 arguments")
+	_then = to_function(_then)
+	_match = to_function(_match or "^$", true)
+	_else = to_function(_else or "")
+
+	local function func(arg)
+		local text = _concat(arg[1])
+		local out = nil
+		if _match(text) then
+			out = _then(text)
+		else
+			out = _else(text)
+		end
+		return vim.split(out, "\n")
+	end
+
+	return F(func, { index })
+end
+
 return {
 	lambda = lambda,
-
+	match = match,
 	--alias
 	l = lambda,
+	m = match,
 }
--- a/lua/luasnip/extras/init.lua
+++ b/lua/luasnip/extras/init.lua
@@ -50,6 +50,8 @@ end
 
 local function match(index, _then, _match, _else)
 	assert(_then, "You have to pass at least 2 arguments")
+	assert(type(index) == "number", "Index has to be a single number")
+
 	_then = to_function(_then)
 	_match = to_function(_match or "^$", true)
 	_else = to_function(_else or "")


lua/luasnip/extras/init.lua
```diff
@@ -28,9 +28,46 @@ end
...
+	assert(false, "Can't convert argument to function")
+end
+
+local function match(index, _then, _match, _else)
+	assert(_then, "You have to pass at least 2 arguments")
+	_then = to_function(_then)
+	_match = to_function(_match or "^$", true)
+	_else = to_function(_else or "")
+	local function func(arg)
+		if _match(arg[1][1]) then
```
L3MON4D3, line 57: Allowing `match` to also be a lambda is a great idea, but then it'd be even better to pass the argument-nodes' text the same way as in the original lambda nodes (eg. '\n'-concat text from each node and pass these strings to `_match`), that way multiple nodes could be used in `_match`(shouldn't interrupt usage with simple patterns, I think).
I just realized that that should probably be done for `_then` and `_else` as well, to keep lambda-usage more uniform.

leiserfg: Sure, I was a bit bussy today.

leiserfg: I did it for a single node, is not clear to me how to do that for several nodes, like, should we do if all the matches work? Or a single match? I guess that will add complexity to the api and at that point better to use a normal  function node IMHO.

L3MON4D3: I was thinking of only matching on the first node, but you're right, there would be some ambiguity about what 'true' means if several nodes are given in `index` (eg. only the first node matches, or all nodes match).
On the other hand, adding it would enable us to insert text when eg. two nodes have the same text (`m({1,2}, "equal!!", l._1:match(_2))`), which could be nice.

Upon thinking, that wouldn't be common at all, I can't think of a single snippet where it would make sense to have such functionality, nevermind that.

lua/luasnip/extras/init.lua
```diff
@@ -28,9 +28,50 @@ end
...
+	_match = to_function(_match or "^$", true)
+	_else = to_function(_else or "")
+
+	local function func(arg)
+		local text = _concat(arg[1])
+		local out = nil
+		if _match(text) then
+			out = _then(arg)
+		else
+			out = _else(arg)
```
L3MON4D3, line 63: Those should also get `text`, shouldn't they?

lua/luasnip/extras/init.lua
```diff
@@ -28,9 +28,50 @@ end
...
+			return vim.split(out, "\n")
```
L3MON4D3, line 64: `ddp<<` :P

lua/luasnip/extras/init.lua
L3MON4D3: Actually, in that case, could you add an `assert()` to make sure that the user is actually passing a single index and not a table?

L3MON4D3: I feel like many of these functions accept a table as indices and we should probably be more explicit about the type of the parameters (I'll add those to some other functions too, as soon as I have some time).

L3MON4D3:
Wonderful, thanks a lot!!

L3MON4D3:
Okay, really quick, I just went to write an example for the match snippet, and the current signature (I suggested :roll_eyes:) is waaayyy to unnatural, you're still alright with `match(index, _match, _then, _else)` instead of `match(index, _then, _match, _else)`?

leiserfg:
Yes, I did it that way as you are the "product owner" but it's unnatural for me to read it too :smile:

L3MON4D3:
Okay, perfect xD
My reasoning at that time was that `match` has a fairly reasonable default value (eg. empty) whereas `_then` doesn't, but that isn't worth sacrificing additional readability for.

leiserfg:
I can do it as If you pass only 2 args then it's considered like match is `"^$"`   and  the  arg is `_then` wdyt?

L3MON4D3:
I think switching the order of arguments to accomodate for that 1 in 1000 case where somebody actually uses an "^$" for pattern would be weird, it's fine like it is now (but thanks for the suggestion :D).
If someone excessively uses "^$", they can still alias `match()` to cover that case without specifying the regex.

leiserfg:
I mean, to cover the original issue of @clason.

L3MON4D3:
Ah, that's where you're coming from.
That should be covered as soon as I add the `nonempty`-node.

leiserfg:
In that case which should be the default values?  cause having `m(1, "a pattern")` without values does not make much sence I guess.

L3MON4D3:
Yeah, that'd be the best candidate _if_ we would like any default.
If we only supported patterns I'd have said to use the first matching group (or the entire matched pattern), but with the generalization to functions, that wouldn't work really.
Or we specify that, if the function returns any non-nil value, that will be inserted, which would be in line with `match()` and it's return value.

PR #86, opened 2021-07-24
Change args order in match

leiserfg:


lua/luasnip/extras/init.lua
```diff
@@ -48,12 +48,16 @@ local function to_function(val, use_re)
 	assert(false, "Can't convert argument to function")
 end
 
+local function _idem(x)
+	return x
```
L3MON4D3, line 52: How about instead doing `return _match(x) or ""`, that would allow to return the entire matched text or the first capture group by default, which would be quite intuitive?

L3MON4D3: (`_match` from after `to_function`)

leiserfg: That will be problematic if someone does something like 
`m(1,  tonumber)`

L3MON4D3: true, I assumed the lambdas we get as `_match` would always end with a `match()`.
Your case is problematic, but what would the expected result be when putting `tonumber` as the match?
(we could also test for type==string to protect against that)

leiserfg: I used isnumber as an example, the idea is that any function could be used there and that make sence as one could want to use a custom function to check some property of the text, like being a number, or other more complex that can't be expressed in a pattern.

L3MON4D3: Ah, yeah the function-as-pattern is very nice to have.
But even in that case, using `return (type(_match(x))=="string" and _match(x)) or ""` wouldn't cause any problems, true, the matched number would have to be inserted using `_then`(or with `m(1, "^%d+$")`), but in the vast majority(I'd think) of cases (simple regex), that would be very useful, eg `m(1, ".?.?.?")` to return up to the first three characters.

L3MON4D3: More complex stuff is bound to be implemented by the user, where they can return a string(which will be inserted if no `_then` is given) on match or `nil`

leiserfg: I'm far from home today if you wanna have it soon feel free to merge and fix it, or I'll do it later.

L3MON4D3: Okay, sounds good:+1:

PR #92, opened 2021-08-01
Add TM_SELECTED_TEXT variable

leiserfg:


lua/luasnip/util/util.lua
```diff
@@ -151,6 +151,51 @@ local function wrap_value(value)
...
+		)
+	end
+	vim.api.nvim_buf_set_var(0, LUASNIP_LAST_SELECTION, chunks)
+end
+
+local function get_selection()
+	local ok, val = pcall(vim.api.nvim_buf_get_var, 0, LUASNIP_LAST_SELECTION)
+	return ok and val or ""
+end
+
```
L3MON4D3, line 198: Very nice, but wouldn't eg. mapping `<tab>` to `s` and storing the unnamed register do the same?

leiserfg: This approach allows having a distinct selection per buffer, I borrowed it mostly from Ultisnips.

L3MON4D3: Per-buffer is great, and that part I wouldn't change, but I think that the function that gets the text could be simplified by deleting the text(maybe not with `s`, `d` would probably be better as we'd stay in normal) and then storing the unnamed register into a buffer-local variable.

L3MON4D3: (I thought ultisnips does it exactly that way tbh)

leiserfg: No, ultisnips do almost what I did, the only difference is that it does not use a buffer variable to store the text.
https://github.com/SirVer/ultisnips/blob/8803960a4e09daf1ad4d82e16a25bbdb4c78530e/pythonx/UltiSnips/vim_state.py#L120

https://github.com/SirVer/ultisnips/blob/e96733b5db27b48943db86dd8623f1497b860bc6/autoload/UltiSnips/map_keys.vim#L64

leiserfg: Using a buffer vars frees me from having to clean the memorized value myself.

leiserfg: Anyway, I will try to replicate the current behavior but using    `s<cmd>...`  and fetching the text from the register, which should be similar to your proposal and will remove the manual vim dance.

L3MON4D3: Thank you! :D
(I do wonder though, why does ultisnips do the 'vim dance'?)

leiserfg: Done!, took a bit as I was having lunch.

L3MON4D3: Sweet, thank you!

L3MON4D3:
I'm not sure about the behaviour, what do you think about setting `b:LUASNIP_LAST_SELECTION` back to "" after getting the stored variable?
Like this, there would always be some text from a previous selection in the var which would be more confusing than only being able to use selected text once.

leiserfg:
Well, it's not what is specified (there is no specification for the expected behavior I guess) but  $VISUAL in Ultisnippets and  TM_SELECTED_TEXT in TextMate (where the TM_ came from) keep the last selection, til there is a new one.

L3MON4D3:
Oh really?
Well, in that case this looks good :+1: 
Thank you!!

PR #93, opened 2021-08-02
Back to the dance floor to avoid overriding the unnamed register. 

leiserfg:


lua/luasnip/config.lua
```diff
@@ -2,6 +2,7 @@ local defaults = {
 	history = false,
 	updateevents = "InsertLeave",
 	store_selection_keys = nil, -- Supossed to be the same as the expand shortcut
+	store_selection_register = "l", --name of the register to use for capturing selections
```
L3MON4D3, line 5: Good idea to make it configurable, I think 9 would be a saner default.

lua/luasnip/util/util.lua
```diff
@@ -185,7 +185,8 @@ local function get_selection()
 end
 
 local function store_selection()
-	local chunks = vim.split(vim.fn.getreg('"'), "\n")
+	local c = require("luasnip.config").config
+	local chunks = vim.fn.getreg(c.store_selection_register, 1, true)
```
L3MON4D3, line 189: I did some quick testing, and it seems like this way, the unnamed register will still contain the visual. Setting it here with `setreg()` and the content of `"1` should make it work as expected.

lua/luasnip/config.lua
leiserfg: I prefer named registers, as numbered ones get populated automatically.

lua/luasnip/util/util.lua
leiserfg: I tested it myself and I don't see it cluttering `"`.

leiserfg: I mean, i did  `"ls`  by hand and it does not overrides `"`

L3MON4D3: Oh, I tested it by just yanking text and then yanking some other text into `"l`, where it did overwrite unnamed.
But if it works, nvm that.

lua/luasnip/config.lua
L3MON4D3: Tbh here I'd just go after what is least likely to erase text needed by the user and I think it's safe to assume that one is more likely to use/need eg. `l` than the text from 9 yanks/deletions/... ago

lua/luasnip/util/util.lua
L3MON4D3: Okay, I tested it on your branch, this seems to still overwrite the unnamed register :/

lua/luasnip/config.lua
leiserfg: That's why I made it configurable  :smile: . Using registers in plugins is always risky as there is always a possibility of someone using the same register for something else.

lua/luasnip/util/util.lua
L3MON4D3: `vim.fn.setreg('"', vim.fn.getreg("0"))` does the trick ;)

leiserfg: If it overrides the register then the fix is a bit more complex, as that way you are recovering the content of the register, but not the type of it.

lua/luasnip/config.lua
L3MON4D3: Oh yeah, configurable is a must, no question, I'm just arguing that `9` is less likely to be actively used than any of the character-registers and should therefore be the default instead of `l`.

lua/luasnip/util/util.lua
L3MON4D3: Yeah, you're right, I looked into the docs, even `0` only works because I yanked the text.

leiserfg: Was looking into this and there is not how to properly reset the unnamed register, as it points to the last named/numbered used register so setting, recovering from `"0` only works if the last action was a yank, if it was a delete then the content won't be there, but in `"1` or `"-`.  And now that I think about it, probably using the ultisnips dance won't fix it either as they do `s` too, so they override the unamed register the same way.

L3MON4D3: We'd have to go back to the manual vim-dance and delete visual using nvim_buf_set_text(), that would leave registers alone.

L3MON4D3: Another option would be to do it on the TextYankPost autocommand, with that there would be no expectation that the register remains untouched.

leiserfg: So, I think there is not a proper way to fix this issue without overengineering (wrapping all the `s` operations to store all the registers and set them again after). So I think if someone wants to copy and paste in a selection snippet, they better use `"0p`, `"1` or `"-` accordingly (yes, that's painful)

L3MON4D3: > We'd have to go back to the manual vim-dance and delete visual using `nvim_buf_set_text()`, that would leave registers alone.

That should work though, the only annoying part would be the complicated manual copying of the selection.
(it would also be possible to set the cursor as it is with `s`)

leiserfg: One option could be to do the vim-dance and use  "_s  as that null register does not update the unnamed one.

L3MON4D3: Ah yeah, that's better :+1:

L3MON4D3:
Great👌👌

Issue #99, opened 2021-08-06
bug: nvim -u mini.lua => MYCONFIG_ROOT=nil


ayoubelmhamdi:
i can use require('luasnip.loaders.from_vscode').load() if i open nvim with -u
```
:require('luasnip.loaders.from_vscode').load(

E5108: Error executing lua ...packer/start/LuaSnip/lua/luasnip/loaders/from_vscode.lua:168: attemp
t to index field 'MYVIMRC' (a nil value)
```
i change line 168 ==> 
from
```
local MYCONFIG_ROOT = vim.env.MYVIMRC:gsub("/[^/]+$", "")

```
to
```
local MYCONFIG_ROOT = vim.fn.expand('$USER') ..'/.config/nvim'
```
and bug gone

L3MON4D3:
See #74, `$MYVIMRC` is supposed to be set, always. You could try installing a different version of neovim, ie. build from source. Another way to circumvent the error is exporting `MYVIMRC` in your `.bashrc`.
Just seeing it to `.config/...` sadly won't do it, users may have it in some other dir and we'd like to still have Luasnip work as expected.

leiserfg:
I see, when using `-u afile`  (n)vim does not sets `$MYVIMRC`. Crap, I'll have to deal with that.

ayoubelmhamdi:
it's work now

Issue #74, opened 2021-07-13
$MYVIMRC nil value


danielnehrig:
apparently $MYVIMRC is not set which throws at
LuaSnip/lua/luasnip/loaders/from_vscode.lua:168

:h $MYVIMRC
```
						*VIMINIT* *EXINIT* *$MYVIMRC*
     b. Locations searched for initializations, in order of preference:
	-  $VIMINIT environment variable (Ex command line).
	-  User |config|: $XDG_CONFIG_HOME/nvim/init.vim.
	-  Other config: {dir}/nvim/init.vim where {dir} is any directory
	   in $XDG_CONFIG_DIRS.
	-  $EXINIT environment variable (Ex command line).
	|$MYVIMRC| is set to the first valid location unless it was already
	set or when using $VIMINIT.

```

also i try to lazy load like this

```lua
require("luasnip/loaders/from_vscode").lazy_load(
    {paths = {"~/.local/share/nvim/site/pack/packer/start/friendly-snippets"}}
)
```

i use init.lua as cfg

nvim version
```
NVIM v0.5.0-dev+1375-gca802046b
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions     -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security     -fstack-clash-protection -fcf-protection -Wp,-U_FORTIFY_SOURCE -Wp,-D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/dnehrig/.cache/yay/neovim-git/src/build/config -I/home/dnehrig/.cache/yay/neovim-git/src/neovim-git/src -I/usr/include -I/home/dnehrig/.cache/yay/neovim-git/src/build/src/nvim/auto -I/home/dnehrig/.cache/yay/neovim-git/src/build/include
[...]

PR #102, opened 2021-08-06
Find the vimrc path in case of calling nvim with -u

leiserfg:


leiserfg:
This solves #99 and probably #74

L3MON4D3:
Looks solid :+1:
Not sure about #74, that looked more like something went wrong with packer or neovim.

PR #104, opened 2021-08-09
Correct typo: namr → name

amenonsen:
I see that doc/luasnip.txt is autogenerated from DOC.md (though I'm not sure how), so I didn't patch that file.


L3MON4D3:
TY!

PR #114, opened 2021-08-15
Make most of the variable lazy evaluated and memoized so we can implement more without a big runtime cost

leiserfg:


leiserfg:
Snippet to test the implementation

 ```
    ls.parser.parse_snippet({trig='foo'},
  [[
  ${$TM_SELECTED_TEXT} --  TM_SELECTED_TEXT The currently selected text or the empty string
  ${$TM_CURRENT_LINE} --  TM_CURRENT_LINE The contents of the current line
  ${$TM_CURRENT_WORD} --  TM_CURRENT_WORD The contents of the word under cursor or the empty string
  ${$TM_LINE_INDEX} --  TM_LINE_INDEX The zero-index based line number
  ${$TM_LINE_NUMBER} --  TM_LINE_NUMBER The one-index based line number
  ${$TM_FILENAME} --  TM_FILENAME The filename of the current document
  ${$TM_FILENAME_BASE} --  TM_FILENAME_BASE The filename of the current document without its extensions
  ${$TM_DIRECTORY} --  TM_DIRECTORY The directory of the current document
  ${$TM_FILEPATH} --  TM_FILEPATH The full file path of the current document
  ${$RELATIVE_FILEPATH} --  RELATIVE_FILEPATH The relative (to the opened workspace or folder) file path of the current document
  ${$CLIPBOARD} --  CLIPBOARD The contents of your clipboard
  ${$WORKSPACE_NAME} --  WORKSPACE_NAME The name of the opened workspace or folder
  ${$WORKSPACE_FOLDER} --  WORKSPACE_FOLDER The path of the opened workspace or folder
  ${$CURRENT_YEAR} --  CURRENT_YEAR The current year
  ${$CURRENT_YEAR_SHORT} --  CURRENT_YEAR_SHORT The current year's last two digits
  ${$CURRENT_MONTH} --  CURRENT_MONTH The month as two digits (example '02')
  ${$CURRENT_MONTH_NAME} --  CURRENT_MONTH_NAME The full name of the month (example 'July')
  ${$CURRENT_MONTH_NAME_SHORT} --  CURRENT_MONTH_NAME_SHORT The short name of the month (example 'Jul')
  ${$CURRENT_DATE} --  CURRENT_DATE The day of the month
  ${$CURRENT_DAY_NAME} --  CURRENT_DAY_NAME The name of day (example 'Monday')
  ${$CURRENT_DAY_NAME_SHORT} --  CURRENT_DAY_NAME_SHORT The short name of the day (example 'Mon')
  ${$CURRENT_HOUR} --  CURRENT_HOUR The current hour in 24-hour clock format
  ${$CURRENT_MINUTE} --  CURRENT_MINUTE The current minute
  ${$CURRENT_SECOND} --  CURRENT_SECOND The current second
  ${$CURRENT_SECONDS_UNIX} --  CURRENT_SECONDS_UNIX The number of seconds since the Unix epoch
  ${$RANDOM} --  RANDOM 6 random Base-10 digits
  ${$RANDOM_HEX} --  RANDOM_HEX 6 random Base-16 digits
  ${$UUID} --  UUID A Version 4 UUID
  ${$BLOCK_COMMENT_START} --  BLOCK_COMMENT_START Example output: in PHP /* or in HTML <!--
  ${$BLOCK_COMMENT_END} --  BLOCK_COMMENT_END Example output: in PHP */ or in HTML -->
  ${$LINE_COMMENT} --  LINE_COMMENT Example output: in PHP //
    ]]),
```

**Comment by leiserfg** [comment]
Only these ones are not yet implemented as they require knowing the current workspace which is something from vscode, nvim-lsp has workspaces (in plural) so maybe that's what we should use but that will need some kind of heuristic to know which one of those is.

  ${$RELATIVE_FILEPATH} --  RELATIVE_FILEPATH The relative (to the opened workspace or folder) file path of the current document

  ${$WORKSPACE_NAME} --  WORKSPACE_NAME The name of the opened workspace or folder
  ${$WORKSPACE_FOLDER} --  WORKSPACE_FOLDER The path of the opened workspace or folder

**Comment by leiserfg** [comment]
This closes #113

File: lua/luasnip/util/util.lua [hunk 1]
```diff
@@ -210,16 +210,13 @@ local SELECT_RAW = "LUASNIP_SELECT_RAW"
...
 	-- if one is set, all are set.
 	if ok then
-		return val,
-			vim.api.nvim_buf_get_var(0, SELECT_DEDENT),
-			vim.api.nvim_buf_get_var(0, TM_SELECT)
+		return val
 	end
-	-- not ok.
-	return {}, {}, {}
+	return {}
```
L3MON4D3, line 219: This should (it doesn't on `master`, but this is an okay place to fix it) reset the *SELECT*-vars so the selection is only available to the current snippet. That would also require making the SELECT-vars eager, it's improbable that somebody might use two types of selection, but it should work as documented.

L3MON4D3:
> Only these ones are not yet implemented as they require knowing the current workspace which is something from vscode, nvim-lsp has workspaces (in plural) so maybe that's what we should use but that will need some kind of heuristic to know which one of those is.
> 
> ${$RELATIVE_FILEPATH} -- RELATIVE_FILEPATH The relative (to the opened workspace or folder) file path of the current document
> 
> ${$WORKSPACE_NAME} -- WORKSPACE_NAME The name of the opened workspace or folder
> ${$WORKSPACE_FOLDER} -- WORKSPACE_FOLDER The path of the opened workspace or folder

I'd assume that the 'highest' directory is used, maybe check what vscode does?

lua/luasnip/util/util.lua
leiserfg: Do you mean, to clean up the buf-vars once they are fetched?

L3MON4D3: Yes, exactly

leiserfg:
> I'd assume that the 'highest' directory is used, maybe check what vscode does?

Vscode has a concept of  workspace (that can be named and so) nvim only have maybe lsp workspace (that does not have a name) I can try to use that or the folder that is the parent of the current file but in any case, it won't match 100% what vscode does.

lua/luasnip/util/environ.lua
```diff
@@ -19,6 +21,10 @@ local function _fill_eager_vars(env)
 	env.TM_CURRENT_WORD = util.word_under_cursor(cur, env.TM_CURRENT_LINE)
 	env.TM_LINE_INDEX = tostring(cur[1])
 	env.TM_LINE_NUMBER = tostring(cur[1] + 1)
+
+  env.SELECT_RAW = util.get_selection(util.SELECT_RAW)
+  env.SELECT_DEDENT = util.get_selection(util.SELECT_DEDENT)
+  env.TM_SELECTED_TEXT= util.get_selection(util.TM_SELECT)
```
L3MON4D3: What do you think about keeping the function more like the older one eg. get and delete all selected at the same time? I don't think we need 3 calls for that.

leiserfg: Sure, I'll do it, let me first finish with the worspaces thingy.

L3MON4D3: Ah, nice :+1:

leiserfg: I tried to do it the old way but doing  `env.var1, env.var2, env.var3 = get_selection()` breaks with the lazy loading, so either I keep it like this or I'll have to store the result in a table before passing it to each eager variable one by one.

L3MON4D3: This is also fine, but I don't quite get what breaks by returning three values there?

lua/luasnip/util/environ.lua
```diff
@@ -0,0 +1,176 @@
...
+			if file_path:find(ws, 1, true) == 1 then
+				ws_parts = { ws, file_path:sub(#ws + 2, -1) }
+				break
+			end
+		end
+		-- If it can't be extracted from lsp, then we use the file path
+		if not ok and not ws_parts then
+			ws_parts = { vim.fn.expand("%:p:h"), vim.fn.expand("%:p:t") }
+		end
+		vim.api.nvim_buf_set_var(0, LSP_WORSKPACE_PARTS, ws_parts)
```
L3MON4D3, line 82: I think the current directory would make more sense here.

lua/luasnip/util/environ.lua
leiserfg: It breaks as if `env.var1, env.var2, env.var3`  calls  `__index` and then as the vars are not set yet, it fails.

lua/luasnip/util/environ.lua
leiserfg: the file can be out of the `cwd`

lua/luasnip/util/environ.lua
L3MON4D3: Oh okay, that I can believe.

lua/luasnip/util/environ.lua
L3MON4D3: I don't think that's too bad, wouldn't it be possible to get the relative path using `fnamemodify`?

leiserfg: Yes, we can use `fnamemodify` but the file is supposed to be inside of the workspace path as far as I get from the vscode docs, which does not have to be the case with `cwd`.

L3MON4D3: Oh, I didn't look at the docs, if there are some requirements it's alright if we keep it compatible.

lua/luasnip/util/util.lua
```diff
@@ -212,14 +212,20 @@ local TM_SELECT = "LUASNIP_TM_SELECT"
...
+
+		vim.api.nvim_buf_del_var(0, SELECT_RAW)
+		vim.api.nvim_buf_del_var(0, SELECT_DEDENT)
+		vim.api.nvim_buf_del_var(0, TM_SELECT)
+
+		return result
 	end
-	-- not ok.
-	return {}, {}, {}
+	return { "", "", "" }
```
L3MON4D3, line 228: I think the default was an empty table, apart from that this lgtm👌

leiserfg:
Found why the issue, I was not unpacking.

L3MON4D3:
Ha, sweet!

PR #122, opened 2021-08-19
lua table as definition is confusing for non lua developper

ayoubelmhamdi:


ayoubelmhamdi (commented):
update

L3MON4D3:
For your first point, I can see how it could seem weird to users who don't know lua, but I'd prefer maybe linking eg. [this line](https://github.com/L3MON4D3/LuaSnip/blob/master/Examples/snippets.lua#L156) to show how to do it instead of removing the sentence.
The '`:luafile` `Examples/snippets.lua`' was intentional, just `:luafile Examples/snippets.lua` could lead users to just run that command wherever they are right now and get an error because the file may not be in their `cwd`.

L3MON4D3:
I'm going to add the link, thank you for mentioning it :)

PR #126, opened 2021-08-23
Fix loader by using the same package name as imported in the AU

leiserfg:


L3MON4D3:
That would still misbehave in the (admittedly contrived, but still imaginable) scenario that both the `/`- and the `.`-version are called, wouldn't it?

leiserfg:
I'm just avoiding cluttering the `luasnip` namespace. If you think that's the only way...

L3MON4D3:
Yeah, that's a bit of a problem, we'll have to clean up there a bit, collect all nodes in a table or something.
Imo this doesn't really make it much worse, so I'd be all for it.

L3MON4D3:
Maybe put both into a table and that into `luasnip`, keep them together.

L3MON4D3:
Jup, that sounds good :+1:

leiserfg:
Like this.

L3MON4D3:
Nice, that looks good.
Thank you!

PR #138, opened 2021-08-28
Handle empty nodes

### LINKED ISSUE #137: Parser fails with TM_SELECTED_TEXT
Example snippet:
```
      "\\section{${1:section name}} % (fold)",
      "\\label{sec:${2:${1/(\\w+)(\\W+$)?|\\W+/${1:?${1:/asciify/downcase}:_}/g}}}",
      "${0:$TM_SELECTED_TEXT}",
      "% section $2 (end)"
```
Expected behavior: I should be able to complete the snippet as usual. Even if `TM_SELECTED_TEXT` isn't implemented yet, the parser should at least let me finish the snippet at `$0` regardless of if there is selected text or not. 

Actual behavior: The parser immediately throws the error ﻿﻿`E5108: Error executing Lua [...]/snippets.lua:605: Attempt to concatenate a nil value.` without letting me fill in any values. It instead inserts the following expansion after the error:
```
\subsection{subsection name} % (fold)
\label{sub:subsection name}

0% subsection  (end)
```

leiserfg:
Fixes #137

### DIFF
--- a/lua/luasnip/util/functions.lua
+++ b/lua/luasnip/util/functions.lua
@@ -2,7 +2,12 @@ return {
 	var = function(_, node, text)
 		local v = node.parent.env[text]
 		if type(v) == "table" then
-			return v
+			-- Avoid issues with empty vars
+			if #v == 0 then
+				return { "" }
+			else
+				return {}
+			end
 		else
 			return { v }
 		end


L3MON4D3:
Tbh I'd rather fix this by checking for an empty table [here](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/util/functions.lua#L4), in `get_static_text()` we can pretty safely assume that the text is actually a nonempty table of strings, perhaps we should even just change the defaults for selection to `{""}` instead of `{}`.

L3MON4D3:
Nice, thank you :D

PR #140, opened 2021-08-29
Ensure no loading files if lang do not exists

carlitux:
without explicit check, it is passing as an empty table and load unused files


leiserfg:
My python showing up

L3MON4D3:
Nice, thank you!

PR #148, opened 2021-09-01
Fix readme code sample syntax highlighting

bronzehedwick:
The code snippets are using both markdown styles that render as code: angle quotes and indentation. This causes the entire snippet, including the angle quotes, to be rendered as code, but without syntax highlighting. The indentation seems accidental, used to nest properly inside the `<details>` element.

To fix, remove the indentation.


L3MON4D3:
Oh, whoops😅
Thank you for fixing it :D

bronzehedwick:
Not a problem 😄

PR #152, opened 2021-09-03
fix mode of <Plug>luasnip-expand-snippet keymap to select

hituzi-no-sippo:
The next two lines are the same.

## A

https://github.com/L3MON4D3/LuaSnip/blob/1db821efe7416483a7288ef86dbae6f63ab0221e/plugin/luasnip.vim#L2

## B

https://github.com/L3MON4D3/LuaSnip/blob/1db821efe7416483a7288ef86dbae6f63ab0221e/plugin/luasnip.vim#L12

Isn't the mode of the keymap of B a selection?


L3MON4D3:
Mhmmmm, I think I just copied the I-mappings for select and then didn't change expand because "Who would like to expand a snippet in select mode?".
Anyway, it won't hurt to change it, thank you! :D

PR #153, opened 2021-09-05
Allow use of env-variables in (dynamic)LambdaNode

L3MON4D3:
Should allow usage as eg.
```lua
s("transform", {
	l(l.TM_SELECTED_TEXT:gsub("a", "e"), {}),
}),
```
@leiserfg What do you think of the implementation? I'm still not that sure around lambdas, so no idea if this is the best way to do it.

### DIFF
--- a/lua/luasnip/extras/_lambda.lua
+++ b/lua/luasnip/extras/_lambda.lua
@@ -68,8 +68,12 @@ end
 setmetatable(lambda, {
 	__index = function(_, key)
 		-- \\n to be correctly interpreted in `load()`.
-		return P({op="X", repr = "table.concat(snip.env."..key..", \"\\n\")", index = 0})
-	end
+		return P({
+			op = "X",
+			repr = "table.concat(snip.env." .. key .. ', "\\n")',
+			index = 0,
+		})
+	end,
 })
 
 local repr
--- a/lua/luasnip/extras/_lambda.lua
+++ b/lua/luasnip/extras/_lambda.lua
@@ -64,6 +64,14 @@ function lambda._(value)
 	return P({ op = "X", repr = value, index = "wrap" })
 end
 
+-- unknown keys are some named variable.
+setmetatable(lambda, {
+	__index = function(_, key)
+		-- \\n to be correctly interpreted in `load()`.
+		return P({op="X", repr = "table.concat(snip.env."..key..", \"\\n\")", index = 0})
+	end
+})
+
 local repr
 
 lambda.Nil = lambda.Var("nil")
@@ -334,6 +342,8 @@ function lambda.instantiate(e)
 	for i = 1, n do
 		append(parms, "_" .. i)
 	end
+	append(parms, "snip")
+
 	consts = concat(consts, ",")
 	parms = concat(parms, ",")
 	rep = repr(e)
--- a/lua/luasnip/extras/_lambda.lua
+++ b/lua/luasnip/extras/_lambda.lua
@@ -76,6 +76,16 @@ setmetatable(lambda, {
 	end,
 })
 
+lambda.captures = {}
+setmetatable(lambda.captures, {
+	__index = function(_, key)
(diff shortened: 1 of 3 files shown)


leiserfg:
Why not an  `l.env` var that gives access to the snippet environment instead so we can add other vars without fearing collisions later?

L3MON4D3:
Tbh only because I was planning to expose captures in a similar fashion eg. `l.CAPTURE1` (`l.captures[1]` or similar wouldn't work if I understand correctly) and then having `env` in its own table would've been weird.
What kind of collisions are you fearing? All-caps should be pretty safe.

L3MON4D3:
Also, it's a bit shorter, but that's not really a good argument against some namespacing.

leiserfg:
> Tbh only because I was planning to expose captures in a similar fashion eg. `l.CAPTURE1` (`l.captures[1]` or similar wouldn't work if I understand correctly) and then having `env` in its own table would've been weird.
> What kind of collisions are you fearing? All-caps should be pretty safe.

Well, that's true, giving that the idea to type less, having then as uppercase should be good enough.

L3MON4D3:
> Tbh only because I was planning to expose captures in a similar fashion eg. `l.CAPTURE1` (`l.captures[1]` or similar wouldn't work if I understand correctly) and then having `env` in its own table would've been weird.
> What kind of collisions are you fearing? All-caps should be pretty safe.

Actually nvm that, we can also expose `captures` with `__index` and return the appropriate values there.

L3MON4D3:
I don't have a big preference between `l.env.VARNAME` and `l.VARNAME`, I feel like `l.env` contains a bit of duplicate info (uppercase somewhat indicates that it is a env-variable), but it would also be a bit closer to the actual data in the snippet (`snip.env.VARNAME`, `snip.captures[1]`).
`l.VARNAME` is a bit shorter to type, maybe that's actually not as little an advantage as the lambdaNodes are for comfort, so sticking with that for now.

The actual implementation seems alright to you?

leiserfg:
I would prefer not to touch `_lambda.lua` so it has no knowledge of the rest of the code. But I couldn't come with a way to pass the `snip` to the generated function without having to do so.

L3MON4D3:
Very true.
We could probably change it so that even `_1, _2,...` don't directly refer to parameters of the function, but to a table passed to it. That would allow us to set the __index of that table to search the snippet, which could be done outside of _lambda.
Downside would be that we lose the ability to do `l.captures[n]`

L3MON4D3:
(missclick)

L3MON4D3:
With that we diverge a bit from the original and manual usage of these lambdas would be more cumbersome, but that's alright.
Afaict it didn't break anything+we have more separation.

leiserfg:
As far as it does not break old snippets... all good!

leiserfg:
By the way, I extended the snippet parser from archseer  so it does not break with regexes containing escapes, the code is  here
https://github.com/leiserfg/LuaSnip/blob/parser_combinator/lua/luasnip/loaders/parser.lua
I have not done a PR as it will require also creating the snippet from the AST, so feel free to use it or I will try to do it myself later.

L3MON4D3:
> By the way, I extended the snippet parser from archseer so it does not break with regexes containing escapes, the code is here
> https://github.com/leiserfg/LuaSnip/blob/parser_combinator/lua/luasnip/loaders/parser.lua
> I have not done a PR as it will require also creating the snippet from the AST, so feel free to use it or I will try to do it myself later.

Ohhh, sick. Best shoot the guys at nvim a message on the [vim.lsp.util.parse_snippet-issue](https://github.com/neovim/neovim/issues/15522), perhaps they will still want to use a modified version of archseers' parser if someone already wrote it :D
There has so far not been any concrete discussion about the AST (although the generated trees probably wouldn't differ by too much), right? It would be really nice to use your implementation now and then switch to nvims' later on.

PR #165, opened 2021-09-12
Add cleanup function

leiserfg:


L3MON4D3:
Interesting, why'd you use an autocommand for implementation?

leiserfg:
The main idea was not to call the clean from the loaders (well the only one we have til now) from the main module, it also makes it possible for the users to subscribe to the same module. Anyway if you think this adds complexity it can be rewritten in another manner.

L3MON4D3:
Okay, that makes sense.
If we want to add more functions like `clean`, registering loaders upon usage and calling the functions from `luasnip` would be nicer I think, but I can't think of any other functions like `clean`.
TY!

lua/luasnip/loaders/from_vscode.lua
```diff
@@ -131,7 +131,7 @@ local function load_snippet_folder(root, opts)
 	)
 end
 
-function filter_list(list, exclude, include)
+local function filter_list(list, exclude, include)
```
akinsho, line 134: This change is currently causing errors because the position of this function is too low in this file as it is used above this line (before it is defined).  This error is visible now, since (thankfully) it is no longer global @leiserfg or @L3MON4D3

leiserfg: Sorry for the Verschlimmbesserung

PR #166, opened 2021-09-12
fix: define function before calling it

ChristianChiarulli:


L3MON4D3:
Whoops😬 
Thanks for the quick fix 👍

PR #169, opened 2021-09-15
Add printf-like snippet formatting 

jedrzejboczar:
I recently discovered LuaSnip and it is really cool. One thing I was missing was to be able to combine the flexibility of snippets defined in Lua with some kind of string interpolation. Lsp-style snippets allow for interpolation but they lack more advanced features. This PR is an attempt to add an alternative method for defining snippets using format strings similar to the Python ones. It's easiest to show as an example:
```lua
s("echoe", fmt('echo {} 1>&2', i(1, 'msg')),
s("fmtlines", fmt([[
	foo({1}, {3}) {
		# ignore {{}} these {{braces}}
		return {2} * {4}
	}
	]], {
	i(1, "x"),
	r(1),
	i(2, "y"),
	r(2),
})),
```
`fmt` parses the format string, splits it into text nodes and inserts nodes from the second argument (`nodes`) in place of placeholders. Placeholders can be numbers (indexing `nodes` with numbers), strings (then indexes `nodes` as key-value map) or can be empty, then uses last number+1.

I also added unit tests that make use of plenary.nvim's busted. You can run these with `tests/run.sh`.

### DIFF
--- a/lua/luasnip/extras/fmt.lua
+++ b/lua/luasnip/extras/fmt.lua
@@ -1,0 +1,203 @@
+local text_node = require("luasnip.nodes.textNode").T
+local wrap_nodes = require('luasnip.util.util').wrap_nodes
+
+-- Interpolate elements from `args` into format string with placeholders.
+--
+-- The placeholder syntax for selecting from `args` is similar to fmtlib and
+-- Python's .format(), with some notable differences:
+-- * no format options (like `{:.2f}`)
+-- * 1-based indexing
+-- * numbered/auto-numbered placeholders can be mixed; numbered ones set the
+--   current index to new value, so following auto-numbered placeholders start
+--   counting from the new value (e.g. `{} {3} {}` is `{1} {3} {4}`)
+--
+-- Arguments:
+--   fmt: string with placeholders
+--   args: table with list-like and/or map-like keys
+--   opts:
+--     delimiters: string, 2 distinct characters (left, right), default "{}"
+--     strict: boolean, set to false to allow for unused `args`, default true
+-- Returns: a list of strings and elements of `args` inserted into placeholders
+local function interpolate(fmt, args, opts)
+    local defaults = {
+        delimiters = '{}',
+        strict = true,
+    }
+    opts = vim.tbl_extend('force', defaults, opts or {})
+
+    -- sanitize delimiters
+    assert(#opts.delimiters == 2,
+        'Currently only single-char delimiters are supported, e.g. delimiters="{}" (left, right)')
+    assert(opts.delimiters:sub(1, 1) ~= opts.delimiters:sub(2, 2),
+        "Delimiters must be two _different_ characters")
+    local delimiters = {
+        left = opts.delimiters:sub(1, 1),
+        right = opts.delimiters:sub(2, 2),
+        esc_left = vim.pesc(opts.delimiters:sub(1, 1)),
+        esc_right = vim.pesc(opts.delimiters:sub(2, 2)),
+    }
+
+    -- manage insertion of text/args
+    local elements = {}
+    local last_index = 0
+    local used_keys = {}
+
+    local add_text = function(text)
+        if #text > 0 then
+            table.insert(elements, text)
+        end
+    end
+    local add_arg = function(placeholder)
+        local num = tonumber(placeholder)
+        local key
+        if num then  -- numbered placeholder
+            last_index = num
+            key = last_index
+        elseif placeholder == '' then  -- auto-numbered placeholder
+            key = last_index + 1
+            last_index = key
+        else  -- named placeholder
+            key = placeholder
+        end
+        assert(args[key], string.format('Missing key `%s` in format arguments: `%s`', key, fmt))
+        table.insert(elements, args[key])
+        used_keys[key] = true
+    end
+
+    -- iterate keeping a range from previous match, e.g. (not in_placeholder vs in_placeholder)
+    -- "Sample {2} string {3}."   OR  "Sample {2} string {3}."
+    --       left^--------^right  OR      left^-^right
+    local pattern = string.format('[%s%s]', delimiters.esc_left, delimiters.esc_right)
+    local in_placeholder = false
+    local left = 0
+
+    while true do
+        local right = fmt:find(pattern, left + 1)
+        -- if not found, add the remaining part of string and finish
+        if right == nil then
+            assert(not in_placeholder, string.format('Missing closing delimiter: "%s"', fmt:sub(left)))
+            add_text(fmt:sub(left + 1))
+            break
+        end
+        -- check if the delimiters are escaped
+        local delim = fmt:sub(right, right)
+        local next_char = fmt:sub(right + 1, right + 1)
+        if not in_placeholder and delim == next_char then
+            -- add the previous part of the string with a single delimiter
+            add_text(fmt:sub(left + 1, right))
+            -- and jump over the second one
+            left = right + 1
+            -- "continue"
+        else  -- non-escaped delimiter
+            assert(delim == (in_placeholder and delimiters.right or delimiters.left),
+                string.format('Found unescaped %s %s placeholder; format[%d:%d]="%s"',
+                    delim, in_placeholder and 'inside' or 'outside', left, right, fmt:sub(left, right)))
+            -- add arg/text depending on current state
+            local add = in_placeholder and add_arg or add_text
+            add(fmt:sub(left + 1, right - 1))
+            -- update state
+            left = right
+            in_placeholder = delim == delimiters.left
+        end
+    end
+
+    -- sanity check: all arguments were used
+    if opts.strict then
+        for key, _ in pairs(args) do
+            assert(used_keys[key], string.format("Unused argument: args[%s]", key))
+        end
+    end
+
+    return elements
+end
+
+-- Find the largest common indent in a list of lines and remove it.
+local function remove_common_indent(lines)
+    local max_indent = math.huge
+    for _, line in ipairs(lines) do
+        -- ignore lines with whitespace only
+        local match = line:match('^(%s*)%S')
+        if match then
+            max_indent = math.min(max_indent, #match)
+        end
+    end
+    if max_indent > 0 then
+        lines = vim.tbl_map(function(line)
+            return line:sub(max_indent + 1)
+        end, lines)
+    end
+    return lines
+end
+
+-- Use a format string with placeholders to interpolate nodes.
+--
+-- See `interpolate` documentation for details on the format.
+--
+-- Arguments:
+--   str: format string
+--   nodes: snippet node or list of nodes
+--   opts: optional table
+--     trim_empty: boolean, remove whitespace-only first/last lines, default true
+--     dedent: boolean, remove all common indent in `str`, default true
+--     ... the rest is passed to `interpolate`
+-- Returns: list of snippet nodes
+local function format_nodes(str, nodes, opts)
+    local defaults = {
+        trim_empty = true,
+        dedent = true,
+    }
+    opts = vim.tbl_extend('force', defaults, opts or {})
+
+    -- allow to pass a single node
+    nodes = wrap_nodes(nodes)
+
+    -- optimization: avoid splitting multiple times
+    local lines = nil
+
+    -- this allows to use [[...]] strings ignoring the first and last lines
+    if opts.trim_empty then
+        lines = vim.split(str, '\n', true)
+        if lines[1]:match('^%s*$') then
+            table.remove(lines, 1)
+        end
+        if lines[#lines]:match('^%s*$') then
+            table.remove(lines)
+        end
+    end
+
+    -- remove common indent
+    if opts.dedent then
+        lines = lines or vim.split(str, '\n', true)
+        lines = remove_common_indent(lines)
+    end
+
+    if lines then
+        str = table.concat(lines, '\n')
+    end
+
+    -- pop format_nodes's opts
+    for key, _ in ipairs(defaults) do
+        opts[key] = nil
+    end
+
+    local parts = interpolate(str, nodes, opts)
+    return vim.tbl_map(function(part)
+        -- wrap strings in text nodes
+        if type(part) == 'string' then
+            return text_node(vim.split(part, '\n', true))
+        else
+            return part
+        end
+    end, parts)
+end
+
+return {
+    interpolate = interpolate,
+    format_nodes = format_nodes,
+    -- alias
+    fmt = format_nodes,
+    fmta = function(str, nodes, opts)  -- fmt with angle brackets
+        opts = vim.tbl_extend('force', opts or {}, { delimiters = '<>' })
+        return format_nodes(str, nodes, opts)
+    end,
+}
--- a/tests/fmt_spec.lua
+++ b/tests/fmt_spec.lua
@@ -1,0 +1,108 @@
+local interpolate = require('luasnip.extras.fmt').interpolate
+
+local works = function(msg, fmt, args, expected, opts)
+    it(msg, function()
+        local result = interpolate(fmt, args, opts)
+        assert.are.same(expected, table.concat(result))
+    end)
+end
+
+local fails = function(msg, fmt, args, opts)
+    it(msg, function()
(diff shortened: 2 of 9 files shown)


bew:
Would it be possible to reuse the LSP snippet syntax? (which can already be parsed, and supports _some_ nodes like simple choices)

jedrzejboczar:
I'm not sure how would it work. Lsp syntax generally encodes type of node in the string. Here all the info about nodes is defined in the second argument to `fmt`, and format string just allows to define which node should go in which place of the string. Maybe you could give some examples of what you mean?

L3MON4D3:
Wow, that's really nice, especially so because it tackles one of the weakest points of Luasnip, which is defining snippets.
I'll look over it asap.
I wouldn't integrate do fmt via LSP-parsing because
- parsing will soon be done by neovim, we'd only construct snippets from the AST
- the parsing code is pretty horrible, stretching it even further won't be fun

L3MON4D3:
Also good on you for adding some unit tests, I honestly should've sone so long ago :grimacing:

lua/luasnip/extras/fmt.lua
```diff
@@ -0,0 +1,159 @@
...
+        if ignored then
+            -- add the previous part of the string
+            add_text(fmt:sub(prev_end + 1, start - 1))
+            -- add what is inside {{}}
+            add_text(fmt:sub(start + 1, end_ - 1))
+            -- move previous end
+            prev_end = end_
+        else
+            -- add the previous part of the string
+            add_text(fmt:sub(prev_end + 1, start - 1))
```
L3MON4D3, line 58: The two `add_text()` could be moved out of the `if`, right?

lua/luasnip/extras/fmt.lua
```diff
@@ -0,0 +1,159 @@
...
+            prev_end = end_
+            -- interpret the placeholder
+            local placeholder = fmt:sub(start + 1, end_ - 1)
+            local is_num = placeholder:match('^%d+$')
```
L3MON4D3, line 62: I think `tonumber(placeholder)` and a check for `nil` should work as well, what do you think of that? (I can think of floats as an example for where the behaviour would differ, but I _think_ that's negligible)

lua/luasnip/extras/fmt.lua
```diff
@@ -0,0 +1,159 @@
...
+-- Returns: list of snippet nodes
+local function format_nodes(str, nodes, opts)
+    opts = opts or {}
+    opts.trim_empty = opts.trim_empty ~= nil and opts.trim_empty or true
+    opts.dedent = opts.dedent ~= nil and opts.dedent or true
+
+    -- allow to pass a single node
+    if type(nodes.is_node) == 'function' and nodes.is_node() then
+        nodes = {nodes}
+    end
```
L3MON4D3, line 118: There's already a function similar to that [here](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/nodes/snippet.lua#L82-L90), I think it would be nice to factor `wrap_nodes` out to util and use it in both cases.

lua/luasnip/extras/fmt.lua
```diff
@@ -0,0 +1,159 @@
...
+
+    if lines then
+        str = table.concat(lines, '\n')
+    end
+
+    local parts = replace_placeholders(str, nodes)
+    return vim.tbl_map(function(part)
+        -- wrap strings in text nodes
+        if type(part) == 'string' then
+            return ls.text_node(vim.split(part, '\n', true))
```
L3MON4D3, line 148: Afaict luasnip is only `require`d to get access to `textNodes`, right? It'd be nicer to directly `require("luasnip.nodes.textNode")`

lua/luasnip/extras/fmt.lua
```diff
@@ -0,0 +1,159 @@
...
+--   of last numbered placeholder + 1 is used
+-- To input regular curly braces use `{{}}` or `{{text}}`. The text between braces
+-- is not matched by [0-9a-zA-Z_-]* then single braces can be used, no need to escape.
+--
+-- Arguments:
+--   fmt: string
+--   args: table
+-- Returns: a list of strings and elements of `args` inserted into placeholders.
+local function replace_placeholders(fmt, args)
+    local pattern = '{[0-9a-zA-Z_-]*}'
```
L3MON4D3, line 19: Does this work for eg. `{{ } }}`? What do you think of separately escaping opening and closing brackets via `\` or `%`?

tests/fmt_spec.lua
```diff
@@ -0,0 +1,46 @@
...
+        'a{}b{3}c{}d', {1, 2, 3, 4},
+        'a1b3c4d')
+
+    check('expands named placeholders',
+        'a{A}b{B}c{C}d', {A = 1, B = 2, C = 3},
+        'a1b2c3d')
+
+    check('expands all mixed',
+        'a {A} b {} c {3} d {} e {B} f {A} g', {1, 2, 3, 4, A = 10, B = 20},
+        'a 10 b 1 c 3 d 4 e 20 f 10 g')
```
bew, line 29: What happens / should happen for `'a {A} b {} c', {1, 2, B = 20, A = 10, C = 30}` ?

tests/fmt_spec.lua
```diff
@@ -0,0 +1,46 @@
...
+        '{}abcd{}', {1, 2},
+        '1abcd2')
+
+    check('ignores {{}}',
+        'a{{}}b{}c{{}}d{}e', {2, 4},
+        'a{}b2c{}d4e')
+
+    check('ignores {{text}}',
+        'a{{d}}b{}c', {2},
+        'a{d}b2c')
```
bew, line 41: Why is it not `a{{d}}b2c` ?

How would you write a snippet to output: `foo {{bar}} baz` ? (using a lot of nested braces? using `\` to escape brances?)

L3MON4D3: The double brackets are to escape single ones, eg `{{.*}}` - > `{.*}`, for literal double brackets you'd need `{{{{text}}}}`, but I'm not sure if that'll work right now.

tests/fmt_spec.lua
jedrzejboczar: This gives `a 10 b 1 c`, which is also what would happen in Python (`'a {A} b {} c'.format(1, 2, B = 20, A = 10, C = 30)` gives `'a 10 b 1 c'`). 

In current implementation this is because the numbered (`{2}`) and auto-numbered (`{}`) fields use an index that starts from 1 and is incremented on each auto-numbered placeholder or set to new value on each numbered one. Named placeholders ignore this index. What [fmtlib](https://fmt.dev/latest/syntax.html) describes and what you see in Python, is that you cannot mix auto-numbered and numbered fields, so the current implementation might be confusing on this part, e.g. in Python both `'{1} {}'` and `'{} {1}'` are errors, while this implementation treats them as `'{1} {2}'` and `'{1} {1}'` respectively.

tests/fmt_spec.lua
jedrzejboczar: The syntax is based on [fmtlib](https://fmt.dev/latest/syntax.html) and [Python formatting](https://www.python.org/dev/peps/pep-3101/). I checked that it currently works only for the case of `{{...}}` -> `{...}`. Currently, for more braces it gives incorrect results. I would need to do some tests to modify the algorithm if we want to handle such cases.

tests/fmt_spec.lua
L3MON4D3: Makes sense to me, maybe add another example that explicitly shows this behaviour? Especially as it differs from pythons' `fmt`

lua/luasnip/extras/fmt.lua
jedrzejboczar: Yes, I'll change that.

lua/luasnip/extras/fmt.lua
jedrzejboczar: Currently floats should not be an issue as the allowed pattern is `'{[0-9a-zA-Z_-]*}'` so I'll just replace the `match` with `tonumber`.

lua/luasnip/extras/fmt.lua
jedrzejboczar: Sure

tests/fmt_spec.lua
L3MON4D3: Maybe doing `%b{}` for balanced brackets is enough?

L3MON4D3: Oh, and reading through pythons fmt, it seems like singular `{` can also be escaped with `{{`, which would make more sense to me then only being able to escape pairs.

lua/luasnip/extras/fmt.lua
jedrzejboczar: Done

tests/fmt_spec.lua
jedrzejboczar: I added an example and a test for that.

lua/luasnip/extras/fmt.lua
jedrzejboczar: Actually currently this would ignore everything in these examples, as the pattern does not contain spaces. I think that we don't really need spaces in placeholders and this reduces the number of escaping the user has to do in most scenarios (like all the `{ int x = 1; }` scopes in C-like languages.

jedrzejboczar: And maybe it's enough to use single braces, as in practice it should rarely happen that a snippet contains braces that are not a placeholder and contain a text without any spaces inside. What do you think?

tests/fmt_spec.lua
jedrzejboczar: Didn't know about the `%b{}` pattern. But it doesn't seem to address the issue with escaping and we would need some extra logic to manually detect if this should be escaped. `%b{}` for a string like `a {{ b }} c` matches the whole `{{ b }}`, so we would need to run it recursively for that ` b ` part. 

Also, as the pattern currently doesn't accept spaces between braces so this should greatly reduce the number of scenarios in which the user has to do the escaping. It might be a bit confusing for someone that expects full compatibility with fmtlib/Python formatting but I feel like it might be more practical in this particular case.

lua/luasnip/extras/fmt.lua
L3MON4D3: Ah right, bad example :sweat_smile:
Mhmm, I think we should in any case provide a way to properly escape `{` and `}`, (in eg. latex `\begin{something}` is commonplace).
In general, I'd prefer always having to escape brackets (if they should be in the text literally) to having some rules on when it's necessary and when it isn't, but that's just my preference, maybe being smart about it is nicer for most users.

tests/fmt_spec.lua
L3MON4D3: Oh, I thought the problem was that it matches
```txt
{{{{text}}}}
^       ^
```
these two together, nvm then.
Not sure if using a pattern to find matching brackets is the best approach(I don't know if it's possible to ignore escaped brackets with it). 
What would probaly work is building a table containing opening brackets as keys and closing ones as the respective values (eg. `aaa{b{e}c}dd` -> `{4 = 10, 6 = 8}`) (see [here](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/util/parser.lua#L24-L48), rules for escaping would need to be adjusted, not entirely sure if factoring out is worth it) and then splitting the text on these values.

jedrzejboczar:
Ok, I modified the algorithm to search for delimiters linearly. I also did some more changes and squashed everything:
* Now bracket always have to be escaped if they are not to be interpreted as a placeholder.
* Because of this it is in theory possible to have weird things like `fmt('{ test $$$ }', { [' test $$$ '] = 'new' }`, though it's rather a side-effect.
* Because code snippets often contain `{`/`}`, it is now possible to use different delimiters. `fmta` is a simple wrapper that uses `<>` instead of `{}`, but user can use `fmt(_, _, { delimiters = ... })` to choose different ones (must be 2 different characters).
* Now all `args` must be used or an error occurs. This can be disabled with `{ strict = false }`.
* Renamed `replace_placeholders` -> `interpolate`
* Added more tests and examples.

jedrzejboczar:
I'm not sure what's wrong with CI.

lua/luasnip/extras/fmt.lua
jedrzejboczar: Now brackets must always be escaped

L3MON4D3:
The changes sound good👌
CI breaking could be caused by stylua flip-flopping between two different formattings, at least that was the reason the last time. You could try running stylua locally and check whether it changes stuff upon being run a second time.

lua/luasnip/extras/fmt.lua
```diff
@@ -0,0 +1,203 @@
+local text_node = require("luasnip.nodes.textNode").T
+local wrap_nodes = require('luasnip.util.util').wrap_nodes
+
+
+-- Escape Lua magic chars in a string for use in a Lua pattern set ("[]")
+local function escape_set_char(c)
+    return (c == ']' or c == '%') and ('%'..c) or c
+end
```
L3MON4D3, line 8: I think we could use the builtin `vim.pesc` instead

lua/luasnip/extras/fmt.lua
```diff
@@ -0,0 +1,203 @@
...
+
+    -- iterate keeping a range from previous match, e.g. (not in_placeholder vs in_placeholder)
+    -- "Sample {2} string {3}."   OR  "Sample {2} string {3}."
+    --       left^--------^right  OR      left^-^right
+    local pattern = string.format('[%s%s]', delimiters.esc_left, delimiters.esc_right)
+    local in_placeholder = false
+    local left = 0
+
+    while true do
+        local right = fmt:find(pattern, left + 1)
```
L3MON4D3, line 81: Could use `find` with `plain` set, then we wouldn't need the whole escape-business at all, right?

L3MON4D3, line 81: What do you think of doing a plain-text search for 
- `delim_left`, if in_placeholder and
- `delim_right`, if not in_placeholder ?
That shouldn't be slower and be a bit simpler (no need to construct a set from escaped delimiters)

L3MON4D3: That would also require adjusting the un-escape code, ie. only skipping escaped matches and actually replacing eg. `{{` with `{` in `add_text`

lua/luasnip/extras/init.lua
```diff
@@ -150,8 +151,14 @@ return {
...
+	format_nodes = format_nodes,
 
 	--alias
 	l = lambda,
 	m = match,
+	fmt = format_nodes,
+	fmta = function(str, nodes, opts)  -- fmt with angle brackets
+		opts = vim.tbl_extend('force', opts or {}, { delimiters = '<>' })
+		return format_nodes(str, nodes, opts)
+	end,
```
L3MON4D3, line 163: Actually, could you leave it as just a submodule for `extras`? (just sticking with `luasnip.extras.fmt` is alright imo) (I'm eventually going to split `luasnip.extras` up a bit, will be one less breaking change then :D)

lua/luasnip/nodes/node.lua
```diff
@@ -11,6 +11,10 @@ function Node:new(o)
 	return o
 end
 
+function Node:is_node()
+	return true
+end
+
```
L3MON4D3, line 17: Can be removed I think

lua/luasnip/extras/init.lua
jedrzejboczar: Done

lua/luasnip/nodes/node.lua
jedrzejboczar: Removed

lua/luasnip/extras/fmt.lua
jedrzejboczar: I would rather stick with this simple regex search for now. I feel like doing plain text search would unnecessarily complicate the escaping logic. I guess it may even make things slower, because on each `text_add` we would need to `gsub` all the `}}` if I'm correct.

L3MON4D3: True, it's alright if we leave it like this👍

L3MON4D3:
Okay, thanks a lot, this should make defining snippets even easier👍👍

PR #173, opened 2021-09-17
Add json options (fixes #172)

### LINKED ISSUE #172: Autotrigger for JSON snippets
I may be wrong, but I do not think that it is currently possible to have auto triggering snippets for the ones defined in JSON and loaded with something like this:
```lua 
require("luasnip/loaders/from_vscode").lazy_load({ paths = { '~/.config/nvim/snippets' } })
```

I would be cool if we could specify an extra field in the JSON file (_e.g._ `autotrigger`) so the snippets get added to `ls.autosnippets` instead of `ls.snippets`. 

I am not overly competent in lua, but I think that just checking if `autotrigger = true` and deciding if the snippet needs to be added to the `ls.snippet` or `ls/autosnippet` should be enough ([here](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/loaders/from_vscode.lua#L84-L93)) (I tried to do it on my own and make a PR, but after 2 hours, I stand defeated).

leiserfg:


lua/luasnip/loaders/from_vscode.lua
```diff
@@ -81,19 +81,25 @@ local function load_snippet_file(langs, snippet_set_path)
...
+								name = name,
+								dscr = parts.description or name,
+								wordTrig = true,
+							}, body)
+
+							if ls_conf.autotrigger then
+								table.insert(auto_lang_snips, snip)
+							end
+
+							table.insert(lang_snips, snip)
```
L3MON4D3, line 97: Double insert? Am I missing something?

leiserfg: Should it go to just the auto list? I did that way the first time and ended by doing it like this as it's not clear to me which one should be, let me know what's your opinion and I will change it as you like.

L3MON4D3: Just auto I think, having it in the normal one won't matter because it will auto expand before expand is pressed.
Not sure if autosnippets show up in cmp_luasnip, but that would be a separate issue.

L3MON4D3:
Nice, thanks a lot :D

PR #189, opened 2021-10-03
feat(events): emit loaded event after filetype snippets loaded

captainko:
Attempt to add a loaded event for cmp_luasnip to handle lazy_load snippets.
See https://github.com/saadparwaiz1/cmp_luasnip/pull/19

### DIFF
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -101,6 +101,9 @@ local function load_snippet_file(langs, snippet_set_path)
 				end
 				ls.snippets[lang] = lang_snips
 				ls.autosnippets[lang] = auto_lang_snips
+				session.latest_load_ft = lang
+				vim.cmd "doautocmd User LuasnipSnippetsAdded"
+
 			end
 		end)
 	)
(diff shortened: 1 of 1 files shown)


L3MON4D3:
AFAICT this could fail if users do `require("luasnip/events/loader").register(...)`, a problem we already encountered, so you'd have to add the loader to the table returned by `require("luasnip")`.

Also, more generally, we already employ the pull-observer-model via autocommands ([LuasnipEnter/LeaveNode](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/nodes/node.lua#L142-L143)) for doing events, why do you think registering callbacks is preferable? Leveraging both to notify users of some change isn't that nice IMO, so we'd have to also adapt existing code (maybe) if we go with callbacks here.

captainko:
> AFAICT this could fail if users do `require("luasnip/events/loaded").register(...)`, a problem we already encountered, so you'd have to add the loader to the table returned by `require("luasnip")`.

I did not know that was a thing :smile:

> Also, more generally, we already employ the pull-observer-model via autocommands ([LuasnipEnter/LeaveNode](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/nodes/node.lua#L142-L143)) for doing events, why do you think registering callbacks is preferable? Leveraging both to notify users of some change isn't that nice IMO, so we'd have to also adapt existing code (maybe) if we go with callbacks here.

Yeah, I think you're right. I have moved the `User  LuasnipSnippetsAdded` at the `load_snippet_file()`.

L3MON4D3:
> I did not know that was a thing :smile: 

Yeah, the `require` caches by string, so `luasnip/events` != `luasnip.events`, meaning module-local variables aren't the same in both. No idea why it is that way :/

Thank you for the contribution :D

PR #190, opened 2021-10-05
Use new (and faster) json implementation if available

leiserfg:


L3MON4D3:
Sweet, I like the startup-only check👍
Thanks!!

PR #193, opened 2021-10-10
Adding cmp description

jonathf:
Until reasently Compe and Cmp had the same interfaces, but [that changed reasently](https://github.com/hrsh7th/nvim-cmp/issues/231) when they stopped using PUM in favor of floating window.

I stumbled upon this problem when my super-tab experience broke. It took a bit to figure out the issue, so this PR adds a description for getting supertab-behavior for cmp, as I assume others will likely stumble upon the same issue in the same way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,62 @@ smap <silent><expr> <C-E> luasnip#choice_active() ? '<Plug>luasnip-next-choice' 
 ```
   </details>
    <details>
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Nice, thanks a lot!

L3MON4D3:
Perhaps I was a bit too quick to merge :P
Could you open a new PR with these changes?

PR #194, opened 2021-10-10
Update with suggestions

jonathf:
I've made the changes as suggested.

My PR was a minimal change to the compe config you already have, and some of your suggestion are valid there as well.
But it might be wirth considering to derecate docs related to compe, as compe is officially deprecated in favor of cmp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,11 @@ vim.api.nvim_set_keymap("s", "<C-E>", "<Plug>luasnip-next-choice", {})
 ```
   </details>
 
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Could you also add a short node at the bottom of the details for cmp with [this](https://github.com/hrsh7th/nvim-cmp/wiki/Example-mappings#luasnip) link, showing an alternative way to setup some mappings?

L3MON4D3:
Yes! Thanks for getting the changes in so quickly :)

PR #196, opened 2021-10-12
Lazy_load on BufWinEnter,FileType to catch other ways of setting ft.

L3MON4D3:
- `BufEnter -> BufWinEnter`: `BufWinEnter` seems to only be triggered upon displaying or reloading the buffer, but importantly, after modeline has been read.
- `FileType`: Plugins, specifically `Neogit` set ft of their windows pretty late using `set ft=...`, `BufEnter` is triggered before ft is set, so no snippets are loaded for the later-set ft. Loading on `FileType` will load snippets in these cases (also, if ft is modified later on, manually, which is a nice feature)

### DIFF
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -247,7 +247,7 @@ function M.lazy_load(opts)
 	vim.cmd([[
 		augroup _luasnip_vscode_lazy_load
 		autocmd!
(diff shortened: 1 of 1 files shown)


bew:
typo?
```diff
-fix(loader): Lazy_load on BEW,FT to catch some weird ways to set ft.
+fix(loader): Lazy_load on BWE,FT to catch some weird ways to set ft.
```

PR #233, opened 2021-11-30
[WIP] Restore position upon changing choice.

L3MON4D3:
With `restoreNode`, it may be possible that the node that was active during `change_choice` also existst in the new choice. If that is the case, jump to it.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -256,9 +256,23 @@ ChoiceNodes allow choosing between multiple nodes.
 jumplist, and as its second a table with nodes, the choices.
 The third parameter is a table of options with the following keys:
 - `restore_cursor`: `false` by default. If it is set and the node that was
-  being edited also appears in the switched-to choice (can be the case if a
-  `restoreNode` is present in both choice) the cursor is restored relative to
-  that node.
+	being edited also appears in the switched-to choice (can be the case if a
+	`restoreNode` is present in both choice) the cursor is restored relative to
+	that node.
+	The default is `false` as enabling might lead to worse performance. It's
+	possible to override the default by wrapping the `choiceNode`-constructor
+	in another function that sets `opts.restore_cursor` to `true` and then using
+	that to construct `choiceNode`s:
+	```lua
+	local function restore_cursor_choice(pos, choices, opts)
+		if opts then
+			opts.restore_cursor = false
+		else
+			opts = {restore_cursor = false}
+		end
+		return c(pos, choices, opts)
+	end
+	```
 
 Jumpable nodes that normally expect an index as their first parameter don't
 need one inside a choiceNode; their index is the same as the choiceNodes'.
--- a/DOC.md
+++ b/DOC.md
@@ -252,8 +252,13 @@ ChoiceNodes allow choosing between multiple nodes.
  }))
 ```
 
-`c()` expects as it first arg, as with any jumpable node, its position in the
+`c()` expects as its first arg, as with any jumpable node, its position in the
 jumplist, and as its second a table with nodes, the choices.
+The third parameter is a table of options with the following keys:
+- `restore_cursor`: `false` by default. If it is set and the node that was
+  being edited also appears in the switched-to choice (can be the case if a
+  `restoreNode` is present in both choice) the cursor is restored relative to
+  that node.
 
 Jumpable nodes that normally expect an index as their first parameter don't
 need one inside a choiceNode; their index is the same as the choiceNodes'.
--- a/DOC.md
+++ b/DOC.md
@@ -266,9 +266,9 @@ The third parameter is a table of options with the following keys:
 	```lua
 	local function restore_cursor_choice(pos, choices, opts)
 		if opts then
-			opts.restore_cursor = false
+			opts.restore_cursor = true
 		else
-			opts = {restore_cursor = false}
+			opts = {restore_cursor = true}
 		end
 		return c(pos, choices, opts)
 	end
--- a/DOC.md
+++ b/DOC.md
@@ -457,11 +457,7 @@ local function simple_restore(args, _)
 	return sn(nil, {i(1, args[1]), r(2, "dyn", i(nil, "user_text"))})
 end
 
-ss("rest", {
-	i(1, "preset"), t{"",""},
-	d(2, simple_restore, 1)
-}),
-("rest", {
+s("rest", {
 	i(1, "preset"), t{"",""},
 	d(2, simple_restore, 1)
 }),
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -291,9 +291,9 @@ The third parameter is a table of options with the following keys:
     `lua
     local function restore_cursor_choice(pos, choices, opts)
         if opts then
-            opts.restore_cursor = false
+            opts.restore_cursor = true
         else
-            opts = {restore_cursor = false}
+            opts = {restore_cursor = true}
         end
         return c(pos, choices, opts)
     end
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -469,11 +469,7 @@ Every time the `i(1)` in the outer snippet is changed, the text inside the
     local function simple_restore(args, _)
     	return sn(nil, {i(1, args[1]), r(2, "dyn", i(nil, "user_text"))})
     end
-    ss("rest", {
-    	i(1, "preset"), t{"",""},
-    	d(2, simple_restore, 1)
-    }),
-    ("rest", {
+    s("rest", {
     	i(1, "preset"), t{"",""},
     	d(2, simple_restore, 1)
     }),
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -281,9 +281,23 @@ ChoiceNodes allow choosing between multiple nodes.
 jumplist, and as its second a table with nodes, the choices.
 The third parameter is a table of options with the following keys:
 - `restore_cursor`: `false` by default. If it is set and the node that was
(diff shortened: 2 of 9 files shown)


stasjok:
I tested it for simple cases. For everything implemented (always restore position, for input mode also stay where you were) looks like it's working. Later I tried to create a long snippet with various types of nodes (because I looked at the code and you introduced a new method for every node, which called recursively). For this long snippet I get an error when trying to switch. A can try to narrow it down by removing part of that snippet, by you can look at it as is (it is possible that I made an error somewhere):

```lua
local function test_restore_pos_dyn(args)
  local count = tonumber(args[1][1])
  local nodes = {}
  if count then
    for j = 1, count do
      nodes[j * 2 - 1] = t(" ")
      nodes[j * 2] = r(j, "dyn" .. j, i(1, "dyn_node" .. j))
    end
  else
    nodes[1] = t("Enter a number!")
  end
  return sn(nil, nodes)
end
    s("test_position_restore", {
      c(1, {
        sn(nil, {
          t({ "PrefixChoice1:", "" }),
          i(1, "Choice1_input"),
          t({ "", "" }),
          r(2, 1),
        }),
        sn(nil, {
          t({ "", "PostfixChoice2:" }),
          r(2, 1),
          t({ "", "" }),
          i(1, "Choice2_input"),
        }),
      }),
    }, {
      stored = {
        [1] = {
          t("Insert nodes: "),
          i(1, "node1"),
          t(" "),
          i(2, "node2"),
          t({ "", "Choice node: " }),
          c(3, {
            sn(nil, {
              r(1, 2, {
                i(1, "choice_input1"),
                t(" "),
                i(2, "choice_input2"),
              }),
              t(" "),
              i(2, "choice1_input1"),
            }),
            sn(nil, {
              i(1, "choice1_input1"),
              t(" "),
              r(2, 2),
            }),
          }),
          t({ "", "Snippet node: " }),
          sn(4, {
            i(1, "snippet_node1"),
            t(" "),
            i(2, "snippet_node2"),
          }),
          t({ "", "Dynamic node (number of inputs): " }),
          i(5, "1"),
          d(6, test_restore_pos_dyn, 5),
          t({ "", "Restore node: " }),
          r(7, "restore", {
            i(1, "restore_node1"),
            t(" "),
            i(2, "restore_node2"),
          }),
          t({ "", "Dynamic lambda (reverse of 1): " }),
          dl(8, l._1:reverse(), 1),
        },
      },
    }),
```
The error is `E5108: Error executing lua ...site/pack/packer/start/LuaSnip/lua/luasnip/util/mark.lua:82: attempt to index a nil val
ue` and later:
```
Error detected while processing TextChangedI Autocommands for "*":
E5108: Error executing lua ...k/packer/start/LuaSnip/lua/luasnip/nodes/dynamicNode.lua:191: attempt to index field 's
nip' (a nil value)```

**Comment by stasjok** [comment]
Looks like the culprit is dynamic lambda. Also I edited the snippet above a bit. For lambda I still don't quite understand what it does and how. I thought that it simply allows you to use lua string functions for node with index (_1, _2 etc), but for example `l._1:len()` doesn't work for me.

**Comment by stasjok** [comment]
OK, without dynamic lambda my long snippet works. LuaSnip manages to always find the correct node. With `l._1:len()` I finally understood, that it returns number, not a string. That's why it's failing.

**Comment by L3MON4D3** [comment]
Nice, that's a great test, if it works there I can't see it failing anywhere :D
Does restore work with the current iteration of that snippet?
Concerning the slowness, so far I haven't experienced any but I'll try with your snippet later and maybe profile it to find what takes longest.

**Comment by stasjok** [comment]
> Nice, that's a great test, if it works there I can't see it failing anywhere :D

It is failing for dynamic lambda. It's working only if I remove it.

> Concerning the slowness, so far I haven't experienced any but I'll try with your snippet later and maybe profile it to find what takes longest.

Concerning performance looks like one restore_node with many input nodes is fast, but if I use separate restore_node for every input_node, than it's considerably slower. For dynamic_node it's the worst (every input is a separate stored_node). I made a test case, that can help you to profile:

```lua
local function gen_nodes(count, node_type)
  local uv = require("luv")
  local time = uv.hrtime()
  count = count or 50
  local nodes = {}
  for j = 1, count do
    nodes[j * 3 - 2] = t("Node" .. j .. " ")
    if not node_type or node_type == 0 then
      nodes[j * 3 - 1] = i(j, "value" .. j)
    elseif node_type == 1 then
      nodes[j * 3 - 1] = r(j, j, i(1, "value" .. j))
    elseif node_type == 2 then
      nodes[j * 3 - 1] = sn(j, { i(1, "value" .. j) })
    else
    end
    nodes[j * 3] = t({ "", "" })
  end
  nodes[#nodes + 1] = t("Generation time: " .. tostring(uv.hrtime() - time) .. "ns")
  return nodes
end

local function gen_nodes_dyn(_, _, _, count, node_type)
  local nodes = gen_nodes(count, node_type)
  return sn(nil, nodes)
end


ls.snippets = {
  all = {
    s("test_many_nodes_simple", {
      c(1, {
        sn(nil, { t({ "Choice 1:", "" }), r(1, "store", gen_nodes()) }),
        sn(nil, { t({ "Choice 2:", "" }), r(1, "store") }),
      }),
    }),
    s("test_many_nodes_restored", {
      c(1, {
        sn(nil, { t({ "Choice 1:", "" }), r(1, "store", gen_nodes(nil, 1)) }),
        sn(nil, { t({ "Choice 2:", "" }), r(1, "store") }),
      }),
    }),
    s("test_many_nodes_snippeted", {
      c(1, {
        sn(nil, { t({ "Choice 1:", "" }), r(1, "store", gen_nodes(nil, 2)) }),
        sn(nil, { t({ "Choice 2:", "" }), r(1, "store") }),
      }),
    }),
    s("test_many_dyn_nodes_simple", {
      c(1, {
        sn(nil, {
          t({ "Choice 1:", "" }),
          r(1, "store", {
            i(1, "trigger"),
            t({ "", "" }),
            d(2, gen_nodes_dyn, 1),
          }),
        }),
        sn(nil, {
          t({ "Choice 2:", "" }),
          r(1, "store"),
        }),
      }),
    }),
    s("test_many_dyn_nodes_stored", {
      c(1, {
        sn(nil, {
          t({ "Choice 1:", "" }),
          r(1, "store", {
            i(1, "trigger"),
            t({ "", "" }),
            d(2, gen_nodes_dyn, 1, nil, 1),
          }),
        }),
        sn(nil, {
          t({ "Choice 2:", "" }),
          r(1, "store"),
        }),
      }),
    }),
    s("test_many_dyn_nodes_snippeted", {
      c(1, {
        sn(nil, {
          t({ "Choice 1:", "" }),
          r(1, "store", {
            i(1, "trigger"),
            t({ "", "" }),
            d(2, gen_nodes_dyn, 1, nil, 2),
          }),
        }),
        sn(nil, {
          t({ "Choice 2:", "" }),
          r(1, "store"),
        }),
      }),
    }),
  },
}
```
I don't know how to measure time that's needed for switching choices, I did it visually by looking at the first line (`Choice 1/Choice 2`).

stasjok:
OK, dynamic_lambda problem has nothing to do with this PR. It's a separate bug. I don't know what is going on, but it is somehow connected with dynamic_node.

```lua
local function test_restore_pos_dyn(args)
  local count = tonumber(args[1][1])
  local nodes = {}
  if count then
    for j = 1, count do
      nodes[j * 2 - 1] = t(" ")
      nodes[j * 2] = r(j, "dyn" .. j, i(1, "dyn_node" .. j))
    end
  else
    nodes[1] = t("Enter a number!")
  end
  return sn(nil, nodes)
end


    s("test_position_restore_dyn_lambda", {
      c(1, {
        sn(nil, {
          t({ "Choice 1:", "" }),
          r(1, 1, {
            i(1, "input"),
            t({ "", "" }),
            dl(2, l._1:reverse(), 1),
          }),
        }),
        sn(nil, {
          t({ "Choice 2:", "" }),
          r(1, 1),
        }),
      }),
    }),
    s("test_lambda_with_dynamic_node", {
      c(1, {
        sn(nil, {
          i(1, "input"),
          t({ "", "" }),
          l(l._1:reverse(), 1),
          t({ "", "" }),
          i(2, "1"),
          d(3, test_restore_pos_dyn, 2),
        }),
        i(1),
      }, { restore_cursor = false }),
    }),
```
Snippet `test_position_restore_dyn_lambda` is working, but `test_lambda_with_dynamic_node` have this problem. You need to make a choice two times to trigger it. If I remove either `l` or `d` from the choices, then separately they are working.

L3MON4D3:
Concerning performance, could you open one more issue for that?

stasjok:
> Nice, I figured out whats wrong, the last sentence was an important hint 👍

Great. It is working now. But I would disable it by default. First, it is slowing down choices even if there is not a single restore_node. Second, I think it is needed for special snippets.

I mapped a key to something like this: `<Cmd>lua local uv = require("luv"); local time = uv.hrtime(); require('luasnip').change_choice(1); vim.notify(tostring((uv.hrtime() - time)/1000000).."ms")<CR>`

Now I can switch a couple of times, then look in `:messages`.

For a snippet `test_many_dyn_nodes_simple` from my earlier post, when i switch from node1, I get like 5-10 ms timing. When I switch from node25 I get like 50-60 ms, from node50 I get 100 ms, and it is noticeable. If `restore_cursor = false` I get constant 5ms. 

> Concerning performance, could you open one more issue for that?

OK. I think restore_nodes is quite heavy (more of them the worse). But current PR also adds up to that depending on current active node.

stasjok:
Should be `opts.restore_cursor = true` in example.

PR #249, opened 2021-12-21
Better argnodes.

L3MON4D3:
This is a general overhaul of the argnode/dependents-system.
Previously, a dynamic/functionNode simply stored references to their argnodes + the argnodes a reference to their dependents. That works as long as 
1. all nodes are available at the time the references are created and
2. the node that is depended upon isn't removed (and newly added).

It will fail if eg. a `functionNode` depends on a node inside a `choiceNode`, which is currently not visible (or on a node that doesn't even exist yet, perhaps due to it being generated by a `dynamicNode`).

One (pretty big, but practical) example for what is possible with the new system:
```lua
local ai = require("luasnip.nodes.absolute_indexer")

s("class", fmta([[
	---@class <>
	local <> = {<>}

	<>
]], {
	rep(1), -- parent-relative indexing still works.
	i(1, 'MyClass'),
	i(2),
	c(3, {
		fmta([[
			<>function <>:new(o)
				o = o or {}
				setmetatable(o, self)
				self.__index = self
				return o
			end
		]], {
			i(1),
			rep(ai[1]) -- refers to first `insertNode` relative to surrounding snippet.
		})
		t''
	}),
}))
```

(The `absolute_indexer` can be constructed as
- `ai[1][2][3]`
- `ai(1,2,3)`
- `ai{1,2,3}`
(I prefer the `[]`-version, but that may be too much to type, which is why the others are implemented))

### Concerns:
- Right now, a node may only update if all argnodes are visible. One alternative could be to just update, but use the old text for missing nodes. The prior approach is probably easier, but I'm unsure if the latter may be better.
- Performance: The absolute positions have to be resolved on each update. I haven't experienced any noticeable lag so far, but I'd like to be certain that it performs comparably to the current system before merging into master.
- Currently we assume that the current node will remain valid after an update, but that may not be the case if the current node is inside a dynamicNode with _cyclic dependencies_, which can cause it to be updated as a result of changing text inside it.
Right now the snippet will just be unlinked, it's possible to handle it so the snippet stays active, 
 but that doesn't seem worth it as I can't think of something practical that could be achieved with cyclic dependencies.

I'd like to get some opinions (...perhaps also some early testing :P), so feel free to give some feedback before this is merged.


uga-rosa:
I tried the above example on this branch, but it gives an error on expanding. Is it still incomplete?

```lua
local ls = require("luasnip")
local s = ls.snippet
local sn = ls.snippet_node
local isn = ls.indent_snippet_node
local t = ls.text_node
local i = ls.insert_node
local f = ls.function_node
local c = ls.choice_node
local d = ls.dynamic_node
local r = ls.restore_node
local fmta = require("luasnip.extras.fmt").fmta
local ai = require("luasnip.nodes.absolute_indexer")

ls.snippets = {
    lua = {
        s(
            "class",
            fmta(
                [[
	---@class <>
	local <> = {<>}

	<>
]],
                {
                    rep(1), -- parent-relative indexing still works.
                    i(1, "MyClass"),
                    i(2),
                    c(3, {
                        fmta(
                            [[
			<>function <>:new(o)
				o = o or {}
				setmetatable(o, self)
				self.__index = self
				return o
			end
		]],
                            {
                                i(1),
                                rep(ai[1]), -- refers to first `insertNode` relative to surrounding snippet.
                            }
                        ),
                        t(""),
                    }),
                }
            )
        ),
    },
}
```
```error
E5108: Error executing lua .../packer/start/LuaSnip/lua/luasnip/nodes/functionNode.lua:108: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
        [C]: in function 'ipairs'
        .../packer/start/LuaSnip/lua/luasnip/nodes/functionNode.lua:108: in function 'set_dependents'
        .../pack/packer/start/LuaSnip/lua/luasnip/nodes/snippet.lua:982: in function 'set_dependents'
        .../pack/packer/start/LuaSnip/lua/luasnip/nodes/snippet.lua:612: in function 'fake_expand'
        .../pack/packer/start/LuaSnip/lua/luasnip/nodes/snippet.lua:654: in function 'get_docstring'
        ...e/pack/packer/start/cmp_luasnip/lua/cmp_luasnip/init.lua:35: in function 'get_documentation'
        ...e/pack/packer/start/cmp_luasnip/lua/cmp_luasnip/init.lua:114: in function 'resolve'
        .../nvim/site/pack/packer/start/nvim-cmp/lua/cmp/source.lua:342: in function 'resolve'
        ...e/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/entry.lua:431: in function 'resolve'
        ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:344: in function <...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:341>
        ...te/pack/packer/start/nvim-cmp/lua/cmp/utils/feedkeys.lua:51: in function <...te/pack/packer/start/nvim-cmp/lua/cmp/utils/feedkeys.lua:49>
```

L3MON4D3:
Ah, I only tested expanding it normally, very possible I missed something about `get_docstring`. I'll have to remember to add tests for those... Ty!

uga-rosa:
Ok. I'm looking forward to it as it seems useful!

PR #276, opened 2022-01-27
fix(from_snipmate): Delete a line that are no longer needed thanks to loader_util.split_lines()

uga-rosa:


uga-rosa:
You can use `jit.os` to detect OS. However, even if it's OSX, it doesn't know if newline code is `\r` or `\n`, so this may be fine.

leiserfg:
The value does not depend on the os used to read, but in the one used to write the file (it can be from elsewhere) so is safer to check the endlines and endfiles chars from the file itself, instead of assuming the ones from the os.

uga-rosa:
Oh, that's a blind spot. Ty.
It seems smart to just look up the last letter.

L3MON4D3:
Completely forgot about removing that, thanks :+1:


> It seems smart to just look up the last letter.

That could be changed for Unix and Mac, we'd need a `find` for DOS at least because lines are separated, not terminated, with `\r\n`.

uga-rosa:
If the last letter is neither \r nor \n, can't we just assume it's CRLF?

L3MON4D3:
Ah right, that would work except when dos+last line is empty, but that could be caught as a special case, eg.
Ends with
* \r\n -> dos
* \n -> unix
* \r -> mac
* anything else -> dos

Would you like to do a PR for that?

PR #285, opened 2022-02-06
Add On the fly snippets

leiserfg:


leiserfg:
An idea borrowed from https://github.com/abo-abo/auto-yasnippet

L3MON4D3:
Oho, that's pretty neat👌👌
How exactly does the new syntax work?
`$word = module.$word`?

leiserfg:
Yes

lua/luasnip/extras/otf.lua
```diff
@@ -0,0 +1,86 @@
...
+	local current_text_arg = { "" }
+	local nodes = {}
+	local know_inputs = {}
+	local last_input_pos = 0
+
+	for _, part in ipairs(chunks) do
+		if part[1] == T.TXT then
+			current_text_arg[#current_text_arg] = current_text_arg[#current_text_arg]
+				.. part[2]
+		elseif #current_text_arg > 0 then
```
L3MON4D3, line 46: IIUC this is for committing merged TXT to the nodes, right? If so, it would generate some empty textNodes because the empty `current_text_arg` is `{""}`, which matches the `if`.
Checking for length > 1 or [1] ~= "" should work (?)

lua/luasnip/extras/otf.lua
```diff
@@ -0,0 +1,86 @@
...
+
+local last_snip = nil
+local last_reg = nil
+
+local M = {}
+
+-- Create snippets On The Fly
+-- It's advaisable not to use the default register as luasnip will probably
+-- override it
+function M.otf()
```
L3MON4D3, line 78: What do you think about passing the register used for generating the snippet to otf? That way one could potentially have multiple otf-snippets at a time.

leiserfg: That's not needed `getreg()` uses the last specified register so you can do `"x` before calling it and that works as you want.

L3MON4D3: Last used in general or last used named register? I think preserving the otf snippet over deletions or similar would be nice.

lua/luasnip/extras/otf.lua
```diff
@@ -0,0 +1,86 @@
...
+
+local function txt_to_snip(txt)
+	local t = ls.t
+	local sn = ls.s
+	local i = ls.i
+	local f = ls.f
+
+	-- The parser does not handle empty strings
+	if txt == "" then
+		return sn(t({ "" }))
```
L3MON4D3, line 32: I think this (and the instantiation below) should both be `s(...)`, right?

bew:
What about giving a more explicit name to the file &  the function? Like `on_the_fly`

L3MON4D3:
> > What about giving a more explicit name to the file & the function? Like `on_the_fly`
> 
> I just wanted to make it short, the longer version is fine to me, I can change it if @L3MON4D3 agree.

Sounds good to me👍👍

lua/luasnip/extras/otf.lua
leiserfg: It takes the value of `v:register` which is the last used register, that's why in the doc I'm specifying the register before calling the function in the shortcuts.

L3MON4D3: Right, I'm thinking about expanding the recorded snippet multiple times, ie. I'd like to expand the recorded snippet 5 times, but after the third I delete a line, which makes another register v:register, so I'd lose the recorded snippet.
(At least that's how I understand the current implementation)

lua/luasnip/extras/otf.lua
```diff
@@ -0,0 +1,86 @@
...
+
+local function txt_to_snip(txt)
+	local t = ls.t
+	local s = ls.s
+	local i = ls.i
+	local f = ls.f
+
+	-- The parser does not handle empty strings
+	if txt == "" then
+		return s(t({ "" }))
```
L3MON4D3, line 32: Only trigger missing here

lua/luasnip/extras/otf.lua
leiserfg: Again, it will work fine if you use the shortcut as I said in the doc,  as that way it's always setting the correct register before calling the function.
```vim
nnoremap <leader>e  "p:lua require('luasnip.extras.otf').on_the_fly()<cr>
```
Note the  `"p`, you could use any other register there instead, like:
```vim
nnoremap <leader>e  "a:lua require('luasnip.extras.otf').on_the_fly()<cr>
```

L3MON4D3: Ahh, I glanced over that and thought it was only for the initial yank/creation of the snippet, apologies :sweat_smile:

L3MON4D3:
Another, more general thought: Opening a popup-window, into which the otf (or other, eg. lsp/lua) snippets can be written seems (to me at least) a bit more comfortable (and flexible) than writing it into the buffer and then selecting that exactly (or the whole line, but then we'll have to deal with trimming spaces correctly)

leiserfg:
I prefer not to use popups, so users can enjoy syntax highlighting, linters, completion...

L3MON4D3:
Okay, I might look into that another time :+1:

leiserfg:
> I'd like to be able to trigger the otf-snippets directly from INSERT, because NORMAL restricts valid cursor-positions (at least with `virtualedit~=all`) eg. it's not possible to position the cursor (and with that, expand a snippet) behind the last character of a line. Is there a way to set `v:register` from INSERT without leaving INSERT (and with that, repositioning the cursor)?

It is posible, you can use <cmd> in the mapping to call the function as if it where in normal mode.

L3MON4D3:
Do you mean like `inoremap <C-a> lua require("luasnip.extras.otf").on_the_fly()`? If so, I'd still need to set the last-used register before the call and I'm a bit unsure on how to do that

L3MON4D3:
What do you think about the spaces? The most comfortable way of using otf-snippets would (I guess) be writing it and then deleting the whole thing with `dd`, potentially n`dd`.
But with that the snippet would include the entire line before the snippet (and would get indented even deeper on expansion I believe), so common indent should probably be removed (reuse code from SELECT_DEDENT?)
(the second problem, eg. indent on multiline snippets would need to be handled even if the user perfectly selects their snippet)

leiserfg:
`:inoremap <c-a>  <C-O>"p:lua require('luasnip.extras.otf').on_the_fly()<cr> `
this works in insert mode.
Regarding the indent. I really won't like to delete them, as like you said, I prefer to use the full like for this kind of snippets.

L3MON4D3:
That doesn't work :/

https://user-images.githubusercontent.com/41961280/152859045-ca48684b-122f-4544-bcdf-f94b6e94c746.mp4

I'm using `lsp_expand`, but both use `snip_expand()` so this will also occur with otf: the last `b` is pushed behind the snippet with `<C-O>` (because, at expansion, the), with `<cmd>` it works correctly, but we cannot easily set `v:register`

We could solve it by passing the register to get the text from to `otf()` :P

leiserfg:
That can be avoided by using `<c-\><c-o>` instead of `<c-o>` so the cursor does not move, I just tested it. I could for sure change the args to the function, at this point I'm just avoiding having to rewrite the doc :smile: 
Let me change it so everybody is happy.

L3MON4D3:
> That can be avoided by using `<c-\><c-o>` instead of `<c-o>` so the cursor does not move, I just tested it. I could for sure change the args to the function, at this point I'm just avoiding having to rewrite the doc smile Let me change it so everybody is happy.

Thank you :smile:
(didn't know about `<c-\><c-o>`, that's useful :+1:)

> Regarding the indent. I really won't like to delete them, as like you said, I prefer to use the full like for this kind of snippets.

I'm not sure what you mean, but when it comes to multiline-snippets the indent on lines following the first will have to be removed :/ (luasnip assumes that the snippet is not indented at all and insert the indent at which the snippet was expanded behind all linebreaks)

L3MON4D3:
Could you rebase? I reenabled stylua for PRs, no idea if it will be applied after merging :sweat_smile:

L3MON4D3:
Weird, not quite sure why committing the docs works fine, while stylua fails... I'll look into it tomorrow

L3MON4D3:
One more thing for this PR: tests :D
Could you add some? I'd be fine with some integration-tests, if you feel like it unit-tests for `dedent` and the parser.

leiserfg:
@L3MON4D3   Do I need to add integration tests? I mean, as the expansion of snippets is already tested elsewhere and  I added tests for the parsing and snippet generation already.

L3MON4D3:
Oh true, yeah this is probably enough :+1:

lua/luasnip/extras/otf.lua
```diff
@@ -0,0 +1,87 @@
...
+end
+
+local last_snip = nil
+local last_reg = nil
+
+-- Create snippets On The Fly
+-- It's advaisable not to use the default register as luasnip will probably
+-- override it
+function M.on_the_fly(regname)
+	local reg = vim.fn.getreg(regname) --Uses the last set register if regname is nil
```
L3MON4D3, line 80: `getreg(nil)` seems to always return `""`, `getreg("")` works correctly.

leiserfg: Then I'll replace it, and I will rebase so the doc changes in master get applied.
Btw I updated the version of md2vim recently as I noticed there were some changes upstream.

L3MON4D3:
I'm just trying them out, looks like one newline is appended; is that intentional?

https://user-images.githubusercontent.com/41961280/153748448-091780be-2757-4e2f-abe9-dcc9a4ec8a7b.mp4

L3MON4D3:
Oh nvm, that's because of linewise selection... that's unfortunate. Should we ignore a trailing newline?

L3MON4D3:
That would be nice :+1:

leiserfg:
Wait, `c` will enter in insert mode so I will have to rework the whole mapping :smile:

L3MON4D3:
Yeah, a bit :D
In the video I used `vnoremap <c-f> "ec<C-\><C-O>:lua require('luasnip.extras.otf').on_the_fly()<Cr>`

lua/luasnip/extras/otf.lua
```diff
@@ -0,0 +1,88 @@
...
+local last_reg = nil
+
+-- Create snippets On The Fly
+-- It's advaisable not to use the default register as luasnip will probably
+-- override it
+function M.on_the_fly(regname)
+	regname = regname or ""
+	local reg = table.concat(vim.fn.getreg(regname, 1, true), "\n") -- Avoid eol in the last line
+	if last_reg ~= reg then
+		last_snip = M._txt_to_snip(reg)
```
L3MON4D3, line 83: Oh, one last thing: `last_reg` is never actually set

Issue #281, opened 2022-02-05
Importing custom vscode snippets stopped working


danjenson:
Running
`:lua require("luasnip.loaders.from_vscode").load({ paths = { "~/.config/snippets" } })`

No longer works. This snippet directory has always worked, and previously I had adjusted the runtime path to include it, but I prefer this method (less messy). However, it seems to have stopped loading the snippets, even when called directly from a file that would use them, i.e. `t.tex` and all my snippets are for tex. For reference

`package.json`
```json
{
  "name": "custom-snippets",
  "contributes": {
    "snippets": [
      {
        "language": [
          "tex",
          "markdown"
        ],
        "path": "tex-math.json"
      },
      {
        "language": [
          "tex",
          "markdown"
        ],
        "path": "tex-stat.json"
      },
      {
        "language": [
          "markdown"
        ],
        "path": "markdown.json"
      }
    ]
  }
}
```
And one of these as an example, `tex-stat.json`:
```json
{
  "probability": {
    "prefix": "pr",
    "body": "\\Pr($1)$0"
  },
  "expectation": {
    "prefix": "E",
    "body": "\\mathbb{E}($1)$0"
  },
  "measurable events": {
    "prefix": "me",
    "body": "\\mathcal{L}"
  },
  "independent": {
    "prefix": "indep",
    "body": "\\perp"
  },
  "event space": {
    "prefix": "es",
    "body": "\\Omega"
  },
  "null set": {
    "prefix": ["null", "ns"],
    "body": "\\varnothing"
  }
}
```
These are all top level of the `~/.config/snippets` directory.

L3MON4D3:
Weird, it works fine for me (with `friendly-snippets`)...
Could you clone `friendly-snippets` (not necessarily to `rtp`) and see whether they load?

danjenson:
I already load friendly snippets, and it works just fine

danjenson:
Here is part of my config where I load luasnip:
```lua
	{
		"L3MON4D3/LuaSnip",
		wants = "friendly-snippets",
		after = "nvim-cmp",
		config = function()
			local luasnip = require("luasnip")
			luasnip.config.set_config({
				history = true,
				updateevents = "TextChanged,TextChangedI",
			})
			require("luasnip.loaders.from_vscode").lazy_load({ paths = { "~/.config/snippets" } })
		end,
	},
```

L3MON4D3:
> I already load friendly snippets, and it works just fine

[...]

PR #286, opened 2022-02-07
Make loading synchronous.

L3MON4D3:
Rip out all `vim.schedule` and luv-callbacks from loaders.
The problem with asynchronous loading is that race-conditions may occur when snippets are added after a call to `load()` (see #281)

@uga-rosa @leiserfg Could you take a look at these changes? I'm not completely sure if it's okay to remove the `vim.schedule`, but we don't call any `vim.api` in there so I guess it is..?

### DIFF
--- a/lua/luasnip/loaders/from_snipmate.lua
+++ b/lua/luasnip/loaders/from_snipmate.lua
@@ -71,22 +71,18 @@ local function load_snippet_file(lang, path)
 		return
 	end
 
-	Path.async_read_file(
-		path,
-		vim.schedule_wrap(function(buffer)
-			local snippet, extends = parse_snipmate(buffer, path)
-			if not snippet then
-				return
-			end
-			table.insert(extends, lang)
-			for _, ft in ipairs(extends) do
-				local lang_snips = ls.snippets[ft] or {}
-				ls.snippets[ft] = vim.list_extend(lang_snips, snippet)
-				session.latest_load_ft = ft
-				vim.cmd("do User LuasnipSnippetsAdded")
-			end
-		end)
-	)
+	local buffer = Path.read_file(path)
+	local snippet, extends = parse_snipmate(buffer, path)
+	if not snippet then
+		return
+	end
+	table.insert(extends, lang)
+	for _, ft in ipairs(extends) do
+		local lang_snips = ls.snippets[ft] or {}
+		ls.snippets[ft] = vim.list_extend(lang_snips, snippet)
+		session.latest_load_ft = ft
+		vim.cmd("do User LuasnipSnippetsAdded")
+	end
 end
 
 local function filter(exclude, include)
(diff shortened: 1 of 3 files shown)


leiserfg (approved):

L3MON4D3:
Nice, okay.
Loading the entirety of `vim-snippets` _does_ introduce a noticeable delay on startup, so we should probably notify users to switch to `lazy_load` or the method described in the wiki before this is merged.

uga-rosa:
Is this really asynchronous?

L3MON4D3:
It's possible I mixed the two up, I looked at a [definition](https://www.merriam-webster.com/dictionary/synchronous) and assumed that it's the same in CS, but it seems like you're right.
I'll change the title, ty👍

uga-rosa:
As @L3MON4D3 said above, even if this change slows down the startup time, we can use lazy load, so I don't think it will be a problem.

L3MON4D3:
Another idea to speed up loading:
Don't actually parse the snippets while loading, just store the string-representation into the snippet-table. The snippet will be parsed when it is triggered.

uga-rosa:
That's an interesting idea. I'd like to try it with larger snippets and see if the latency is a problem. (Considering the speed of Lua, I don't think it will be a problem)

L3MON4D3:
You can get a pretty close approximation by binding `ls.lsp_expand(snippet_body)` to some key.
In my experience it's instantaneous.

uga-rosa:
I tested it a bit and it seems to be fine.
Do you want to implement this feature in this PR?

L3MON4D3:
Yup, although I won't have time for a few weeks

L3MON4D3:
On master, `load()`ing `honza/vim-snippets` takes 1ms of startuptime (everything loaded asynchronally), here the startup is delayed by the full 55ms (3ms with `lazy_load()`).

What does work, however, is wrapping `load()` in `vim.schedule`. That allows one to both have <1ms delay during startup and do some work on the loaded snippets.
```lua
vim.schedule(function()
	require("luasnip.loaders.from_snipmate").load()
end)
```

I'm going to post as much in `Breaking Changes`, that way users/maintainers of neovim-configs already know what's going on/can prepare for when this PR is merged.

leiserfg:
@L3MON4D3  try 
```vim
vim.cmd [[au BufRead * ++once require("luasnip.loaders.from_snipmate").load() ]]
```

L3MON4D3:
Right, or like that 👍
(Does the autocommand have any additional benefit with `load`? I'll change the recommendation to it, if it has)

leiserfg:
It should take the same amount of time but should be less noticeable for the user (and won't appear in the startup time for those that are in that kind of competition).

L3MON4D3:
Ah sorry, I meant benefit compared to wrapping `load` in `vim.schedule`.

PR #306, opened 2022-02-13
Fix otf doc for when luasnip messes with the last register somehow

leiserfg:


leiserfg:
I noticed that sometimes the second usage does not work fine as the last register gets overridden, so let's make it explicit there.

leiserfg:
Don't merge it, I will it look even nicer :smile:

L3MON4D3:
Nice👌👌

L3MON4D3:
Ty!

Issue #312, opened 2022-02-19
jumps broken with langmap option


columbarius:
When setting the langmap luasnip will jump to the wrong positions.

Example config (init.lua):
```
----


--- install packer if unable to load
if not pcall(require, 'packer') then
  local packer_path = vim.fn.stdpath 'data' .. '/site/pack/packer/start/packer.nvim'
  vim.cmd('!git clone https://github.com/wbthomason/packer.nvim ' .. packer_path)
  print 'packer.vim installed ... please restart neovim'
end

-- init plugin manager
require('packer').startup(function(use)
  ---- load plugins
  -- packer can manage itself
  use 'wbthomason/packer.nvim'
  use 'neovim/nvim-lspconfig'
  ---- completion
  use {
    'hrsh7th/nvim-cmp',
    requires = {
      'hrsh7th/cmp-buffer',
      'hrsh7th/cmp-path',
      'hrsh7th/cmp-nvim-lua',
      'hrsh7th/cmp-nvim-lsp',
    },
  }
  ---- snippets
  use 'L3MON4D3/LuaSnip'
  use 'saadparwaiz1/cmp_luasnip'
  use 'rafamadriz/friendly-snippets'
  ---- treesitter
  use 'nvim-treesitter/nvim-treesitter'
end)

vim.opt.completeopt = "menu,menuone,noselect"
-- Setup nvim-cmp.
local cmp = require'cmp'
local luasnip = require("luasnip")
require("luasnip.loaders.from_vscode").load()

local has_words_before = function()
  local line, col = unpack(vim.api.nvim_win_get_cursor(0))
  return col ~= 0 and vim.api.nvim_buf_get_lines(0, line - 1, line, true)[1]:sub(col, col):match("%s") == nil
end

luasnip.config.setup({
  history = true,
})

cmp.setup({
  snippet = {
    -- REQUIRED - you must specify a snippet engine
    expand = function(args)
      -- vim.fn["vsnip#anonymous"](args.body) -- For `vsnip` users.
      require('luasnip').lsp_expand(args.body) -- For `luasnip` users.
      -- require('snippy').expand_snippet(args.body) -- For `snippy` users.
      -- vim.fn["UltiSnips#Anon"](args.body) -- For `ultisnips` users.
    end,
  },
  mapping = {
    --['<C-b>'] = cmp.mapping(cmp.mapping.scroll_docs(-4), { 'i', 'c' }),
[...]

PR #313, opened 2022-02-19
Make cursor movement more robust against langmap

columbarius:
Replacing "l" in nvim_feedkeys through "<Right>" and using
nvim_replace_termcodes first will prevent broken behaviour when
langmap is used to remap the l key.

Resolves: #312

### DIFF
--- a/lua/luasnip/util/util.lua
+++ b/lua/luasnip/util/util.lua
@@ -107,12 +107,18 @@ end
 local function normal_move_before(new_cur_pos)
 	-- +1: indexing
 	if new_cur_pos[2] - 1 > 0 then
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Ah nice, ty!
I'll probably extend that a bit to make it work with other langmaps (+improve in general, I learned new trick that should make selecting even nicer (feeding `<cmd>lua ... <cr>` for setting the cursor position)), but this is a good fix for the immediate problem :+1:

columbarius:
Thanks for merging. After you pointed me to the right place I had to fix it, it drove me mad ^^'.
But I'm really interested and will take a look on the new trick.

PR #315, opened 2022-02-21
fix(from_snipmate): extends

uga-rosa:
This PR is incomplete.
If you use load to load all the snippets at once, this is fine, but if you use lazy_load, there is a bug where the snippets are loaded repeatedly when the filetype changes, resulting in duplicate candidates.
If you prepare sh.snippets as follows and open hoge.sh, then run `:set ft=zsh`, you should be able to reproduce the problem.
```snippets
extends zsh
snippet test
    hi!
```
Furthermore, with lazy_load, if you open a file with ft=zsh without going through ft=sh, the snippet will not be added.

### DIFF
--- a/lua/luasnip/loaders/from_snipmate.lua
+++ b/lua/luasnip/loaders/from_snipmate.lua
@@ -7,7 +7,7 @@ local session = require("luasnip.session")
 local sp = require("luasnip.nodes.snippetProxy")
 
 local function parse_snipmate(buffer, filename)
-	local snippet = {}
+	local snippets = {}
 	local extends = {}
 
 	---@type string[]
@@ -44,7 +44,7 @@ local function parse_snipmate(buffer, filename)
 			dscr = description,
 			wordTrig = true,
 		}, body)
-		table.insert(snippet, snip)
+		table.insert(snippets, snip)
 	end
 
 	while i <= #lines do
@@ -52,8 +52,7 @@ local function parse_snipmate(buffer, filename)
 		if vim.startswith(line, "snippet") then
 			_parse()
 		elseif vim.startswith(line, "extends") then
-			local scopes = vim.split(vim.trim(line:sub(8)), "[,%s]+")
-			vim.list_extend(extends, scopes)
+			extends = vim.split(vim.trim(line:sub(8)), "[,%s]+")
 			i = i + 1
 		elseif vim.startswith(line, "#") or line:find("^%s*$") then
 			-- comment and blank line
@@ -63,43 +62,67 @@ local function parse_snipmate(buffer, filename)
 		end
 	end
 
-	return snippet, extends
+	return snippets, extends
 end
 
-local function load_snippet_file(lang, path)
+local function load_snippet_file(path)
 	if not Path.exists(path) then
 		return
 	end
 
-	Path.async_read_file(
-		path,
-		vim.schedule_wrap(function(buffer)
-			local snippet, extends = parse_snipmate(buffer, path)
-			if not snippet then
-				return
+	local snippet, extends
+
+	if cache.path_snippets[path] then
+		snippet = cache.path_snippets[path].snippet
+		extends = cache.path_snippets[path].extends
+	else
+		local buffer = Path.read_file(path)
+		snippet, extends = parse_snipmate(buffer)
+		cache.path_snippets[path] = { snippet = snippet, extends = extends }
+	end
+
+	return snippet, extends
+end
+
+local function _append(tbl, name, elem)
+	if tbl[name] == nil then
+		tbl[name] = {}
+	end
+	table.insert(tbl[name], elem)
+end
+
+---Get paths of .snippets files
+---@param roots string[] @snippet directory paths
+---@return table @keys are file types, values are paths
+local function get_ft_paths(roots)
+	local ft_path = {}
+	for _, root in ipairs(roots) do
+		local files, dirs = Path.scandir(root)
+		for _, file in ipairs(files) do
+			local ft, ext = Path.basename(file, true)
+			if ext == "snippets" then
+				_append(ft_path, ft, file)
 			end
-			table.insert(extends, lang)
-			for _, ft in ipairs(extends) do
-				local lang_snips = ls.snippets[ft] or {}
-				ls.snippets[ft] = vim.list_extend(lang_snips, snippet)
-				session.latest_load_ft = ft
-				vim.cmd("do User LuasnipSnippetsAdded")
+		end
+		for _, dir in ipairs(dirs) do
+			local ft = dir
+			files, _ = Path.scandir(Path.join(root, dir))
+			for _, file in ipairs(files) do
+				if vim.endswith(file, ".snippets") then
+					_append(ft_path, ft, file)
+				end
 			end
-		end)
-	)
+		end
+	end
+	return ft_path
 end
 
 local function filter(exclude, include)
-	vim.validate({
-		exclude = { exclude, "table" },
-		include = { include, "table", true },
-	})
-
 	exclude = loader_util.filetypelist_to_set(exclude)
 	include = loader_util.filetypelist_to_set(include)
 
 	return function(lang)
-		if exclude[lang] then
+		if exclude and exclude[lang] then
 			return false
 		end
 		if include == nil or include[lang] then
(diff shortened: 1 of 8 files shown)


uga-rosa:
I think I've solved the problem, but the WiFi in my room is not working, so the update will have to wait (this post is from my iPhone).

L3MON4D3:
xD
Okay, I'll be looking forward to that :+1:

uga-rosa:
In the case of lazy, only look at `extends` first, otherwise, examine `extends` at the timing of parsing the snippet.
Since `filetype_extend` is supported by `util.get_snippet_filetypes()` in `M._lazyload()`, this should not be taken into account in the filter (`loader_util.filetypelist_to_set()`). This was causing duplication.

L3MON4D3:
Ah, I understand.
What do you think of keeping the `extends` snipmate/snippet-library-local (so if `snippets/cpp.snippets` has `extends c` it would only include snippets from `snippets/c.snippet`)? I'd treat the `extends` more like a `#include`, eg. "use the snippet listed in that file here as well", not a global "extend the current filetype with this other filetype". That way the `extends` would also do something different than `ls.filetype_extend`, which IMO is a bonus.

uga-rosa:
Oh I see, #include. That's easier to understand and use.
I'll try it.

L3MON4D3:
TY, I'm glad we're on the same page :+1:
`lazy_load` probably needs to be a bit more complicated as there might be both `extends somefiletype` (in `filetype.snippets`) and `ls.fileteype_extend(filetype, {"somefiletype"})`, which could lead to duplicate snippets if we don't keep track of which files were already loaded for a given filetype.
(The root problem, of course, is that sources have their own ways of extending filetypes, which we have to respect. Everything would be easier if those wouldn't exist :| )

uga-rosa:
It took some design changes, but I was able to implement it. Reading files asynchronously was complicated and confusing, so I borrowed Path.read_file from that PR and used it. I was also able to support nested extends. Example, typescriptreact -> typescript -> javascript
WiFi is broken again, so I'll push tomorrow :'(

leiserfg:
Time to start coding in your iPhone bro :smile:

uga-rosa:
It works fine for my own testing.

leiserfg:
@uga-rosa  could you profile the loading? Maybe when can find out what's what makes it "slow", you just have to be using luajit 2.1 (it comes with the neovim from the releases, is the default in debian, for arch it's in AUR) and do something like:
```
require("jit.p").start("10,i1,s,m0,G", "/tmp/output.log")
-- your function call here
require'jit.p'.stop()
```
And share your  `output.log`  that way we can try to find what's the slowest part of the load.

uga-rosa:
It's like this, isn't it?
```
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;fix_node_indices 39
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;init_nodes 4
[string]:0;load;_load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;fix_node_indices 4
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet 3
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;parse_text 2
[string]:0;load;_load;load_snippet_file;parse_snipmate;startswith 2
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;parse_text;gsplit 2
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;to_line_table;split;gsplit;validate; 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;brackets_offset 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;I 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;to_line_table;split;gsplit 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;new 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;fn;parse_snippet;parse_text;gsplit;validate; 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;fn;I 1
[string]:0;load;_load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;fn;split;gsplit 1
[string]:0;load;_load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;init_nodes 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;fn;text_only 1
[string]:0;load;_load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet 1
[string]:0;load;_load;load_snippet_file;parse_snipmate;_parse;parse_snippet;fn;parse_snippet;parse_text;gsplit 1
```

leiserfg:
flame

leiserfg:
@L3MON4D3  check this, maybe if you try it in your branch (the one for sync vscode loading) we can get more precise results (as we have more files there) but seems like `fix_node_indices` is a bottleneck.

L3MON4D3:
Oh holy.. `fix_node_indices` shouldn't be taking that long xD

L3MON4D3:
@uga-rosa You also included the previously unsupported (iirc :D ) `snippets/filetype/*.snippets`, right?

uga-rosa:
Hmmm? I think it supported {filetype}.snippets and {filetype}/*.snippets before.

leiserfg:
BTW the flame graph has extra data, if you click it you can see the percent of each bar while hovering it.

lua/luasnip/loaders/util.lua
```diff
@@ -1,16 +1,11 @@
...
 	if not list then
 		return list
 	end
 	local out = {}
 	for _, ft in ipairs(list) do
-		-- include redirected filetypes.
-		for _, resolved_ft in ipairs(session.ft_redirect[ft]) do
-			out[resolved_ft] = true
-		end
+		out[ft] = true
```
L3MON4D3, line 8: I don't think we can ignore the redirects (esp. not here, this is also used by the vscode-loader).

lua/luasnip/loaders/from_snipmate.lua
```diff
@@ -107,60 +130,50 @@ local function filter(exclude, include)
...
 
-	for _, path in ipairs(opts.paths) do
-		load_snippet_folder(path, opts)
+	for ft, _ in pairs(cache.ft_paths) do
+		if ft_filter(ft) then
+			local snippets = M._load(ft)
+			local lang_snips = ls.snippets[ft] or {}
+			ls.snippets[ft] = vim.list_extend(lang_snips, snippets)
+			session.latest_load_ft = ft
+			vim.cmd("do User LuasnipSnippetsAdded")
```
L3MON4D3, line 175: Could we store which snippet-files were added to which filetype and then skip duplicates here?

L3MON4D3:
> Hmmm? I think it supported {filetype}.snippets and {filetype}/*.snippets before.

Oh, nvm then :D

lua/luasnip/loaders/util.lua
uga-rosa: The vscode is also fine because it calls util.get_snippet_filetypes() in M._lazy(). Rather, the snippet file loading was duplicated when this was present.

L3MON4D3:
> BTW the flame graph has extra data, if you click it you can see the percent of each bar while hovering it.

It's only a picture for me :sweat_smile: But it's clear that `fix_node_indices` kind of drags everything down, which it definitely shouldn't.
I can investigate+fix what's going on there in my PR (when I get to it, in a bit), together with the on-expand-parsing, then we can first merge that and then this one (I think that'd be best, but not sure).

lua/luasnip/loaders/from_snipmate.lua
uga-rosa: I don't think snippet-files necessarily corresponds to a single filetype.
I believe that the extends syntax is only expected to work for .snippets.
For example, if you have extends javascript in typescript.snippets, and you open .ts and then open .js, then javascript.snippets should be referenced twice. So I'm caching it.

lua/luasnip/loaders/util.lua
L3MON4D3: Oh you're right, there is some duplication in vscode-lazy-load. AFAICT regular `load({include=some_extended_filetype})` has to load the snippets for the extended filetypes as well, and they are only included here.
But maybe it's best to not resolve redirected filetypes here, but only explicitly.

lua/luasnip/loaders/from_snipmate.lua
L3MON4D3: Ah, I was thinking completely wrong, nvm.

lua/luasnip/loaders/util.lua
uga-rosa: Yes, I think it would be better to explicitly select the snippet file you want to use if you are a user for such purposes.

L3MON4D3:
Should we warn users that both a (in `filetype.snippet`) `extends otherfiletype` and `ls.filetype_extend("filetype", {"otherfiletype"})` leads to duplicate snippets and the only fix is to change the `.snippets` (which might be non-trivial in cases of snippet-libraries, eg. `vim-snippet`)?

L3MON4D3:
> > It's only a picture for me sweat_smile But it's clear that fix_node_indices kind of drags everything down, which it definitely shouldn't.
> > I can investigate+fix what's going on there in my PR (when I get to it, in a bit), together with the on-expand-parsing, then we can first merge that and then this one (I think that'd be best, but not sure).
> 
> It was 57.35%. I'm curious...

Damn, that's quite a bit...

uga-rosa:
> Should we warn users that both a (in filetype.snippet) extends otherfiletype and ls.filetype_extend("filetype", {"otherfiletype"}) leads to duplicate snippets and the only fix is to change the .snippets (which might be non-trivial in cases of snippet-libraries, eg. vim-snippet)?

Hmm. It would certainly be better to write it in the help.

L3MON4D3:
I fixed the horrendous performance of `fix_node_indices`, which should make parsing snippets much faster.

before:

after:


(with `require("jit.p").start("10,i1,s,m0,G", "pre"/"post")` and on master, but it should look similar here)
Would it be correct to assume that `parse_snipmate` took ~289ms and now takes ~142ms (eg samples * sample interval roughly equals runtime)?

L3MON4D3:
I think the synchronous_load-PR can be merged tomorrow or so, @uga-rosa would you extend (haha get it :P) the existing snipmate-test suite with one for `extends`? (maybe just add `extends somefile` to the existing `all.snippets`, and make sure that snippets from `somefile.snippets` are present)

I'll be happy to help if you have any questions concerning the tests :)

uga-rosa:
Is it something like this? Please let me know if something is wrong.

L3MON4D3:
Jup, looks good 👍
Not at home rn, I'll look into merging everything when I get back

L3MON4D3:
I accidentially merged an older version of this PR (didn't update my local branch before merging), but now it should all be in order.

L3MON4D3:
I'm sorry, nvm, I'm getting this error when calling the regular `load()`, but not with `lazy_load()`:
```
E5108: Error executing lua ...cker/start/luasnip/lua/luasnip/loaders/from_snipmate.lua:151: bad argument #1 to 'ipairs' (table expected, got nil)
stack traceback:
        [C]: in function 'ipairs'
        ...cker/start/luasnip/lua/luasnip/loaders/from_snipmate.lua:151: in function '_load'
        ...cker/start/luasnip/lua/luasnip/loaders/from_snipmate.lua:155: in function '_load'
        ...cker/start/luasnip/lua/luasnip/loaders/from_snipmate.lua:172: in function 'load'
        [string ":lua"]:1: in main chunk
```
Somehow it doesn't happen in the tests.

I've got to go rn, could you look into it @uga-rosa?

uga-rosa:
Ah, let add `or {}`

L3MON4D3:
Thank you ❤️

uga-rosa:
Ah, I see, because scandir now returns the full path.
Ty.

PR #322, opened 2022-02-26
Follow symbolic links when looking for files

davidatsurge:
PR title should be self explanatory. Let me know if it's unclear in any way.

### DIFF
--- a/lua/luasnip/loaders/from_snipmate.lua
+++ b/lua/luasnip/loaders/from_snipmate.lua
@@ -109,7 +109,7 @@ end
 
 local function get_paths(root)
 	local ret = {}
-	local files = Path.scandir(root, "file")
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Looks good :+1: @uga-rosa could you also take a look?

Sidenote, and this doesn't have to be implemented in this PR, we should consider returning full paths from `Path.scandir`, then the symlinks could be fully resolved in there. It doesn't _really_ matter considering `luv` takes care of resolving when reading the files, but seems a bit cleaner to me.

uga-rosa:
Looks good. I changed `Path.scandir` in #315, so I'll respond to this.

L3MON4D3:
Perfect :ok_hand:
Thank you very much @davidatsurge :D

PR #323, opened 2022-02-28
SnippetProxy (delay parsing of lsp-snippet)

L3MON4D3:
This allows parsing textmate-style snippets on expand (currently on `matches()` because it's not present in `snippetProxy`, but that can be improved), not on startup, therefore (once that part is implemented) lowering startuptime when loading snippets from some package/library significantly, to only the time required to extract the snippet-definitions and trig/name/description from the files.

This reduced cost of parsing snippet will be required because of the synchronous loading, but it is definitely a nice-to-have as most snippets from big libraries aren't used even once in a session, so there's no reason to parse them.

Some parts are missing as of now, but it won't hurt to have the concept out in the open.

It's conceivable to implement something similar for lua-native snippets, although the effect is probably negligible.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -909,6 +909,26 @@ compatible with luasnip
 - We do not implement eval using \` (backtick). This may be implemented in the future.
 
 
+# SNIPPETPROXY
+
+`SnippetProxy` is used internally to alleviate the upfront-cost of
+loading snippets from eg. a snipmate-library or a vscode-package. This is
+achieved by only parsing the snippet on expansion, not immediately after reading
+it from some file.  
+`SnippetProxy` may also be used from lua directly, to get the same benefits:
+
+This will parse the snippet on startup...
+```lua
+ls.parser.parse_snippet("trig", "a snippet $1!")
+```
+
+... and this will parse the snippet upon expansion.
+```lua
+local sp = require("luasnip.nodes.snippetProxy")
+sp("trig", "a snippet $1")
+```
+
+
 # EXT\_OPTS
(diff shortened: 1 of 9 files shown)


L3MON4D3:
Can be tested via eg.
```lua
local ls = require("luasnip")
local sp = require("luasnip.nodes.snippetProxy")

ls.snippets.lua = {
    sp.new("trig", "$1 aaaa $2")
}
```

`new` might be changed, would fit in better with other nodes (which this isn't, really, but it makes sense to have in `nodes.snippetProxy` (does it? maybe `util` would be a better fit..))

leiserfg:
Why  exporting `{new=new}`  instead of `new`  so it works as:
```lua
local ls = require("luasnip")
local sp = require("luasnip.nodes.snippetProxy")

ls.snippets.lua = {
    sp("trig", "$1 aaaa $2")
}
```
I mean, to keep the same API as the other nodes.

L3MON4D3:
You're right, that's nicer.
I only had internal use (`loaders`) in mind initially, but we should just treat it like `snippet`/`s`, as it allows having lazily parsed snippets via lua.

L3MON4D3:
Obligatory comparison (with `"10,i1,s,m0,G"`), `parse_snipmate` on all of `honza/vim-snippets` takes ~63ms -> ~19ms.

pre
post

L3MON4D3:
I can't really measure the runtime-effect while expanding, but looks like it's <1ms as it doesn't show up in the profiler.

PR #324, opened 2022-03-05
add in_snippet to docs

hqurve:
`in_snippet` was exposed in https://github.com/L3MON4D3/LuaSnip/pull/292 however, it was not placed in the docs.

Also, im not quite too sure what the two different bullet notations mean ... so it probably needs changes

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1085,6 +1085,9 @@ the lazy_load.
 # API-REFERENCE
 
 `require("luasnip")`:
+
+- `in_snippet()`: returns true if the cursor is inside the current snippet.
+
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Ah, thank you, could you make that change in `DOC.md`? (The vim-help is generated from that file).
Also, I'm not sure why only some `-` are changed to `*`, could you try adding a newline above the first point?

L3MON4D3:
Ty!

L3MON4D3:
Ah, and it did fix the issue with the different bullet points👍

PR #331, opened 2022-03-12
fix(docs): md code block can't be indented

kunzaatko:
For the generation of docs from `md`, the code block cannot be indented, because it gets translated to
> `lua
> --- code
> `

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -384,16 +384,16 @@ The third parameter is a table of options with the following keys:
 	possible to override the default by wrapping the `choiceNode`-constructor
 	in another function that sets `opts.restore_cursor` to `true` and then using
 	that to construct `choiceNode`s:
-	```lua
-	local function restore_cursor_choice(pos, choices, opts)
-		if opts then
-			opts.restore_cursor = true
-		else
-			opts = {restore_cursor = true}
-		end
-		return c(pos, choices, opts)
+```lua
+local function restore_cursor_choice(pos, choices, opts)
+	if opts then
+		opts.restore_cursor = true
+	else
+		opts = {restore_cursor = true}
 	end
-	```
+	return c(pos, choices, opts)
+end
+```
 
 Jumpable nodes that normally expect an index as their first parameter don't
 need one inside a choiceNode; their index is the same as the choiceNodes'.
--- a/DOC.md
+++ b/DOC.md
@@ -769,8 +769,8 @@ The complete signature for the node is `match(argnodes, condition, then, else)`,
     "ABC" exactly, nothing otherwise.
   * `match(n, lambda._1:match(lambda._1:reverse()), "PALINDROME")` inserts
     "PALINDROME" if the nth jumpable node is a palindrome.
-  *
-    ```lua
+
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Nice catch, ty :)
I'd rather leave that code block inside the list though, does inserting an empty line above it help?
There's also other places where the indentation works alright (`:h snippets/callbacks`, the code block is indented with spaces there).
(There's also another occurrence of `lua` in `luasnip-extras/Examples`, would you fix that here as well?)

L3MON4D3:
I noticed that CI doesn't run in PRs, that should be fixed now. Could you rebase?

kunzaatko:
Is this right? I do not understand the CI that failed.

kunzaatko:
Also, there is a lot of trailing spaces. Should I also fix that?

L3MON4D3:
> Is this right? I do not understand the CI that failed.

Ah, right, I remember now why that was set to push only. `auto-commit` tries to commit to `L3MON4D3/patch-1`, instead of `kunzaatko/patch-1`, which of course fails. Not sure how to fix that.

L3MON4D3:
> Also, there is a lot of trailing spaces. Should I also fix that?

I think those are in `DOC.md` for paragraphs, better leave them in :D

kunzaatko:
OK, then. That would be better than to remove the paragraphs :D.
I will use this opportunity to thank you for the plugin. It is amazing and you have done a great job! I am astonished by the thought that must have gone into it. Thanks!

L3MON4D3:
> I will use this opportunity to thank you for the plugin. It is amazing and you have done a great job! I am astonished by the thought that must have gone into it. Thanks!

Thank you very much, really happy you're enjoying it :D

I'm sorry, could you rebase one more time? CI should work now, I think.

L3MON4D3:
Huh, weird, looks like it didn't change anything in `help.txt`.
You can run https://github.com/FooSoft/md2vim/releases/download/21.12.14.0/md2vim_linux_amd64.tar.gz locally too (`md2vim DOC.md help.txt`), but ci should start showing the changes to `help.txt` too.

kunzaatko:
I found out, that the second fix does not work and I can not get it to work with the bulletpoint. I think that maybe the only fix could be to remove the bulletpoint entirely, but that would ruin the markdown...

L3MON4D3:
```diff
-  *
-    ```lua
+
+  * ```lua
```
works, it generates slightly incorrect vimdoc, but that's alright as far as I'm concerned.

L3MON4D3:
Replacing the bulletpoints in `Extras` with subheadings could also look alright.

kunzaatko:
I think this is quite OK. Not perfect, but maybe this should be a problem upstream at the `md2vim` side.
Also, did you consider to setup GitHub discussions? They could be nice for sharing and helping for discussing different possibilities. It may be better than handling everything with issues...

L3MON4D3:
Hadn't yet thought about enabling discussions, but I can't see a reason not to. Thanks for the tip👍

PR #332, opened 2022-03-12
edit(docs): change expresion to more idiomatic lua

kunzaatko:
```lua
a = a or {}
```
is more lua like


kunzaatko:
This may be very opinionated, so feel free to reject and sorry if you do not agree and this is the case...

L3MON4D3:
No worries, it's definitely more readable :)
Thank you! :D

PR #333, opened 2022-03-12
Replace md2vim with panvimdoc

leiserfg:


leiserfg:
This is a POC, this one supports tables, is written in lua and the dev is very quick to reply. While the guys md2vim haven't even commented in a PR I did half a year ago.

L3MON4D3:
Yeah, that's a good idea. Did you notice any problems with the generated docs?
Trailing spaces are removed from the vimdoc, so that's a plus already, not sure if I'd make use of tables, but the possibility is nice.

leiserfg:
There was an small problem with the generator and I did a pr to fix it, and the fix got merged like 5 minutes after.

L3MON4D3:
Okay nice, then this is definitely the way forward
Thank you!

L3MON4D3:
Could you resolve the conflicts? I'm on the go and have no idea how to just use your version of the file via web-interface.

L3MON4D3:
Perfect, thanks a lot👍👍👍

Issue #339, opened 2022-03-15
[Question] Adding tree-sitter based expansion conditions


yilinwei:
Hi,

First of all, thanks for the awesome package; it's been really fun to play around with so far.

Would it be possible to upstream some more functions for tree-sitter based expansions e.g. expand only when found inside a node block? I've found them to be quite useful for autosnippets.

L3MON4D3:
Yeah those are pretty nice, and more would be great. I don't have much experience writing queries, if you already have some, feel free to open a PR :)

max397574:
that doesn't require queries
It's about node types
An example for this can be found here:
https://github.com/nvim-treesitter/nvim-treesitter/issues/1184#issuecomment-830388856
I think this could be done with a function snippet

L3MON4D3:
*"much experience with treesitter", as is evident at least now :D
That link looks promising, we could use a function that takes that table of node-names and returns the condition-function, which returns if the cursor is inside any of these nodes.

max397574:
you could do that
or just put it in the wiki and let users to it themselfs

molleweide:
Another tip on ts is to follow 'refactor.nvim'

L3MON4D3:
> you could do that or just put it in the wiki and let users to it themselfs

I think this is something I'd actually put in the repo, sounds flexible+useful enough. We also have a few conditions in `extras` already.

L3MON4D3:
> Another tip on ts is to follow 'refactor.nvim'. They have written nice helpers for ts that one can learn from.

Could you post a link? Couldn't find them :/

dvdsk:
> Could you post a link? Couldn't find them :/

I think he means https://github.com/ThePrimeagen/refactoring.nvim/blob/master/lua/refactoring/treesitter/treesitter.lua

molleweide:
Sorry, yeah the link provided by dvdsk is the correct one!

yilinwei:
> ... we could use a function that takes that table of node-names and returns the condition-function, which returns if the cursor is inside any of these nodes.

Yes, that's how I'm doing i currently. It works fairly well, but _is_ dependent on how good the grammar is for the target language.

dvdsk:
TJ has a youtube series on luasnip, he explains a treesitter snippet here: https://youtu.be/KtQZRAkgLqo?t=1350 and he mentions his next video will look at hem in detail. Maybe he has some useful insights?

molleweide:
[...]

PR #340, opened 2022-03-16
[WIP] Treesitter based conditions

yilinwei:
Initial speculative PR for discussion for issue #339.

### DIFF
--- a/lua/luasnip/extras/expand_conditions.lua
+++ b/lua/luasnip/extras/expand_conditions.lua
@@ -5,4 +5,24 @@ function M.line_begin(line_to_cursor, matched_trigger)
 	return line_to_cursor:sub(1, -(#matched_trigger + 1)):match("^%s*$")
 end
 
+function M.node_types(node_types)
+
+	local ts_utils = require'nvim-treesitter.ts_utils'
+	local node_types = type(node_types) == "table" and node_types or { node_types }
+
+	local function cond()
+		local node = ts_utils.get_node_at_cursor()
+                if node ~= nil then
+		        for _, node_type in ipairs(node_types) do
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Nice! Thanks for whipping something up quickly :+1:

AFAICT this works only if the cursor is directly inside one of the passed nodes, right? (Eg. if you're inside a function inside a class, the class-snippets wouldn't match). imo it should check all nodes up to the root (is there a function by `ts_utils` for doing so?).

Apart from that, I'd just be in favor of naming the function`in_node`, but that's pretty minor.

What do you think?

yilinwei:
@L3MON4D3

Yes, this probably isn't ideal for certain grammars; I'd like to leave this PR hanging for a while to understand some of the requirements of other languages.

I'm currently using it for rust, where an exact match is very useful. I've written a few (modified) snippets below which I personally use.

```rust
fn foo() {
  // the treesitter cursor within here is "block"
  let mut bar = 1; // expand lm. Note that I don't want to expand `lm` unless I'm within the "block" e.g. expanding on the rhs is invalid.
  let x = || {
    // this is also "block"
  };
}

// this is "source_file", `let` here is invalid

impl Foo {
  // this is "declaration_list"
  fn test() {
    // this is also "block"
  }
}

// this is "source_file"

#[derive(Default)] // within the arguments, the node is `meta_arguments`. Expand 'df'
struct Foo();
```

```lua
  local let_mut_snippet = s("lm", { t("let mut "), i(1), t(" = "), i(2), t(";") }, { cond = cond.in_node("block") })
  local default_derive_snippet = s("df", t("Default"), { cond = cond.in_node("meta_arguments") })
```

PR #344, opened 2022-03-17
Loader for lua-native snippets.

L3MON4D3:
As discussed [here](https://github.com/L3MON4D3/LuaSnip/discussions/341), a loader for lua-snippets.
Some stuff is still custom to my setup ([global scope for snippet-files](https://github.com/L3MON4D3/Dotfiles/blob/master/.config/nvim/lua/plugins/luasnip/helpers.lua#L3-L56)), and some is missing (reload+edit).

To avoid having to define all of `local s = ls.snippet local i = ls.insert_node ...` at the beginning of each file, I propose using `setfenv` for adding these definitions to all of the loaded lua-files.
The definition would be globally configurable, and default to the names (`s`, `i`, `t`, ...) used in `Examples/snippets.lua` (see the link above for the implementation, I'll port add that in this PR too). This has the added benefit of being able to set global defaults for some snippet-options, eg. for [choiceNode](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#choicenode).

### DIFF
--- a/lua/luasnip/loaders/from_lua.lua
+++ b/lua/luasnip/loaders/from_lua.lua
@@ -24,8 +24,7 @@ local function load_files(ft, files)
 		local size = vim.loop.fs_fstat(fd).size
 		local func_string = vim.loop.fs_read(fd, size)
 		-- bring snippet-constructors into global scope for that function.
-		func_string = 'dofile("/home/simon/.config/nvim/lua/plugins/luasnip/helpers.lua").setup_snip_env() '
-			.. func_string
+		func_string = 'require("luasnip").setup_snip_env() ' .. func_string
 		local file_res = loadstring(func_string)()
 
 		-- make sure these aren't nil.
--- a/lua/luasnip/loaders/from_lua.lua
+++ b/lua/luasnip/loaders/from_lua.lua
@@ -25,17 +25,20 @@ local function load_files(ft, files)
 		local func_string = vim.loop.fs_read(fd, size)
 		-- bring snippet-constructors into global scope for that function.
 		func_string = 'require("luasnip").setup_snip_env() ' .. func_string
-		local file_res = loadstring(func_string)()
+		local file_snippets, file_autosnippets = loadstring(func_string)()
 
 		-- make sure these aren't nil.
-		file_res.snippets = file_res.snippets or {}
-		file_res.autosnippets = file_res.autosnippets or {}
+		file_snippets = file_snippets or {}
+		file_autosnippets = file_autosnippets or {}
 
 		-- keep track of snippet-source.
-		cache.path_snippets[file] = file_res
+		cache.path_snippets[file] = {
+			snippets = file_snippets,
+			autosnippets = file_autosnippets,
+		}
 
-		vim.list_extend(ft_snippets, file_res.snippets)
-		vim.list_extend(ft_autosnippets, file_res.autosnippets)
+		vim.list_extend(ft_snippets, file_snippets)
+		vim.list_extend(ft_autosnippets, file_autosnippets)
 
 		-- use lua autocommands here as soon as they're stable.
 		-- stylua: ignore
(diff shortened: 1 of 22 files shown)


L3MON4D3:
:P

leiserfg:
As you see in the last commit the enumerations in the doc were fixed, I fixed the panvimdoc upstream :tada:

L3MON4D3:
Yeah, I saw that and kind of wondered what changed xD
Very nice :ok_hand:

lua/luasnip/loaders/from_lua.lua
```diff
@@ -0,0 +1,209 @@
...
+			]],
+			-- augroup name may not contain spaces.
+			file:gsub(" ", "_"),
+			-- escape for autocmd-pattern.
+			file:gsub(" ", "\\ "),
+			file
+		))
+	end
+
+	ls.add_snippets(ft, ft_snippets, { type = "snippets" })
```
mathjiajia, line 60: here, `ls` has not been defined

L3MON4D3: Ouh, great catch, it's global in my config, so I didn't notice it.
TY :+1:

weilbith:
I have a custom implementation of that myself. One thing I think would be a great addition is a typical `Vim` concept where you can have either a `<filetype>.lua` file or a `<filetype>/*` directory where everything gets loaded from. I use the simple `init.lua` for this case and just `require` the filetype. Thereby it doesn't matter if it is a single file or a whole directory.
At least for me, the number of snippets for a single language quickly sum up to many lines of code. And I'm generally a fan of small/medium long files for better maintainability.

weilbith:
Are the Lua language servers good enough to understand these environments for functions? Would be ugly to have tons of undefined whatsoever diagnostics. And defining them as read-only globals for every Lua file is also not optimal.

leiserfg:
> Are the Lua language servers good enough to understand these environments for functions? Would be ugly to have tons of undefined whatsoever diagnostics. And defining them as read-only globals for every Lua file is also not optimal.

They are not. If you need that you could import it as usual.

L3MON4D3:
> I have a custom implementation of that myself. One thing I think would be a great addition is a typical `Vim` concept where you can have either a `<filetype>.lua` file or a `<filetype>/*` directory where everything gets loaded from. I use the simple `init.lua` for this case and just `require` the filetype. Thereby it doesn't matter if it is a single file or a whole directory. At least for me, the number of snippets for a single language quickly sum up to many lines of code. And I'm generally a fan of small/medium long files for better maintainability.

Allowing a `ft/`-directory inside the luasnippet-collection shouldn't be a problem, I left it out for now for simplicity. Would you open an issue for tracking that?

L3MON4D3:
> Are the Lua language servers good enough to understand these environments for functions? Would be ugly to have tons of undefined whatsoever diagnostics. And defining them as read-only globals for every Lua file is also not optimal.

Yeah, they have no chance, line setting up the environment is added at [runtime](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/loaders/from_lua.lua#L26) :D
There are supposed to be project-local settings (`.luarc.json`), but I couldn't get them to work.

weilbith:
> > I have a custom implementation of that myself. One thing I think would be a great addition is a typical `Vim` concept where you can have either a `<filetype>.lua` file or a `<filetype>/*` directory where everything gets loaded from. I use the simple `init.lua` for this case and just `require` the filetype. Thereby it doesn't matter if it is a single file or a whole directory. At least for me, the number of snippets for a single language quickly sum up to many lines of code. And I'm generally a fan of small/medium long files for better maintainability.
> 
> Allowing a `ft/`-directory inside the luasnippet-collection shouldn't be a problem, I left it out for now for simplicity. Would you open an issue for tracking that?

👍🏾  #349

weilbith:
> > Are the Lua language servers good enough to understand these environments for functions? Would be ugly to have tons of undefined whatsoever diagnostics. And defining them as read-only globals for every Lua file is also not optimal.
> 
> Yeah, they have no chance, the line setting up the environment is added at [runtime](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/loaders/from_lua.lua#L26) :D There are supposed to be project-local settings (`.luarc.json`), but I couldn't get them to work.

Okay, that is sad. But it is an opt-in feature. So even if I decide to switch to this loader here, I can still just import everything (I rename them anyways...).

PR #346, opened 2022-03-18
Update action

leiserfg:
Not sure if it will fix the issue, but just in case


L3MON4D3:
I'll take it, let's see :D

PR #348, opened 2022-03-18
add(comment): note for user for the filetype from treesitter

kunzaatko:


L3MON4D3:
That's a good tip👍
Thank you!

kunzaatko:
Also, what are your thoughts on renaming the `updateevents` in config to `update_events`? It would adhere to the other keys as they are snake case... It would be a breaking change though...

leiserfg:
I think we are almost at the point of having a 1.0 release WDYT?

L3MON4D3:
> Also, what are your thoughts on renaming the `updateevents` in config to `update_events`? It would adhere to the other keys as they are snake case... It would be a breaking change though...

Right, changing that makes a lot of sense. We can keep the old `updateevents`, and change all the documentation to `update_events`, accepting either key will be no problem I think.

> I think we are almost at the point of having a 1.0 release WDYT?

Yes! Once `add_snippets` is merged we can probably release 1.0.
For releasing, there's [`packspec`](https://github.com/nvim-lua/nvim-package-specification), but it doesn't seem to be ready yet, so for now we should probably just use tags.

PR #353, opened 2022-03-26
fix(docs): typo in demonstration of `InsertNode`

mathjiajia:
also added `The` in front of `0-th`
since a sentence starting with a number will lead to some confusion
with the period of the last sentence.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -203,11 +203,11 @@ s("trigger", {
 ```
 
 The InsertNodes are jumped over in order from `1 to n`.
-0-th node is special as it's always the last one.
+The 0-th node is special as it's always the last one.
 So the order of InsertNode jump is as follows:
-- After expansion, we will be at InsertNode 1.
-- After jumping forward, we will be at InsertNode 2.
-- After jumping forward again, we will be at InsertNode 0.
+1. After expansion, we will be at InsertNode 1.
+2. After jumping forward, we will be at InsertNode 2.
+3. After jumping forward again, we will be at InsertNode 0.
 
 If no 0-th InsertNode is found in a snippet, one is automatically inserted
 after all other nodes.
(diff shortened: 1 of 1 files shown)


DOC.md
```diff
@@ -221,8 +221,8 @@ s("trigger", {
 })
 ```
 The above snippet will behave as follows:
-- After expansion, we will be at InsertNode 1.
 - After jumping forward, we will be at InsertNode 2.
+- After expansion, we will be at InsertNode 1.
```
L3MON4D3, line 225: I don't think this should be moved down, we want to demonstrate a sequence of actions (expand -> jump -> jump), these shouldn't follow the order of the insertNodes in the snippet.

Would you make the list numbered to express that more clearly?

DOC.md
```diff
@@ -286,7 +286,7 @@ The first parameter of `f` is the function. Its parameters are:
       parent (a `snippetNode`) will contain neither `captures` nor `env`. Those
       are only stored in the `snippet`, which can be accessed as `parent.snippet`.
 
-3-n. The `user_args` passed in `opts`.
+3. The `user_args` passed in `opts`.
```
L3MON4D3, line 289: Oh, this isn't a typo, there might be multiple user_args, hence the `n`😅
If you want, you can mention that there may be multiple user_args, I think that might make this more clear.

mathjiajia: Thanks for the explaination.
I will revise this.

DOC.md
mathjiajia: I changed the order because, after expansion the snippets gives
```
After jumping forward, we will be at ->
After expansion, we will be at ->
After jumping forward again, we will be at ->
```
Should we follow this order?

L3MON4D3: No, I don't think so, the list should show where the cursor is after a given action, the first action is `expand`, followed by two jumps.

It makes more sense to have the list ordered by the actions than by the position of the lines in the snippet I think.

L3MON4D3:
Nice, TY!
Could you squash all of those into a single commit? I think that would be a bit cleaner.

mathjiajia:
> 

Sure, I will do that.
It is ugly now.

L3MON4D3:
Nice, ty!
Definitely an improvement :D

PR #354, opened 2022-03-26
Make scandir resolve relative symlinks

itepechi:
This PR allows LuaSnip to work smoothly with GNU Stow, which only creates relative symlinks.

### DIFF
--- a/lua/luasnip/util/path.lua
+++ b/lua/luasnip/util/path.lua
@@ -69,27 +69,6 @@ function Path.expand(filepath)
 	return uv.fs_realpath(expanded)
 end
 
----Resolve a (chain of) symbolic link(s) to their final destination.
----@param path string
----@return string|nil
----Returns nil if the sym link points to an inexistent file.
----Assumes that path is a symlink, will return nil otherwise.
-local function resolve_symlink(path)
-	while true do
-		local followed_path = uv.fs_readlink(path)
-		if followed_path then
-			local stat = uv.fs_stat(followed_path)
-			if stat and stat.type ~= "link" then
-				return followed_path, stat.type
-			else
-				path = followed_path
-			end
-		else
-			return nil
-		end
-	end
-end
-
 ---Return files and directories in path as a list
 ---@param root string
 ---@return string[] files, string[] directories
@@ -106,13 +85,13 @@ function Path.scandir(root)
 			elseif type == "directory" then
 				table.insert(dirs, path)
 			elseif type == "link" then
-				local followed_path
-				followed_path, type = resolve_symlink(path)
+				local followed_path = uv.fs_realpath(path)
 				if followed_path then
-					if type == "file" then
-						table.insert(files, followed_path)
-					elseif type == "directory" then
-						table.insert(dirs, followed_path)
+					local stat = uv.fs_stat(followed_path)
+					if stat.type == "file" then
+						table.insert(files, path)
+					elseif stat.type == "directory" then
+						table.insert(dirs, path)
 					end
 				end
 			end


L3MON4D3:
Nice! I wasn't aware that readlink doesn't resolve relative links, but it's kind of in the name, so makes sense.

While looking into the doc for `luv`, I found `uv.fs_realpath` (which we actually use at another point), which could resolve links (including relative ones) for us.

While I appreciate the work that has gone into this patch, I think just using `realpath` is the better option when it comes to maintainability.
Would you look into using that to resolve symlinks?

itepechi:
I don't know if we should care about it, but fs_realpath has some platform-specific caveats -> https://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_realpath

L3MON4D3:
I think we can ignore those :sweat_smile:
If `fs_realpath` messes up in some way we can look into rolling our own again, but I don't think it will.

itepechi:
Removed `resolve_symlink` as there is no way to access it from the outside of the module.
Please let me know if this is too much.

L3MON4D3:
That's alright :+1:

There's one potential problem though: we resolve the filetype a given file provides snippets for from the name of the real file, not from the name of the link, which could lead to subtle problems (I tested your patch by linking to a file just named `luasnippets` from `all.lua`, but `luasnippets` doesn't match the `<filetype>.lua`-pattern, and therefore no `all`-snippets were loaded).

Would you like to improve that?
IMO the simplest fix would be to return lists of `{ft, path}`-tables from `scandir` instead of just lists of file/directorynames.

L3MON4D3:
OTOH, the simplest option could almost be to only resolve symlinks before actually reading the files/directories.

lua/luasnip/util/path.lua
```diff
@@ -106,12 +94,12 @@ function Path.scandir(root)
 			elseif type == "directory" then
 				table.insert(dirs, path)
 			elseif type == "link" then
-				local followed_path
-				followed_path, type = resolve_symlink(path)
+				local followed_path = uv.fs_realpath(path)
+				local stat = uv.fs_stat(path)
```
L3MON4D3, line 98: This should be `followed_path` instead of `path`, right?

itepechi: My bad.

itepechi:
Tested with

- `/Users/itepechi/.config/nvim/luasnippets/markdown.lua@ -> /Users/itepechi/Documents/git/dotfiles/nvim/.config/nvim/luasnippets/markdown`
- `/Users/itepechi/.config/nvim/luasnippets/markdown.lua@ -> ../../../Documents/git/dotfiles/nvim/.config/nvim/luasnippets/markdown.lua`
- `/Users/itepechi/.config/nvim/luasnippets/markdown.lua@ -> /Users/itepechi/Documents/git/dotfiles/nvim/.config/nvim/luasnippets/markdown.lua`

and it worked well.
Thank you for the suggestion.

L3MON4D3:
Nice, looks good.
Could you try if it works with linked directories, like snippets/markdown/file.lua where markdown is a symbolic link?
scandir might not resolve symlinks before iterating.

itepechi:
I am not sure if I understand it correctly, but here is the test I did:

With symlink `/Users/itepechi/.config/nvim/luasnippets/markdown@ -> /Users/itepechi/Documents/git/dotfiles/nvim/.config/nvim/luasnippets/markdown`

```console
$ ls -l /Users/itepechi/Documents/git/dotfiles/nvim/.config/nvim/luasnippets/markdown
total 8
-rw-r--r--  1 some  staff  122 Mar 26 02:08 code_block.lua
```

```lua
:LuaSnipListAvailable
{
  all = {},
  markdown = { {
      description = { "code" },
      name = "code",
      regTrig = false,
      trigger = "code",
      wordTrig = true
    } }
}
```

L3MON4D3:
Ouh, indeed it does, nice.

I'm not sure if we should append the realpath: the filenames created there are used for 
1. reading the content of the file (where links are resolved in [`fs_open`](https://github.com/L3MON4D3/LuaSnip/blob/4b601dc956c2b54ab5dd07ac071d9575474323d3/lua/luasnip/loaders/from_lua.lua#L17) it seems). But it could actually be problematic to use realpath: one file could be loaded for two filetypes (ft1 then ft2), which would mean that the snippets loaded for ft1 are removed when the ones for ft2 are added ([ft2 would reuse ft1's key](https://github.com/L3MON4D3/LuaSnip/blob/4b601dc956c2b54ab5dd07ac071d9575474323d3/lua/luasnip/loaders/from_lua.lua#L57))
2. in [edit_snippet_files](https://github.com/L3MON4D3/LuaSnip/blob/4b601dc956c2b54ab5dd07ac071d9575474323d3/lua/luasnip/loaders/from_lua.lua#L176), which opens a `vim.ui.select` and prompts to open some file which contained snippets. This would currently show realpath, but I think it might be nicer to show the path containing the link instead.

What do you think?

itepechi:
I'm trying to keep both paths as a table like `{ path = path, realpath = realpath }`, but it seems like it will take some time to rewrite things.

itepechi:
Wait, i came up with a better idea. If we lookup realpaths from the `Cache`, there's no need to rewrite the codebase massively, or access the filesystem hundreds of times with fs_realpath. I don't know why i didn't think of that.

itepechi:
Sorry for the mess.

- Tested the loading with several types of symlinks:
  ```
  html.lua@ -> /Users/itepechi/Documents/git/dotfiles/nvim/.config/nvim/luasnippets/html
  lua.lua@ -> ../../../Documents/git/dotfiles/nvim/.config/nvim/luasnippets/lua.lua
  markdown@ -> /Users/itepechi/Documents/git/dotfiles/nvim/.config/nvim/luasnippets/markdown
  xhtml.lua@ -> /Users/itepechi/Documents/git/dotfiles/nvim/.config/nvim/luasnippets/html
  ```
- `:lua require("luasnip.loaders.from_lua").edit_snippet_files()` shows the path containing the link
- Using the path containing the link when generating keys for snippets

L3MON4D3:
Are you sure we need to store the realpath? 😅
I'm pretty sure both fs_open and edit work with links as well.

itepechi:
Oh, I can confirm that.

L3MON4D3:
Nice :D

Thank you (especially for your patience :D), this is a good addition, it really simplifies that code :+1:

PR #355, opened 2022-03-26
Update example for bash.

xiaket:
Since we have updated function signature in 48a79b7f.


L3MON4D3:
Ouh, I must've missed that when converting the documentation after that change.
TY!

PR #368, opened 2022-04-01
adding additional information about the path for `from_lua`

TornaxO7:
I don't know if this is really necessary but this would had helped me to understand the `./` when I read the example.

### DIFF
--- a/Examples/snippets.lua
+++ b/Examples/snippets.lua
@@ -542,7 +542,12 @@ ls.filetype_extend("all", { "_" })
 
 require("luasnip.loaders.from_snipmate").load({ include = { "c" } }) -- Load only snippets for c.
 
-require("luasnip.loaders.from_snipmate").load({ path = { "./my-snippets" } }) -- Load snippets from my-snippets folder
+-- Load snippets from my-snippets folder
+-- The "." refers to the directory where of your `$MYVIMRC` (you can print it
+-- out with `:lua print(vim.env.MYVIMRC)`.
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Right my bad, it's the directory containing that file: https://github.com/L3MON4D3/LuaSnip/blob/9ecb715cf5480b144af2043753abbc196c4665dd/lua/luasnip/util/path.lua#L54

L3MON4D3:
One last thing, Could you squash the commits?

TornaxO7:
> Right my bad, it's the directory containing that file:
> 
> https://github.com/L3MON4D3/LuaSnip/blob/9ecb715cf5480b144af2043753abbc196c4665dd/lua/luasnip/util/path.lua#L54

Eh, so should I add a link to this line? I'm sorry that I don't really understand what you mean with this ;-;

TornaxO7:
> One last thing, Could you squash the commits?

Done!

TornaxO7:
Eh... I'm a little bit irritated what I should do here now (sorry). Could you please explain again what I should do?

L3MON4D3:
No problem: There are cases where `$MYVIMRC` is not set at all (ie. when nvim is started via `nvim -u somefile`, as the vimrc won't be loaded).
We should probably at least mention that this might happen, so maybe just add something like "if neovim is started with nvim -u, $MYVIMRC is not set".

TornaxO7:
Done!

L3MON4D3:
`yeet.txt`, ofc :P
Nice, thank you! :)

PR #371, opened 2022-04-02
fix: Skip "order" key in snippets cleanup

Mike325:


L3MON4D3:
Oh, yeah that couldn't work😅

lua/luasnip/session/snippet_collection.lua
```diff
@@ -172,8 +172,10 @@ function M.clean_invalidated(opts)
 
 	-- remove invalidated snippets from all tables.
 	for _, type_snippets in pairs(by_prio) do
-		for _, prio_snippets in pairs(type_snippets) do
-			prio_snippets = without_invalidated(prio_snippets)
+		for priority, prio_snippets in pairs(type_snippets) do
+			if priority ~= 'order' then
+				prio_snippets = without_invalidated(prio_snippets)
```
L3MON4D3, line 177: This should also be `by_prio[priority] = ...`.
Imo key instead of priority would be a bit better, what do you think?

Mike325: that actually sounds better

L3MON4D3: My bad, I didn't look properly, it should be `type_snippets[key] = ...` ofc :facepalm:

L3MON4D3:
Yup, thanks a lot!!

PR #378, opened 2022-04-09
feat(API): add emmylua annotations

kylo252:
I think it might be better to let the API documentation live inside the code, and it also benefits from LSP capabilities

image

---

**TODO**
- [ ]  I've only added a few so far
- [ ] autocompletion seems busted, I have a feeling it's something to do with some internal cache invalidation

### DIFF
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -8,6 +8,11 @@ local next_expand_params = nil
 local ls
 local luasnip_data_dir = vim.fn.stdpath("cache") .. "/luasnip"
 
+---@alias direction
+---| '-1' # for previous
+---| '1' # for next
+
+---returns the currently active snippet (not node!).
 local function get_active_snip()
 	local node = session.current_nodes[vim.api.nvim_get_current_buf()]
 	if not node then
@@ -104,6 +109,10 @@ local function safe_jump(node, dir, no_move)
 		end
 	end
 end
+
+---check if the jump was successful.
+---@param dir direction
+---@return boolean
 local function jump(dir)
 	local current = session.current_nodes[vim.api.nvim_get_current_buf()]
 	if current then
@@ -119,11 +128,16 @@ local function jump(dir)
 	end
 end
 
+---check if it's possible to jump forward or backward to another node.
+---@param dir direction
+---@return boolean
 local function jumpable(dir)
 	local node = session.current_nodes[vim.api.nvim_get_current_buf()]
 	return (node ~= nil and node:jumpable(dir))
 end
 
+---check if a snippet can be expanded at the current cursor position.
+---@return boolean
 local function expandable()
 	next_expand, next_expand_params = match_snippet(
 		util.get_current_line_to_cursor(),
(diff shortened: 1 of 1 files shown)


L3MON4D3:
I agree, it makes sense to have the documentation next to the code, should prevent it becoming outdated.

It would be really cool to automatically generate the entries in `DOC.md` (because I'd like to keep those as well) from these annotations, do you know if there are ways to do that?

kylo252:
> I agree, it makes sense to have the documentation next to the code, should prevent it becoming outdated.
> 
> It would be really cool to automatically generate the entries in `DOC.md` (because I'd like to keep those as well) from these annotations, do you know if there are ways to do that?

I was looking earlier at `ldoc` which is able to generate html, and in theory markdown as well, but I could never get it to work with markdown. I think it's missing a template or something. 

There's also this script from neovim-core, https://github.com/neovim/neovim/blob/8055f9857b8e384634d457533dfdb08618fc36f0/scripts/gen_vimdoc.py#L2

Lastly, I had an idea to leverage LSP for this, but it's a bit fiddley (redirect `hover` request output)
```lua
  local write_tmpfile = function(data)
    local tmpname = os.tmpname()
    local file = io.open(tmpname, "w")
    if type(data) == "table" then
      data = table.concat(data, "\n")
    end
    file:write(data)
    file:close()
    return tmpname
  end
  local util = require "vim.lsp.util"
  vim.lsp.handlers["textDocument/hover"] = function(_, result, ctx, config)
    config = config or {}
    config.focus_id = ctx.method
    if not (result and result.contents) then
      -- return { 'No information available' }
      return
    end
    local markdown_lines = util.convert_input_to_markdown_lines(result.contents)
    markdown_lines = util.trim_empty_lines(markdown_lines)
    if vim.tbl_isempty(markdown_lines) then
      -- return { 'No information available' }
      return
    end
    local tmpfile = write_tmpfile(markdown_lines) -- can also just copy the lines into some register instead
    print(tmpfile)
    vim.cmd([[edit ]] .. tmpfile)
    return tmpfile
  end
```

L3MON4D3:
LDoc looks promising, but I'm surprised that there is no tool that straightforwardly generates markdown
The LSP approach is interesting, is there a way to automate it?

kylo252:
> is there a way to automate it?

Here's my script so far, just give it a valid bufnr (check `:LspInfo`)
https://gist.github.com/kylo252/7b3edb23b55197316a33916fd79d942f

Here are the results, which as you can see need better formatting, but that's why I included the raw response as well
https://gist.github.com/kylo252/1dcc3f9da60e42786d7088caa8441f67

I'm also counting on the fact that you have more experience with string manipulation 😃

L3MON4D3:
Oh wow, that looks pretty good already :+1::+1:
I won't be able to really get into helping you with it for some time though :/

saadparwaiz1:
There's also this [treesitter](https://github.com/tjdevries/tree-sitter-lua) module available for generating documentation thorough Emmy

L3MON4D3:
Interesting, but it only generates vimdoc (afaict) :/

saadparwaiz1:
seems like [this](https://github.com/boolangery/sphinx-lua) does the job? spinx generates rst by default but that works on github. if markdown is required this [extension](https://github.com/clayrisser/sphinx-markdown-builder) can be used?

hinell:
Wonder why this one wasn't merged. Nice PR. Many projects are underdocumented.

L3MON4D3:
Hey @kylo252, are you fine with merging this as-is (or, more-or-less as-is, after looking into how up to date it still is)?

PR #380, opened 2022-04-09
Fix `luasnip.loaders.from_lua` documentation

9999years:
The example snippet files use a global variable `ls` which isn't defined in the `luasnip.config.snip_env`. (Due to circular dependency issues? Just speculation.)


L3MON4D3:
Ah right, that's an oversight😅
And true, `ls` can't be defined directly due to circular dependencies, but these globals (for the snippet-files only) should make it so accessing them through `ls` isn't necessary.

LGTM, Ty! :)

Issue #372, opened 2022-04-03
Add autotrigger to loaders/from_snipmate


nullchilly:
[loaders/from_vscode](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/loaders/from_vscode.lua) supports autotrigger. 

[loaders/from_snipmate](https://github.com/L3MON4D3/LuaSnip/blob/master/lua/luasnip/loaders/from_snipmate.lua) does not. Can we adopt [UltiSnips.txt#Autotrigger](https://github.com/SirVer/ultisnips/blob/master/doc/UltiSnips.txt#L1732)?

I can't find any documentation on this? It would be nice to see actual examples about autosnippets in Examples/

Pardon me for my ignorance!

L3MON4D3:
Yep, there is no Autotrigger-support for snipmate snippets.
It'd be nice to have, I agree, but directly using the ultisnips-format, eg postfix a, won't be possible because there we have the docstrings, which isn't enclosed in `""`.

Not sure how to best pass options, maybe specific keywords that only affect the next snippet?

tuwuna:
> Not sure how to best pass options, maybe specific keywords that only affect the next snippet?

Maybe this wouldn't hurt?

```
if vim.startswith(line, "snippet") then
	_parse()
elseif vim.startswith(line, "autosnippet") then
	_parse(1)
```

Maybe consider  `loaders/from_ultisnips` in the future with condition support

L3MON4D3:
Oh yeah, that looks pretty nice.

Conditions may be doable, I'll think about it.

Loading ultisnips-snippets would be pretty nice, but we won't be able to do the python-stuff (not without great effort, at least).
There also is https://github.com/smjonas/snippet-converter.nvim for converting snippets between different formats.

tuwuna:
I tried to do add this but it seems like I have to add the auto_snippets table everywhere I go, also not exactly sure how the cache works.

Low on time so I am using `ls.add_snippets(ft, snippets, { type = "autosnippets"})` instead...

@uga-rosa would you like to take a look? Sorry for my impatient but I think it would be easier for you to add this 😅

uga-rosa:
Sorry, I'm not feeling well right now and will be delayed in responding.

leiserfg:
> Sorry, I'm not feeling well right now and will be delayed in responding.

Hope it's nothing serious.

L3MON4D3:
[...]

PR #390, opened 2022-04-16
Enhance snipmate loader.

L3MON4D3:
Includes small improvements:
* correctly handle multiple `lazy_load`-calls (previously only the last is stored)
* adding autosnippets is now possible via `autosnippet`-keyword instead of `snippet` (as proposed in #372)

Should be functional, still needs docs+tests.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -983,6 +983,8 @@ Here is a summary of the differences from the original snipmate format.
 - `${VISUAL}` will be replaced by `$TM_SELECTED_TEXT` to make the snippets
 compatible with luasnip
 - We do not implement eval using \` (backtick). This may be implemented in the future.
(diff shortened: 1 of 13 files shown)


L3MON4D3:
@uga-rosa If you have time, could you take a look to make sure this doesn't break any previous behaviour?

Doesn't need to be soon though, no reason this cannot stay as a PR for a while.

uga-rosa:
I'm feeling better and I've read through the whole thing and I think it's probably fine.

L3MON4D3:
Nice, that's good to hear (both :D)
I'll see when I get to writing some tests for this, shouldn't be too long

L3MON4D3:
Merged but I rebased :|

PR #399, opened 2022-04-24
Expose plugin hint for vim.ui.select

Weypare:
Useful for custom `vim.ui.select` handlers.


L3MON4D3:
Makes sense, thank you👍

PR #408, opened 2022-04-29
snip_expand: add option to prevent jump into snippet.

L3MON4D3:
This can be useful when luasnip is used as part of a bigger workflow and the mode shouldn't change.

Example:
```lua
ls.snip_expand(
    ls.parser.parse_snippet("testtest", "works$1works"),
    { skip_snippet = true })
```

### DIFF
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -154,6 +154,10 @@ local function expand_or_locally_jumpable()
 	return expandable() or (in_snippet() and jumpable())
 end
 
+local function _jump_into_default(snippet)
+	return util.no_region_check_wrap(snippet.jump_into, snippet, 1)
+end
+
 -- opts.clear_region: table, keys `from` and `to`, both (0,0)-indexed.
 local function snip_expand(snippet, opts)
 	local snip = snippet:copy()
@@ -162,6 +166,7 @@ local function snip_expand(snippet, opts)
 	opts.expand_params = opts.expand_params or {}
 	-- override with current position if none given.
 	opts.pos = opts.pos or util.get_cursor_0ind()
+	opts.jump_into_func = opts.jump_into_func or _jump_into_default
 
 	snip.trigger = opts.expand_params.trigger or snip.trigger
 	snip.captures = opts.expand_params.captures or {}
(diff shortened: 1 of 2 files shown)


kylo252:
works for me! 🚀
the only thing I'm unsure about is the option name, I'm not sure `skip_snippet` is obvious.

lua/luasnip/init.lua
```diff
@@ -202,12 +202,20 @@ local function snip_expand(snippet, opts)
 		end
 	end
 
-	session.current_nodes[vim.api.nvim_get_current_buf()] =
-		util.no_region_check_wrap(
-			snip.jump_into,
-			snip,
-			1
-		)
+	if not opts.skip_snippet then
```
kylo252, line 205:
```suggestion
	if snip:is_interactive() then
```
this seems to work for me, any drawbacks? wdyt?

kylo252: we can probably make more robust by adding a fallback condition (also) 
```lua
  if snip:is_interactive() or vim.api.nvim_get_mode() == "i" then
```
basically jump if you're already in insert mode just in case, since it's highly unlikely that you'd want to expand a snippet in insert mode and not have the cursor as well.

L3MON4D3: IMO this is pretty specific and I wouldn't really want to include it (we'd also have to disable it for triggered expansion where you'd want your cursor to be at the position of the `$0` every time)
What do you think of accepting a callback responsible for jumping into the snippet instead? Then you could get as fancy as you like :D

kylo252: hmm, something is up with`vim.api.nvim_get_mode()` but `vim.fn.mode()` works fine

L3MON4D3: Getting the mode is a bit finnicky :/
I think we also had some problems with using api for it and use `vim.fn.mode` when we need the current mode.

kylo252: > What do you think of accepting a callback responsible for jumping into the snippet instead?

so my original approach was trying to check `jumpable()` before attempting to call `jump_into`, but I wasn't able to figure out how to do that, since `jumpable()` is accessed by the node and not the snippet, and I'm not sure which _node_ it would be in this case 😅

L3MON4D3: You can get the nodes inside a snippets either as `snippet.nodes`, for the nodes in the order they appear in the buffer, or `snippet.insert_nodes`, where the keys are the position of the node (ie. `insert_nodes[1]` ->`i(1)`, `insert_nodes[0]` -> `i(0)`, `insert_nodes[2]` -> `c(2, {...})`)

Not sure if `jumpable` really is what you want here, `is_interactive` is the right approach I think.

L3MON4D3:
I added the callback for expanding, you can now use something like this:
```lua
ls.snip_expand(parse("", "asdf$1asdf"), {
	jump_into_func = function(snip)
		-- callback has to return the new current node.
		return snip.insert_nodes[0]
	end })
```

L3MON4D3:
(name is debatable of course :P)

PR #409, opened 2022-04-29
Improve loaders

L3MON4D3:
Pass any options to loaders through `(lazy_)load`-call.
Not all options are supported, `override_priority` and `default_priority` work, the others might be overriden.

Not sure if we should use this implementation, which just passes through the options to `add_opts` or if we should expose select (`override_priority` and `default_priority`) options in the `load` instead.
```lua
require("luasnip.loaders.from_snipmate").load({add_opts = {override_priority = 2000}})
-- vs
require("luasnip.loaders.from_snipmate").load({override_priority = 2000})
```
Having it named `add_opts` suggests that all opts will have an effect, which
1. depends on the loader (`from_snipmate` overrides `type`, `from_lua` overrides `type` and `key`)
2. doesn't really make sense: either the other currently options are controlled by the loader, or they might lead to detrimental behaviour (if `key` is passed to `from_snipmate`, only the last `add_snippets` with `key` will be visible)

Maybe we should just expose the valid options directly instead, we'll have to do some modifications to expose new ones should they become available in `add_snippets`, but it seems to be the better option (now, after implementing a first version :D)

- [x] snipmate override/default
- [x] lua override/default
- [x] vscode override/default
- [x] vscode priority per-snippet (pass in `snippet->luasnip`-json-object)
- [x] snipmate priority per-snippet in the line immediately before snippet-definition? Like
  ```snippets
  priority 2000
  snippet trigger a description
  	snippet body
  ```

Edit 1:
Implementing that stuff also (kind of) required doing some general improvements, which now also make adding reload-capability to all loaders much easier. I'll just add that onto this PR.

Edit 2:
The syntax is now the second version, eg `{override,default}_priority` can be passed directly in the table passed to load.
All that needs doing now is updating the docs.

### DIFF
--- a/lua/luasnip/loaders/from_snipmate.lua
+++ b/lua/luasnip/loaders/from_snipmate.lua
@@ -83,15 +83,15 @@ local function load_snippet_files(add_ft, paths, collection_files, add_opts)
 		local snippet, autosnippet, extends
 
 		if cache.path_snippets[path] then
-			snippet = cache.path_snippets[path].snippet
-			autosnippet = cache.path_snippets[path].autosnippet
+			snippet = vim.deepcopy(cache.path_snippets[path].snippet)
+			autosnippet = vim.deepcopy(cache.path_snippets[path].autosnippet)
 			extends = cache.path_snippets[path].extends
 		else
 			local buffer = Path.read_file(path)
 			snippet, autosnippet, extends = parse_snipmate(buffer, path)
 			cache.path_snippets[path] = {
-				snippet = snippet,
-				autosnippet = autosnippet,
+				snippet = vim.deepcopy(snippet),
+				autosnippet = vim.deepcopy(autosnippet),
 				extends = extends,
 				-- store for reload.
 				add_opts = add_opts,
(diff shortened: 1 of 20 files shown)


L3MON4D3:
@leiserfg could you take a look at the changes in 31fe9d9 and a9efee6?
friendly-snippets works, but I'm not sure that there aren't some edge cases that don't :sweat_smile:

lua/luasnip/loaders/from_snipmate.lua
```diff
@@ -208,6 +208,10 @@ function M.lazy_load(opts)
 	end
 end
 
+function M.edit_snippet_files()
+	loader_util.edit_snippet_files(cache.ft_paths)
+end
```
L3MON4D3, line 213: I'd like to provide a better default here.
It'd be great to substitute the paths to packer with something shorter.
Is there a env-var that is set to the path packer uses?
I use a simple 
```lua
function(path, _)
	local clean_path, _ = path:gsub("/home/simon/.local/share/nvim/site/pack/packer/start", "$PACKER"):gsub("/home/simon/.config/nvim", "$NVIM")
	return clean_path
end
```
which would be a nicer default.

L3MON4D3: Added a better default in 680e559

L3MON4D3:
This is merged, just witout 4c7351e, so github can't tell :|

PR #418, opened 2022-05-04
Fix `ls.cleanup`.

L3MON4D3:
We don't clean the loaders' caches at the moment, this should make it work.


L3MON4D3:
Merged in cc00863

PR #438, opened 2022-05-22
Implement pre-expand-callback.

L3MON4D3:
This adds a pre-expand-callback, which is executed before any text is inserted into the buffer.
Inside it, it's okay to remove/insert text (we keep track of the snippet's expand-position using an extmark).

One example where this is useful: a snippet that always expands without indent.
```lua
s("trig36",
	t{"", "no indent"},
	{
		callbacks = {
			-- event for snippet.
			[-1] = {
				[events.pre_expand] = function(_, opts)
					local pos = opts.expand_pos
					-- simply remove all text before the snippet's expand-position.
					vim.api.nvim_buf_set_text(0, pos[1], 0, pos[1], pos[2], {""})
				end
			}
		}
	}
)
```

Another application would be logging expansions (seems useless though)
```vim
au User LuasnipPreExpand lua print("expanding " .. table.concat(ls.session.event_node:get_docstring(), "\n") )
```

I'll have to think about whether this really is the best way to handle pre-expand-callbacks, then add some docs.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -141,14 +141,16 @@ The third argument (`opts`) is a table with the following valid keys:
   should be set as follows:
   ```lua
   {
+  	-- position of the node, not the jump-position!!
+  	-- s("trig", {t"first node", t"second node", i(1, "third node")}).
   	[2] = {
-  		[events.enter] = function(node) print("2!") end
+  		[events.enter] = function(node, _event_args) print("2!") end
   	}
   }
   ```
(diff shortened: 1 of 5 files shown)


ranebrown:
Seems to be working for my use case. Let me know if I can help with any more testing.

L3MON4D3:
Just use it :P
If you have any weird ideas for what might be done with `pre_expand`-callbacks, try them, but I don't have anything specific that needs testing.

Issue #410, opened 2022-04-30
ext_opt Additional Table For Unvisited Nodes


p0da:
Would you consider adding adding a ext_opt table for nodes which are unvisited? There does not  seem to be a way to specify the application of a highlight group to exclusively unvisited nodes with the currently available tables: passive, active, and snippet_passive. Therefore it would be helpful if there was an additional table which applied attributes exclusively to nodes that have not yet been jumped to. Thanks!

L3MON4D3:
Oh that's a nice idea, I'm definitely not opposed.
Should also be rather easy to implement :+1:

L3MON4D3:
One problem: with `active`, `passive`, `snippet_passive` it's very clear when any of these highlights should be applied. Should `unvisited` only show while the snippet is active?
Eg. `snippet_passive` ->(snippet is jumped into) `unvisited` -> (snippet is exited again) `snippet_passive` or `unvisited`?
IMO it should be `snippet_passive`, but that isn't communicated directly by `unvisited`

p0da:
I'd consider it most intuitive if unvisited was applied only when the snippet is active. But yeah the name "unvisited" doesn't really communicate that, maybe something like "active_unvisited"? Really anything you choose would be fine with me.

L3MON4D3:
Okay, that's good, I wanted to make sure you didn't have some niche usecase that wouldn't play well with what I had in mind.

![ext_opts_hierarchy](https://user-images.githubusercontent.com/41961280/166138092-b9be3f22-eb4b-41fa-8361-2e6b75ade37f.svg)
I'm thinking of adding `visited` as well, then we could have the above inheritance for `ext_opts` and, during runtime, use either `visited` or `unvisited` for a given node (feels cleaner than `passive` and `unvisited`).

p0da:
Looks good. Can't wait to use it.

L3MON4D3:
Finally got to it, check #440 :)

L3MON4D3:
Oh, it's merged now, closing this.

p0da:
Looks good, thanks for your work.

PR #440, opened 2022-05-24
`ext opts` for (un)visited nodes.

L3MON4D3:
As proposed in #410.
For now the inheritance is (with arrows `to` _gets-and-extends-ext_opts-from_ `from`)
```mermaid
flowchart TD
	passive --> active
	passive --> visited
	passive --> unvisited
	snippet_passive --> passive
```
although that feels a bit weird, maybe it would be better to combine `ext_opts` from `visited` and `unvisited` for `active`.

One example:
```lua
ls.setup_snip_env()

local node_ext_opts = {
	visited = {
		hl_group = "GruvboxPurple"
	},
	unvisited = {
		hl_group = "GruvboxBlue"
	},
}

ls.add_snippets("all", {
	s("trig36", {
		t{"no indent"}, c(3, {
			t"a",
			t"b",
			t"c"
		}, {
			node_ext_opts = node_ext_opts
		}), i(2, "some text"), i(1, "some text")
	}, {
		child_ext_opts = {
			[types.insertNode] = {
				passive = {
					hl_group = "GruvboxRed"
				},
				visited = {
					hl_group = "GruvboxOrange"
				},
				unvisited = {
					hl_group = "GruvboxYellow"
				},
			}
		}
	})
}, {
	key = "test"
})
```

Needs docs and tests.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1861,6 +1861,13 @@ local ext_opts = {
 		-- add virtual text on the line of the node, behind all text.
 		virt_text = {{"virtual text!!", "GruvboxBlue"}}
 	},
+	-- visited or unvisited are applied when a node was/was not jumped into.
+	visited = {
+		hl_group = "GruvboxBlue"
+	},
+	unvisited = {
+		hl_group = "GruvboxGreen"
+	},
 	-- and these are applied when both the node and the snippet are inactive.
 	snippet_passive = {}
 }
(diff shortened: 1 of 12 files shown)


L3MON4D3:
Starting to look into this again, I think I'm going to change the dependencies/extensions from
```mermaid
flowchart TD
	passive --> active
	passive --> visited
	passive --> unvisited
	snippet_passive --> passive
```
to
```mermaid
flowchart TD
	visited --> active
	passive --> visited
	passive --> unvisited
	snippet_passive --> passive
```
The problem of the first is actually visible in the first example:
`active` inherits from `passive`, so if a visited node is active, it'll be highlighted with `passive`, even though it already was/is in a more specific state (`visited`) (a visited node returning to its passive highlight feels weird)
Inheriting from visited in active also makes sense since active is just a special case of visited.

Issue #441, opened 2022-05-25
**Illegal back reference** Error when lazyloading from_vscode.lua


konosubakonoakua:
## Error
```
Error detected while processing C:\Users\13567\AppData\Local\nvim\init.lua:
E5113: Error while calling lua chunk: ...packer\start\LuaSnip/lua/luasnip/loaders/from_vscode.lua:81: Vim(autocmd):E65: Illegal back reference
stack traceback:
        [C]: in function 'cmd'
        ...packer\start\LuaSnip/lua/luasnip/loaders/from_vscode.lua:81: in function 'load_snippet_files'
        ...packer\start\LuaSnip/lua/luasnip/loaders/from_vscode.lua:244: in function 'lazy_load'
        C:\Users\13567\AppData\Local\nvim/lua/user/completions.lua:11: in main chunk
        [C]: in function 'require'
        C:\Users\13567\AppData\Local\nvim\init.lua:5: in main chunk
```
### cmp.lua
```lua
local cmp_status_ok, cmp = pcall(require, "cmp")
if not cmp_status_ok then
  return
end

local snip_status_ok, luasnip = pcall(require, "luasnip")
if not snip_status_ok then
  return
end

require("luasnip/loaders/from_vscode").lazy_load()

local check_backspace = function()
  local col = vim.fn.col "." - 1
  return col == 0 or vim.fn.getline("."):sub(col, col):match "%s"
end

--   פּ ﯟ   some other good icons
local kind_icons = {
  Text = "",
  Method = "m",
  Function = "",
  Constructor = "",
  Field = "",
  Variable = "",
  Class = "",
  Interface = "",
  Module = "",
  Property = "",
  Unit = "",
  Value = "",
  Enum = "",
  Keyword = "",
  Snippet = "",
  Color = "",
  File = "",
  Reference = "",
  Folder = "",
  EnumMember = "",
  Constant = "",
  Struct = "",
  Event = "",
  Operator = "",
  TypeParameter = "",
}
-- find more here: https://www.nerdfonts.com/cheat-sheet

cmp.setup {
  snippet = {
    expand = function(args)
      luasnip.lsp_expand(args.body) -- For `luasnip` users.
    end,
  },
  mapping = {
    ["<C-k>"] = cmp.mapping.select_prev_item(),
		["<C-j>"] = cmp.mapping.select_next_item(),
[...]

PR #443, opened 2022-05-25
 properly escape string filename including backslashes

konosubakonoakua:
Solve issue [**Illegal back reference** Error when lazyloading from_vscode.lua](https://github.com/L3MON4D3/LuaSnip/issues/441)

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -20,6 +20,7 @@ Table of Contents                                  *luasnip-table-of-contents*
   - FMT                                                          |luasnip-fmt|
   - On The Fly snippets                          |luasnip-on-the-fly-snippets|
   - Select_choice                                      |luasnip-select_choice|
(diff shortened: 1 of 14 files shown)


L3MON4D3:
Oh, is `\` not escaped on windows? I thought that exactly this was the purpose of `fnameescape`..

konosubakonoakua:
> Oh, is `\` not escaped on windows? I thought that exactly this was the purpose of `fnameescape`..

 I printed escaped result returned by `fnameescape`. '\' was escaped but it still run into same error.
It seems that only `\\` works on windows

konosubakonoakua:
> After reading through `:h file-pattern`, I think we might also get by by replacing `\` with `/` (and then `fnameescape`ing). Could you try if that also works?

Hi，replacing `\` with `/` also works, I'll create a new PR

PR #444, opened 2022-05-25
properly handle string filename including backslashes

konosubakonoakua:
Solve issue [**Illegal back reference** Error when lazyloading from_vscode.lua](https://github.com/L3MON4D3/LuaSnip/issues/441) by replacing `\\` with `/`

### DIFF
--- a/lua/luasnip/util/str.lua
+++ b/lua/luasnip/util/str.lua
@@ -20,7 +20,7 @@ end
 
 function M.aupatescape(s)
 	local comma_escaped = s:gsub(",", "\\,")
-    local backslash_replaced = comma_escaped:gsub("\\", "/")
+    	local backslash_replaced = comma_escaped:gsub("\\", "/")
 	return vim.fn.fnameescape(backslash_replaced)
 end
 
--- a/lua/luasnip/util/str.lua
+++ b/lua/luasnip/util/str.lua
@@ -20,7 +20,7 @@ end
 
 function M.aupatescape(s)
 	local comma_escaped = s:gsub(",", "\\,")
-    	local backslash_replaced = comma_escaped:gsub("\\", "/")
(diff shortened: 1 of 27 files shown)


lua/luasnip/util/str.lua
```diff
@@ -20,7 +20,7 @@ end
 
 function M.aupatescape(s)
 	local comma_escaped = s:gsub(",", "\\,")
-    	local backslash_replaced = comma_escaped:gsub("\\", "/")
+  local backslash_replaced = comma_escaped:gsub("\\", "/")
```
konosubakonoakua, line 23: sorry, the tabsize went wrong :(

L3MON4D3:
Oh sorry, could you rebase (or maybe just cherry-pick your commits) onto `origin/aupatescape`? The initial branch included unrelated commits (from #440), but now it should be correct

konosubakonoakua:
> Oh sorry, could you rebase (or maybe just cherry-pick your commits) onto `origin/aupatescape`? The initial branch included unrelated commits (from #440), but now it should be correct

Sorry, im not very familar with GITHUB PullRquest and GIT, can u add the code?

```lua
-- str.lua
function M.aupatescape(s)
	local comma_escaped = s:gsub(",", "\\,")
	local backslash_replaced = comma_escaped:gsub("\\", "/")
	return vim.fn.fnameescape(backslash_replaced)
end

```

L3MON4D3:
Okay, I had another look, I think we cannot just replace `\`, it might occur naturally in an unix-path.
Could you make sure `aupatescape` still works for you with the newest commit?

konosubakonoakua:
> Okay, I had another look, I think we cannot just replace `\`, it might occur naturally in an unix-path. Could you make sure `aupatescape` still works for you with the newest commit?

Yes, the latest commit works well for me. :)

L3MON4D3:
Nice :)

PR #446, opened 2022-05-26
docs: Correct command name for selecting previous choice

kjerstadius:
The documentation currently refers to `luasnip-previous-choice`, which doesn't exist/work. This change updates the docs to use the proper command name, as recognized by the plugin.


L3MON4D3:
Right, thank you :+1:

PR #447, opened 2022-05-28
Fix name of `enable_autosnippets` parameter in documentation

maxmahlke:
The error in the docs tripped me up when trying to get autosnippets working. I grepped the repository and fixed all places where the `enable_autosnippets` was misspelled.

Cheers for the plugin!


L3MON4D3:
Ah, that's a nasty typo. Thanks for taking care of it :+1:

PR #462, opened 2022-06-13
otf: Prevent crash with $ before end of string.

L3MON4D3:
Reproducable with `asdf $`, but even `asdf $ asdf` leads to an unexpected `asdf ${1:asd}`.
@leiserfg I'm not completely sure where how the loop (I assume) is caused, do you have an idea (maybe the position in the string is getting messed up somewhere)?
I guess this should be `*` in any case, to allow for empty placeholders?


leiserfg:
Well never thought about having  `$` alone, if it  makes sense, the change does not breaks the old otf snippets so :+1:

Issue #457, opened 2022-06-11
Suggested way to create Rust Analyzer style postfix snippets


zdcthomas:
Hey! Love the plugin, it's so powerful, and feels awesome!

I'm wondering if there's a straightforward way to create Rust Analyzer style postfix snippets? I know I could get the cursor position and get the word before the `.` before the cursor, but before I dug too much into that, I thought I'd see if there was an easier way. 
If not, no problem! 

Thanks for the help!

L3MON4D3:
> Hey! Love the plugin, it's so powerful, and feels awesome!

Nice, that's great to hear :)

`regTrig` should work here:
```lua
ls.setup_snip_env()

-- match multiple %w before a `.` and store them in the first capture-group.
ls.add_snippets("all", {s({trig="(%w+)%.parens", regTrig=true, wordTrig=false}, {
	f(function(_, parent)
		return "("..parent.snippet.captures[1]..")"
	end),
	l("("..l.CAPTURE1..")", {})
})}, {key = "testsnippetshehe"})
```

In general the captures can be accessed as `snippet.captures[capture_group]`, for simple string-manipulations (like adding parenthesis above) you can just use `lambda` from `luasnip.extras`, makes it all a bit more readable.

Also, make sure to set `wordTrig`, otherwise luasnip will, when matching triggers, only consider text from the cursor back to the `.`, which is obviously problematic with postfix-snippets :D

zdcthomas:
Ooo that sounds perfect! I gave it a try and I'm running into a few issues.
I copied your snippet example, but when I expand on `foo.parens`, it expands to `foo.%w+%.parens` with `%w+` selected for replacement. 

my config setup looks like 
```lua
ls.config.set_config({
	updateevents = "TextChanged,TextChangedI",
	delete_check_events = "TextChanged,InsertEnter",
	region_check_events = "TextChanged",
	enable_autosnippets = false,
	store_selection_keys = "<Tab>",
	ext_opts = {
		[types.choiceNode] = {
			active = {
				virt_text = { { "●", "Boolean" } },
			},
		},
		[types.insertNode] = {
			active = {
				virt_text = { { "●", "Special" } },
			},
		},
	},
})
```

My guess is that I'm missing something in the config, but I can't quite find it. 
Sorry! really appreciate the help!

zdcthomas:
I'm using nvim cmp if that makes any difference

zdcthomas:
[...]

PR #463, opened 2022-06-17
Add postfix snippet extra

zdcthomas:
In a conversation over on issues #457, @L3MON4D3 and I discussed various ways
to best support postfix snippets. @L3MON4D3 came up with the initial callback
that supports this pr, thanks!!

This tries to handle all options intelligently, but if I missed something, please let me know!

questions:
- [x] right now, when a line doesn't match, I'm handing that by passing an empty string as `snippet.pre_dot` but I also toyed around with the idea of not letting the snippet expand. What do you think?

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -34,6 +34,7 @@ local ai = require("luasnip.nodes.absolute_indexer")
 local fmt = require("luasnip.extras.fmt").fmt
 local m = require("luasnip.extras").m
 local lambda = require("luasnip.extras").l
+local postfix = require("luasnip.extras.postfix").postfix
 ```
 
 # BASICS
@@ -70,7 +71,7 @@ It is possible to make snippets from one filetype available to another using
 # NODE
 
 Every node accepts, as its' last parameter, an optional table of arguments.
-There are some common ones (eg. [`node_ext_opts`](#ext_opts)), and some that
+There are some common ones (e.g. [`node_ext_opts`](#ext_opts)), and some that
 only apply to some nodes (`user_args` for both function and dynamicNode).
 These `opts` are only mentioned if they accept options that are not common to
 all nodes.
@@ -88,7 +89,7 @@ s({trig="trigger"}, {})
 entries:
 
 - `trig`: string, plain text by default. The only entry that must be given.
-- `name`: string, can be used by eg. `nvim-compe` to identify the snippet.
+- `name`: string, can be used by e.g. `nvim-compe` to identify the snippet.
 - `dscr`: string, description of the snippet, \n-separated or table
   for multiple lines.
 - `wordTrig`: boolean, if true, the snippet is only expanded if the word
@@ -126,12 +127,12 @@ The third argument (`opts`) is a table with the following valid keys:
   that just returns `true`).
   The function is called before the text is modified in any way.
   Some parameters are passed to the function: The line up to the cursor, the
-  matched trigger and the captures (table).
+  matched trigger, and the captures (table).
 - `show_condition`: Function with signature `f(line_to_cursor) -> bool`.
   It is a hint for completion-engines, indicating when the snippet should be
   included in current completion candidates.
   Defaults to a function returning `true`.
-  This is different than `condition` because `condition` is evaluated by
+  This is different from `condition` because `condition` is evaluated by
   LuaSnip on snippet expansion (and thus has access to the matched trigger and
   captures), while `show_condition` is evaluated by the completion-engine when
   scanning for available snippet candidates.
@@ -154,10 +155,10 @@ The third argument (`opts`) is a table with the following valid keys:
 - `child_ext_opts`, `merge_child_ext_opts`: `ext_opts` applied to the children
   of this snippet. More info [here](#ext_opts).
 
-This `opts`-table can also be passed to eg.	`snippetNode` or `indentSnippetNode`,
+This `opts`-table can also be passed to e.g.	`snippetNode` or `indentSnippetNode`,
 but only `callbacks` and the `ext_opts`-related options are used there.
 
-Snippets contain some interesting tables, eg. `snippet.env` contains variables
+Snippets contain some interesting tables, e.g. `snippet.env` contains variables
 used in the LSP-protocol like `TM_CURRENT_LINE` or `TM_FILENAME` or
 `snippet.captures`, where capture-groups of regex-triggers are stored.
 Additionally, the string that was used to trigger the snippet is stored in
@@ -198,7 +199,7 @@ s("trigger", {
 
 # INSERTNODE
 
-These Nodes contain editable text and can be jumped to- and from (eg.
+These Nodes contain editable text and can be jumped to- and from (e.g.
 traditional placeholders, like `$1` in textmate-snippets).
 
 The functionality is best demonstrated with an example:
@@ -250,7 +251,7 @@ s("trigger", {
 })
 ```
 
-as opposed to eg. the textmate-syntax, where tabstops are snippet-global:
+as opposed to e.g. the textmate-syntax, where tabstops are snippet-global:
 ```snippet
 ${1:First jump} :: ${2: ${3:Third jump} : ${4:Fourth jump}}
 ```
@@ -263,7 +264,7 @@ for potentially keeping some default-value:
 ```lua
 	s("trigger", i(1, "This text is SELECTed after expanding the snippet."))
 ```
-This initial text is defined the same way as textNodes, eg. can be multiline.
+This initial text is defined the same way as textNodes, e.g. can be multiline.
 
 `i(0)`s can have initial text, but do note that when the SELECTed text is
 replaced, its' replacement won't end up in the `i(0)`, but behind it (for
@@ -286,18 +287,18 @@ user-defined function:
 The first parameter of `f` is the function. Its parameters are:
 
 1. A table of the text of currently contained in the argnodes.
-      (eg. `{{line1}, {line1, line2}}`). The snippet-indent will be removed from
+      (e.g. `{{line1}, {line1, line2}}`). The snippet-indent will be removed from
       all lines following the first.
 
 2. The immediate parent of the `functionNode`. It is included here as it allows
-      easy access to anything that could be useful in functionNodes (ie.
+      easy access to anything that could be useful in functionNodes (i.e.
       `parent.snippet.env` or `parent.snippet.captures`, which contains capture
       groups of regex-triggered snippets). In most cases `parent.env` works,
       but if a `functionNode` is nested within a `snippetNode`, the immediate
       parent (a `snippetNode`) will contain neither `captures` nor `env`. Those
       are only stored in the `snippet`, which can be accessed as `parent.snippet`.
 
-3. The `user_args` passed in `opts`. Note that there may be multiple user_args (eg. `user_args1, ..., user_argsn`).
+3. The `user_args` passed in `opts`. Note that there may be multiple user_args (e.g. `user_args1, ..., user_argsn`).
 
 The function shall return a string, which will be inserted as-is, or a table
 of strings for multiline-string, here all lines following the first will be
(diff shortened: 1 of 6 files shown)


leiserfg:
Not sure if `pre_dot` is the best name for the function, sims like it's tied to the fact that the prefix is supposed to start with a `.` but that is not necessarily true.

L3MON4D3:
Nice, that looks good so far :+1:

I agree with @leiserfg, maybe `postfix_text` or `postfix_content`?

Another idea: it would be nice to put this value into `env`, that way we 
* don't "clutter" the snippet (and `snip.env.SOME_VAR` seems nicer than `snip.some_var`) and
* can actually use it in lambdas, which would be pretty cool (`l("("..l.POSTIFIX_TEXT..")", {})` is more readable than the function-node-version)

Actually implementing that would require some way to set/extend the `env` from the `pre_expand`-callback, I think it should be alright to just return the extensions.
I'll look into how to properly do that, sounds fun :D

L3MON4D3:
@zdcthomas you can play around with #464, I'll let it sit for a bit to see if there are better ways to modify the `env`, but this works

zdcthomas:
That looks great! I'll merge that in an try to incorporate it.

Also, it looked like the last workflow failed and it seemed like it was because I modified the luasnip.txt directly. I moved those changes to the DOC.md so hopefully that can run successfully now, but let me know if I need to change anything else.

I'm also going to take a crack at writing some tests!

L3MON4D3:
CI was acting up lately, stupid workflows, but I think they should be okay now.

Tests :+1::+1:

lua/luasnip/extras/postfix.lua
```diff
@@ -0,0 +1,75 @@
+local snip = require("luasnip").snippet
+local events = require("luasnip.util.events")
+local matches = {
+	default = [[[%w%.%_%-%"%']+$]],
+	line = "^.+$",
```
L3MON4D3, line 5: Does this have any use currently?

lua/luasnip/extras/postfix.lua
```diff
@@ -0,0 +1,75 @@
...
+	else
+		match_pattern = context.match_pattern or matches.default
+	end
+
+	context = vim.tbl_deep_extend("keep", context, { wordTrig = false })
+	opts = vim.tbl_deep_extend(
+		"force",
+		opts,
+		generate_opts(match_pattern, user_callback)
+	)
```
L3MON4D3, line 68: This could be replaced with a single set on `callbacks.-1.pre_expand`, right? Would be a tiny bit clearer IMO

lua/luasnip/extras/postfix.lua
zdcthomas: no, but I was thinking this section could be a set of commonly used matches, for people to plug in to the postfix snippets. We could totally get rid of it though

lua/luasnip/extras/postfix.lua
zdcthomas: Sounds good to me!

lua/luasnip/extras/postfix.lua
L3MON4D3: Ahh, I didn't see that it was exported.
Let's keep it, it should be useful, would you mention that in the doc as well?

lua/luasnip/extras/postfix.lua
zdcthomas: Is there a nicer way to do this than a couple of nested nil checks?

zdcthomas:
I'm having a small issue with the tests. My test looks like
```lua
local helpers = require("test.functional.helpers")(after_each)
local exec_lua, feed, exec = helpers.exec_lua, helpers.feed, helpers.exec
local ls_helpers = require("helpers")
local Screen = require("test.functional.ui.screen")

describe("postfix snippets", function()
	local screen

	before_each(function()
		helpers.clear()
		ls_helpers.session_setup_luasnip()

		screen = Screen.new(50, 3)
		screen:attach()
		screen:set_default_attr_ids({
			[0] = { bold = true, foreground = Screen.colors.Blue },
			[1] = { bold = true, foreground = Screen.colors.Brown },
			[2] = { bold = true },
			[3] = { background = Screen.colors.LightGray },
		})
	end)

	after_each(function()
		screen:detach()
	end)

	it(
		"creates a postfix snippet which changes the previous text once expanded",
		function()
			local postfix_snip = [[
	    pf(".parens", {
	      f(function(_, parent)
	        return "(" .. parent.postfix_match .. ")"
        end, {})
      })
	  ]]

			feed("bar.parens")
			exec_lua("ls.snip_expand(" .. postfix_snip .. ")")
			screen:expect({
				grid = [[
			(bar)^                                        |
			{0:~                                                 }|
			{2:-- INSERT --}                                      |]],
			})
		end
	)
end)
```
but the test is failing with this error message
```
FAILED   ../../tests/integration/postfix_snippet_spec.lua @ 27: postfix snippets creates a postfix snippet which changes the previous text once expanded
../../tests/integration/postfix_snippet_spec.lua:40: Row 1 did not match.
Expected:
  |*(bar)^                                        |
Actual:
  |*(r.parens)^                                        |
```

zdcthomas:
Looking more closely, it looks like none of the tests feed the trigger before expanding. If I match that behavior, I get the expected test result, but I'd like to understand why. I'll keep digging to understand more

zdcthomas:
Alright, I think that's pretty good for testing at the moment. Anything else y'all can see that needs changing? (Thanks for the patience with the first time contributor questions!)

lua/luasnip/extras/postfix.lua
L3MON4D3: Not really, but those could be moved into a separate function, `luasnip.util.util.tbl_set`,with an interface similar to `vim.tbl_get`, to keep the code readable

L3MON4D3:
> Looking more closely, it looks like none of the tests feed the trigger before expanding. If I match that behavior, I get the expected test result, but I'd like to understand why. I'll keep digging to understand more

Not sure if you've figured it out, in case not, the snippets don't need their trigger if they are `snip_expand`ed (of course there are exceptions, if they need `expand_params` for example)

L3MON4D3:
Ah, as for your question(which I've managed to completely miss so far, sorry😅): I think it's alright to pass an empty string, it shouldn't occur during regular usage, and if it happens through an error I'd rather that something unexpected happens, than nothing.

DOC.md
```diff
@@ -378,6 +379,81 @@ s("trig", {
...
+in most cases.
+
+The simplest example, which surrounds the text preceeding the `.br` with
+brackets `[]`, looks like:
+
+```lua
+
+      postfix(".br", {
+              f(function(_, parent)
+                    return "[" .. parent.postfix_match .. "]"
```
L3MON4D3, line 397: Would you update this to env (ofc) and use `parent.snippet` instead of `parent`? (in this case it's the same table, but there are snippets where they are not, and I'd rather the examples still work then).
Also, could you add one example with lambda? (is it obvious I like them yet? :P)

zdcthomas: Oh crap I missed one! 
And absolutely about the lambda lol :)

zdcthomas: Alright, Those examples are in!

L3MON4D3: Nice, that looks good :+1:

L3MON4D3:
Ah, one last thing, `postfix_match` should be `POSTFIX_MATCH` I think, just to be more in line with the other variables in `env` (`TM_SELECTED_TEXT`, `UUID` or `LINE_COMMENT`, for example)

zdcthomas:
Alright! s/postfix_match/POSTFIX_MATCH/g is done

L3MON4D3:
Perfect :ok_hand:
I'll take one last look at everything, merge #464 and then you can squash+rebase this into 1 or 2 commits on top of that. Sounds good?
BTW really nice job here, love to see PRs this clean :)

DOC.md
```diff
@@ -378,6 +379,96 @@ s("trig", {
...
+    against. The default match pattern is `"[%w%.%_%-]+$"`. Note the `$`. This
+    matches since only the line _up until_ the beginning of the trigger is
+    matched against the pattern, which makes the character immediately
+    preceeding the trigger match as the end of the string.
+
+Some other match strings, including the default, are available from the postfix
+module. `require("luasnip.extras.postfix).matches`:
+
+- `default`: [%w%.%_%-%"%']+$
+- 'line': `^.+$`
```
L3MON4D3, line 445: I think "line" and the default-pattern are also supposed to be inside backticks, right?

zdcthomas: yup!

zdcthomas: i was also thinking of taking out the `"` and `'` in the default pattern, since it wouldn't match on something like `"a string with spaces in it"`

L3MON4D3: `[%w%.%_%-%"%']` would match on `"somestring"`, wouldn't it? That's alright I think

Oh an interesting addition here would be treesitter-awareness, to make the postfix capture the node before the trigger. No pressure to implement this though, I just wanted to throw the thought out there

zdcthomas: Yeah I've been playing with that idea too. I think in another pr I might give it a shot, but I'd like to see how people use/ need from the normal postfix snips first I think

L3MON4D3:
Okay, apart from [this](https://github.com/L3MON4D3/LuaSnip/pull/463#discussion_r900707660), we're golden :+1:
(If you'd rather just leave it at `tbl_deep_extend`, that's alright too)

zdcthomas:
I was looking at it, and I think it's going to end up being cleaner with the `tbl_deep_extend` but it's 100% up to you, I can put a quick helper in utils.utils with the interface you were talking about.

L3MON4D3:
Naah, it's alright as-is, `tbl_set` vs `tbl_deep_extend` doesn't make a big difference

zdcthomas:
Sounds good to me 👍

zdcthomas:
Oh should I add a mention of postfix to the readme?

zdcthomas:
just in the features section? 
```
# Features
- Tabstops
- Text-Transformations using Lua functions
- Conditional Expansion
- Defining nested Snippets
- Filetype-specific Snippets
- Choices
- Dynamic Snippet creation
- Regex-Trigger
- Autotriggered Snippets
- Easy Postfix Snippets
- Fast
- Parse [LSP-Style](https://microsoft.github.io/language-server-protocol/specification#snippet_syntax) Snippets either directly in lua, as a vscode package or a snipmate snippet collection.
- Expand LSP-Snippets with [nvim-compe](https://github.com/hrsh7th/nvim-compe) (or its' successor, [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) (requires [cmp_luasnip](https://github.com/saadparwaiz1/cmp_luasnip)))
- Snippet history (jump back into older snippets)
- Resolve filetype at the cursor using Treesitter
```

L3MON4D3:
Oh yeah, please do

zdcthomas:
Alright, that's added to the readme. Hope that looks good!

L3MON4D3:
LGTM :ok_hand: 
squash it into just a few commits, then it's g2g

zdcthomas:
Alright, squashed down and gtg hopefully!

L3MON4D3:
Aaalmost, 1390052 contains some changes from other commits, eg. the `pre_expand_env` and a fix for the `otf`-snippets.
It would also be pretty nice if the changes to `luasnip.txt` weren't included (but this isn't necessary). (maybe do a soft reset and filter out all changes other than `extras/postfix.lua`, `DOC.md` and that test-file)

zdcthomas:
Ah yeah, sorry I totally missed that, I went one commit too recent with the reset. Also, should luasnip.txt be added to the gitignore? r is it sometimes updated manually as well?

L3MON4D3:
Nice, that's perfect :+1:
> Also, should luasnip.txt be added to the gitignore? or is it sometimes updated manually as well?

`.gitignore` would be the best solution for it, but since this should be usable after a single clone it's simpler to just check in git. (also no, it isn't updated manually, just by that workflow)

Okay I'm merging, thank you very much for your patience+work here :D 
This is a great feature :+1:

zdcthomas:
Woo! thank you for your patience and help with my first contribution! It definitely won't be my last!

Issue #454, opened 2022-06-08
Dynamic lambda caching?


ranebrown:
Does the dynamic lambda do some sort of caching? This results in the same "random" string every time. Works as a function node but then isn't selected or over writable.

```lua
local charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
math.randomseed(os.clock())
local function randomString(length)
    local ret = {}
    local r
    for _ = 1, length do
        r = math.random(1, #charset)
        table.insert(ret, charset:sub(r, r))
    end
    return table.concat(ret)
end

dl(1, l.TM_FILENAME:upper():gsub("%.", "_") .. "_" .. randomString(6), {}),
```

leiserfg:
The variables in the snippet environment (like `TM_IFILENAME`) keep the value for the snippet expansion, so maybe you could use `l.RANDOM_HEX`

ranebrown:
`l.RANDOM_HEX` does work. Is there a way to reset the variables and allow a function call to work each time?

ranebrown:
This also works but isn't as clean.
```lua
d(1, function(args, snip)
    return sn(
        nil,
        {
            i(
                1,
                snip.env.TM_FILENAME:upper():gsub("%.", "_")
                    .. "_"
                    .. randomString(6)
            ),
        }
    )
end, {}),

```

L3MON4D3:
> `l.RANDOM_HEX` does work. Is there a way to reset the variables and allow a function call to work each time?

Although this is a bit convoluted, you could override the `__concat`-metamethod on some table and pass that:
```lua
local charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
math.randomseed(os.clock())
local function randomString(length)
    local ret = {}
    local r
    for _ = 1, length do
        r = math.random(1, #charset)
        table.insert(ret, charset:sub(r, r))
    end
    return table.concat(ret)
end

local o = setmetatable({}, {
	__concat = function(x)
		return x .. randomString(6)
	end
})

ls.setup_snip_env()

ls.snip_expand(s("", {
	dl(2, l._1 .. o, 1), i(1), t"texttext"
}))
```

I think adding functions to (dynamic)lambda is a bit unnecessary, at that point one might just as well use regular dynamic/functionNode.

PR #464, opened 2022-06-17
allow modification of env through pre_expand-callback.

L3MON4D3:
The `pre_expand`-callback may return a table used to extend the snippet's env.

I'm a tiny bit unsure about repurposing the `pre_expand`-callback for extending/modifying the snippet's state (no other event-callback does this), but introducing a different mechanism for this feels more wrong, since the pre_expand-callback is already functional and used for modifying eg. the text around the snippet.

example:
```lua
ls.setup_snip_env()

ls.add_snippets("all", {
	s("trig6", {
		l(l.P, {})
	}, {
		callbacks = {
			[-1] = {
				[events.pre_expand] = function()
					return {env_override = {
						P = "asdf"
					}}
				end
			}
		}
	})
}, {
	key ="asdff"
})
```

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1561,9 +1561,11 @@ Events can be used to react to some action inside snippets. These callbacks can
 be defined per-snippet (`callbacks`-key in snippet constructor) or globally
 (autocommand).
 
-`callbacks`: `fn(node[, event_args])`  
+`callbacks`: `fn(node[, event_args]) -> event_res`  
 All callbacks get the `node` associated with the event and event-specific
 optional arguments, `event_args`.
+`event_res` is only used in one event, `pre_expand`, where some properties of
+the snippet can be changed.
 
 `autocommand`:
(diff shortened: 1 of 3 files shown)


L3MON4D3:
Also, the new variables won't work in lsp-snippets because we do a check for var-existence at [parse-time](https://github.com/L3MON4D3/LuaSnip/blob/f08226a2e8b77a718b3c36bc19b4b7b53ff98541/lua/luasnip/util/parser.lua#L84), when we have no idea which vars actually exist in the snippet. Simply moving that check to runtime should be enough.

Should be ok to tackle that in #445, I doubt it'll be very relevant in practice, but it's a small deficit.

bew:
This could be used to give some initial state to a snippet? (to properly impl #454)

L3MON4D3:
Good thought, yes that should work!

PR #479, opened 2022-07-08
Add env_namespace

### LINKED ISSUE #476: How to include environment variable in a snippet definition
### Discussed in https://github.com/L3MON4D3/LuaSnip/discussions/475

<div type='discussions-op-text'>

<sup>Originally posted by **jimcornmell** July  5, 2022</sup>
Allow the ability to include any environment variable in a snippet definition json file eg `$HOME`:

For example:
```json
{
  "A Description": {
    "prefix": "home",
    "body": [
         "$HOME",
    ],
....
}
```
</div>

leiserfg:
Fixes #476
The function to check if a variable is valid was removed because now there is no how to know 
which variables will be available until the snippet gets expanded.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1094,6 +1094,50 @@ To use any `*SELECT*` variable, the `store_selection_keys` must be set via
 hitting `<Tab>` while in Visualmode will populate the `*SELECT*`-vars for the next
 snippet and then clear them.
 
+## Environemt Namespaces
+
+You can also add your own variables by using the `ls.env_namespace(name, opts)` where:
+
+* `name`  is string that does not contains the character "_" and names the namespace.
+* `opts` is a table containing:
+    * `vars` 
+    Is a function that receives a string and returns a value for the var with that name
+    or a table from var name to a value
+    (in this case, if the value is a function it will be executed  lazily once per snippet expansion).
+    * `init` A function that gets a parameter pos with the 0-based position of the cursor and returns
+        a table of variables that will set to the environment of the snippet on expansion,
+        use this for vars that have to be calculated in that moment or that depend on each other.
+    * `eager` A list of names of vars that will be taken from `vars` and appended  eagerly (like those in init)
+    * `multiline_vars` Says if certain vars are a table or just a string, can be a function that get's the name of the var and returns
+        true if the var is a key, a list of vars that are tables or a boolean for the full namespace, it's false by default.
+
+The four fields of `opts` are optional but you need to provide either `init` or  `vars`, and `eager` can't be without `vars`.
+Also you can't use namespaces that override default vars.
+
+You can also add your own variables by using the `ls.extend_env(namespace, vars)` where:
+
+* `namespace` is string that does not contains the character "_".
+* `vars` 
+  * is a function that receives a string and returns a value for the var with that name
+  * or a table from var name to a value 
+    (in this case, if the value is a function it will be executed  lazily once per snippet expansion)
+
+A simple example to make it more clear:
+
+```lua
+local random_lang()
+    return return ({"LUA", "VIML"})[math.floor(math.random()+0.5)]
+end
+
+ls.extend_env(MY, { NAME="LuaSnip",  LANG=random_lang })`
+
+-- then you can use  $MY_NAME and $MY_LANG in your snippets
+
+ls.env_namespace("SYS", {vars=os.getenv, eager={"HOME"}})
+
+-- then you can use  $SYS_HOME and $SYS_USER (or any other system environment var) in your snippets
+```
+
 
 # LOADERS
 
--- a/DOC.md
+++ b/DOC.md
@@ -1098,17 +1098,17 @@ snippet and then clear them.
 
 You can also add your own variables by using the `ls.env_namespace(name, opts)` where:
 
-* `name`  is string that does not contains the character "_" and names the namespace.
-* `opts` is a table containing:
-    * `vars` 
+* `name`: `string` the names the namespace, can't contain the character "_"
+* `opts` is a table containing (in every case `EnvVal` is the same as `string|list[string]`:
+    * `vars`: `(fn(name:string)->EnvVal) | map[string, EnvVal]`
     Is a function that receives a string and returns a value for the var with that name
     or a table from var name to a value
     (in this case, if the value is a function it will be executed  lazily once per snippet expansion).
-    * `init` A function that gets a parameter pos with the 0-based position of the cursor and returns
+    * `init`: `fn(pos: pair[int])->map[string, EnvVal]` Takes the 0-based position of the cursor and returns
         a table of variables that will set to the environment of the snippet on expansion,
         use this for vars that have to be calculated in that moment or that depend on each other.
-    * `eager` A list of names of vars that will be taken from `vars` and appended  eagerly (like those in init)
-    * `multiline_vars` Says if certain vars are a table or just a string, can be a function that get's the name of the var and returns
+    * `eager`: `list[string]` names of variables that will be taken from `vars` and appended eagerly (like those in init)
+    * `multiline_vars`: `(fn(name:string)->bool)|map[sting, bool]|bool` Says if certain vars are a table or just a string, can be a function that get's the name of the var and returns
         true if the var is a key, a list of vars that are tables or a boolean for the full namespace, it's false by default.
 
 The four fields of `opts` are optional but you need to provide either `init` or  `vars`, and `eager` can't be without `vars`.
--- a/DOC.md
+++ b/DOC.md
@@ -1114,28 +1114,25 @@ You can also add your own variables by using the `ls.env_namespace(name, opts)` 
 The four fields of `opts` are optional but you need to provide either `init` or  `vars`, and `eager` can't be without `vars`.
 Also you can't use namespaces that override default vars.
 
-You can also add your own variables by using the `ls.extend_env(namespace, vars)` where:
-
-* `namespace` is string that does not contains the character "_".
-* `vars` 
-  * is a function that receives a string and returns a value for the var with that name
-  * or a table from var name to a value 
-    (in this case, if the value is a function it will be executed  lazily once per snippet expansion)
 
 A simple example to make it more clear:
 
 ```lua
 local random_lang()
-    return ({"LUA", "VIML"})[math.floor(math.random()+1.5)]
+    return ({"LUA", "VIML", "VIML9"})[math.floor(math.random()/2 + 1.5)]
 end
 
-ls.extend_env(MY, { NAME="LuaSnip",  LANG=random_lang })`
+ls.env_namespace("MY", {vars={ NAME="LuaSnip",  LANG=random_lang }})`
 
 -- then you can use  $MY_NAME and $MY_LANG in your snippets
 
 ls.env_namespace("SYS", {vars=os.getenv, eager={"HOME"}})
 
--- then you can use  $SYS_HOME and $SYS_USER (or any other system environment var) in your snippets
+-- then you can use  $SYS_HOME which was eagerly initialized but also $SYS_USER (or any other system environment var) in your snippets
+
+lsp.env_namespace("POS", {init=function(pos) return {"VAL": vim.inspect(pos)}} end)
+
+-- then you can use  $POS_VAL in your snippets
 ```
 
 
--- a/DOC.md
+++ b/DOC.md
@@ -1094,7 +1094,7 @@ To use any `*SELECT*` variable, the `store_selection_keys` must be set via
 hitting `<Tab>` while in Visualmode will populate the `*SELECT*`-vars for the next
 snippet and then clear them.
 
-## Environemt Namespaces
+## Environment Namespaces
 
 You can also add your own variables by using the `ls.env_namespace(name, opts)` where:
 
--- a/DOC.md
+++ b/DOC.md
@@ -1126,7 +1126,7 @@ A simple example to make it more clear:
 
 ```lua
 local random_lang()
-    return return ({"LUA", "VIML"})[math.floor(math.random()+0.5)]
+    return ({"LUA", "VIML"})[math.floor(math.random()+1.5)]
 end
 
 ls.extend_env(MY, { NAME="LuaSnip",  LANG=random_lang })`
--- a/lua/luasnip/nodes/snippet.lua
+++ b/lua/luasnip/nodes/snippet.lua
@@ -689,12 +689,7 @@ function Snippet:fake_expand(opts)
 	if opts.env then
 		self.env = opts.env
 	else
-		self.env = {}
-		setmetatable(self.env, {
-			__index = function(_, key)
-				return Environ.is_table(key) and { "$" .. key } or "$" .. key
-			end,
-		})
+		self.env = Environ.fake()
 	end
 
 	self.captures = {}
--- a/lua/luasnip/util/_builtin_vars.lua
+++ b/lua/luasnip/util/_builtin_vars.lua
@@ -175,7 +175,7 @@ local _is_table = {
 
 return {
     is_table = function(key) return _is_table[key] or false end,
-    lazy = lazy_vars,
-    eager = eager_vars,
+    vars = lazy_vars,
+    init = eager_vars,
     builtin_ns = builtin_ns,
 }
--- a/lua/luasnip/util/_builtin_vars.lua
+++ b/lua/luasnip/util/_builtin_vars.lua
@@ -1,113 +1,112 @@
 local util = require("luasnip.util.util")
 local lazy_vars = {}
 
-
 -- Variables defined in https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables
 
 -- Inherited from TextMate
 function lazy_vars.TM_FILENAME()
-    return vim.fn.expand("%:t")
+	return vim.fn.expand("%:t")
 end
 
 function lazy_vars.TM_FILENAME_BASE()
-    return vim.fn.expand("%:t:s?\\.[^\\.]\\+$??")
+	return vim.fn.expand("%:t:s?\\.[^\\.]\\+$??")
 end
 
 function lazy_vars.TM_DIRECTORY()
-    return vim.fn.expand("%:p:h")
+	return vim.fn.expand("%:p:h")
 end
 
 function lazy_vars.TM_FILEPATH()
-    return vim.fn.expand("%:p")
+	return vim.fn.expand("%:p")
 end
 
 -- Vscode only
 
 function lazy_vars.CLIPBOARD() -- The contents of your clipboard
-    return vim.fn.getreg('"', 1, true)
+	return vim.fn.getreg('"', 1, true)
 end
 
 local function buf_to_ws_part()
-    local LSP_WORSKPACE_PARTS = "LSP_WORSKPACE_PARTS"
-    local ok, ws_parts = pcall(vim.api.nvim_buf_get_var, 0, LSP_WORSKPACE_PARTS)
-    if not ok then
-        local file_path = vim.fn.expand("%:p")
+	local LSP_WORSKPACE_PARTS = "LSP_WORSKPACE_PARTS"
+	local ok, ws_parts = pcall(vim.api.nvim_buf_get_var, 0, LSP_WORSKPACE_PARTS)
+	if not ok then
+		local file_path = vim.fn.expand("%:p")
 
-        for _, ws in pairs(vim.lsp.buf.list_workspace_folders()) do
-            if file_path:find(ws, 1, true) == 1 then
-                ws_parts = { ws, file_path:sub(#ws + 2, -1) }
-                break
-            end
-        end
-        -- If it can't be extracted from lsp, then we use the file path
-        if not ok and not ws_parts then
-            ws_parts = { vim.fn.expand("%:p:h"), vim.fn.expand("%:p:t") }
-        end
-        vim.api.nvim_buf_set_var(0, LSP_WORSKPACE_PARTS, ws_parts)
-    end
-    return ws_parts
+		for _, ws in pairs(vim.lsp.buf.list_workspace_folders()) do
+			if file_path:find(ws, 1, true) == 1 then
+				ws_parts = { ws, file_path:sub(#ws + 2, -1) }
+				break
+			end
+		end
+		-- If it can't be extracted from lsp, then we use the file path
+		if not ok and not ws_parts then
+			ws_parts = { vim.fn.expand("%:p:h"), vim.fn.expand("%:p:t") }
+		end
+		vim.api.nvim_buf_set_var(0, LSP_WORSKPACE_PARTS, ws_parts)
+	end
+	return ws_parts
 end
 
 function lazy_vars.RELATIVE_FILEPATH() -- The relative (to the opened workspace or folder) file path of the current document
-    return buf_to_ws_part()[2]
+	return buf_to_ws_part()[2]
 end
 
 function lazy_vars.WORKSPACE_FOLDER() -- The path of the opened workspace or folder
-    return buf_to_ws_part()[1]
+	return buf_to_ws_part()[1]
 end
 
 function lazy_vars.WORKSPACE_NAME() -- The name of the opened workspace or folder
-    local parts = vim.split(buf_to_ws_part()[1] or "", "[\\/]")
-    return parts[#parts]
+	local parts = vim.split(buf_to_ws_part()[1] or "", "[\\/]")
+	return parts[#parts]
 end
 
 -- DateTime Related
 function lazy_vars.CURRENT_YEAR()
-    return os.date("%Y")
+	return os.date("%Y")
 end
 
 function lazy_vars.CURRENT_YEAR_SHORT()
-    return os.date("%y")
+	return os.date("%y")
 end
 
 function lazy_vars.CURRENT_MONTH()
-    return os.date("%m")
+	return os.date("%m")
 end
 
 function lazy_vars.CURRENT_MONTH_NAME()
-    return os.date("%B")
+	return os.date("%B")
 end
 
 function lazy_vars.CURRENT_MONTH_NAME_SHORT()
-    return os.date("%b")
+	return os.date("%b")
 end
 
 function lazy_vars.CURRENT_DATE()
-    return os.date("%d")
+	return os.date("%d")
 end
 
 function lazy_vars.CURRENT_DAY_NAME()
-    return os.date("%A")
+	return os.date("%A")
 end
 
 function lazy_vars.CURRENT_DAY_NAME_SHORT()
-    return os.date("%a")
+	return os.date("%a")
 end
 
 function lazy_vars.CURRENT_HOUR()
-    return os.date("%H")
+	return os.date("%H")
 end
 
 function lazy_vars.CURRENT_MINUTE()
-    return os.date("%M")
+	return os.date("%M")
 end
 
 function lazy_vars.CURRENT_SECOND()
-    return os.date("%S")
+	return os.date("%S")
 end
 
 function lazy_vars.CURRENT_SECONDS_UNIX()
-    return tostring(os.time())
+	return tostring(os.time())
 end
 
 -- For inserting random values
@@ -115,67 +114,71 @@ end
 math.randomseed(os.time())
 
 function lazy_vars.RANDOM()
-    return string.format("%06d", math.random(999999))
+	return string.format("%06d", math.random(999999))
 end
 
 function lazy_vars.RANDOM_HEX()
-    return string.format("%06x", math.random(16777216)) --16^6
+	return string.format("%06x", math.random(16777216)) --16^6
 end
 
 function lazy_vars.UUID()
-    local random = math.random
-    local template = "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx"
-    local out
-    local function subs(c)
-        local v = (((c == "x") and random(0, 15)) or random(8, 11))
-        return string.format("%x", v)
-    end
+	local random = math.random
+	local template = "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx"
+	local out
+	local function subs(c)
+		local v = (((c == "x") and random(0, 15)) or random(8, 11))
+		return string.format("%x", v)
+	end
 
-    out = template:gsub("[xy]", subs)
-    return out
+	out = template:gsub("[xy]", subs)
+	return out
 end
 
 function lazy_vars.LINE_COMMENT()
-    return util.buffer_comment_chars()[1]
+	return util.buffer_comment_chars()[1]
 end
 
 function lazy_vars.BLOCK_COMMENT_START()
-    return util.buffer_comment_chars()[2]
+	return util.buffer_comment_chars()[2]
 end
 
 function lazy_vars.BLOCK_COMMENT_END()
-    return util.buffer_comment_chars()[3]
+	return util.buffer_comment_chars()[3]
 end
 
 -- These are the vars that have to be populated once the snippet starts to avoid any issue
 local function eager_vars(pos)
-    local vars = {}
-    vars.TM_CURRENT_LINE = vim.api.nvim_buf_get_lines(0, pos[1], pos[1] + 1, false)[1]
-    vars.TM_CURRENT_WORD = util.word_under_cursor(pos, vars.TM_CURRENT_LINE)
-    vars.TM_LINE_INDEX = tostring(pos[1])
-    vars.TM_LINE_NUMBER = tostring(pos[1] + 1)
-    vars.SELECT_RAW, vars.SELECT_DEDENT, vars.TM_SELECTED_TEXT = util.get_selection()
-    return vars
+	local vars = {}
+	vars.TM_CURRENT_LINE =
+		vim.api.nvim_buf_get_lines(0, pos[1], pos[1] + 1, false)[1]
+	vars.TM_CURRENT_WORD = util.word_under_cursor(pos, vars.TM_CURRENT_LINE)
+	vars.TM_LINE_INDEX = tostring(pos[1])
+	vars.TM_LINE_NUMBER = tostring(pos[1] + 1)
+	vars.SELECT_RAW, vars.SELECT_DEDENT, vars.TM_SELECTED_TEXT =
+		util.get_selection()
+	return vars
 end
 
 local builtin_ns = { SELECT = true }
 
 for name, _ in pairs(lazy_vars) do
-    local parts = vim.split(name, "_")
-    if #parts > 1 then
-        builtin_ns[ parts[1] ] = true
-    end
+	local parts = vim.split(name, "_")
+	if #parts > 1 then
+		builtin_ns[parts[1]] = true
+	end
 end
 
 local _is_table = {
-        TM_SELECTED_TEXT = true,
-        SELECT_RAW = true,
-        SELECT_DEDENT = true,
-    }
+	TM_SELECTED_TEXT = true,
+	SELECT_RAW = true,
+	SELECT_DEDENT = true,
+}
 
 return {
-    is_table = function(key) return _is_table[key] or false end,
-    vars = lazy_vars,
-    init = eager_vars,
-    builtin_ns = builtin_ns,
+	is_table = function(key)
+		return _is_table[key] or false
+	end,
+	vars = lazy_vars,
+	init = eager_vars,
+	builtin_ns = builtin_ns,
 }
--- a/lua/luasnip/util/_builtin_vars.lua
+++ b/lua/luasnip/util/_builtin_vars.lua
@@ -1,0 +1,180 @@
+local util = require("luasnip.util.util")
+local lazy_vars = {}
+
+
+-- Variables defined in https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables
+
+-- Inherited from TextMate
+function lazy_vars.TM_FILENAME()
+    return vim.fn.expand("%:t")
+end
+
+function lazy_vars.TM_FILENAME_BASE()
+    return vim.fn.expand("%:t:s?\\.[^\\.]\\+$??")
+end
+
+function lazy_vars.TM_DIRECTORY()
+    return vim.fn.expand("%:p:h")
+end
+
+function lazy_vars.TM_FILEPATH()
+    return vim.fn.expand("%:p")
+end
+
+-- Vscode only
+
+function lazy_vars.CLIPBOARD() -- The contents of your clipboard
+    return vim.fn.getreg('"', 1, true)
+end
+
+local function buf_to_ws_part()
+    local LSP_WORSKPACE_PARTS = "LSP_WORSKPACE_PARTS"
+    local ok, ws_parts = pcall(vim.api.nvim_buf_get_var, 0, LSP_WORSKPACE_PARTS)
+    if not ok then
+        local file_path = vim.fn.expand("%:p")
+
+        for _, ws in pairs(vim.lsp.buf.list_workspace_folders()) do
+            if file_path:find(ws, 1, true) == 1 then
+                ws_parts = { ws, file_path:sub(#ws + 2, -1) }
+                break
+            end
+        end
+        -- If it can't be extracted from lsp, then we use the file path
+        if not ok and not ws_parts then
+            ws_parts = { vim.fn.expand("%:p:h"), vim.fn.expand("%:p:t") }
+        end
+        vim.api.nvim_buf_set_var(0, LSP_WORSKPACE_PARTS, ws_parts)
+    end
+    return ws_parts
+end
+
+function lazy_vars.RELATIVE_FILEPATH() -- The relative (to the opened workspace or folder) file path of the current document
+    return buf_to_ws_part()[2]
+end
+
+function lazy_vars.WORKSPACE_FOLDER() -- The path of the opened workspace or folder
+    return buf_to_ws_part()[1]
+end
+
+function lazy_vars.WORKSPACE_NAME() -- The name of the opened workspace or folder
+    local parts = vim.split(buf_to_ws_part()[1] or "", "[\\/]")
+    return parts[#parts]
+end
+
+-- DateTime Related
+function lazy_vars.CURRENT_YEAR()
+    return os.date("%Y")
+end
+
+function lazy_vars.CURRENT_YEAR_SHORT()
+    return os.date("%y")
+end
+
+function lazy_vars.CURRENT_MONTH()
+    return os.date("%m")
+end
+
+function lazy_vars.CURRENT_MONTH_NAME()
+    return os.date("%B")
+end
+
+function lazy_vars.CURRENT_MONTH_NAME_SHORT()
+    return os.date("%b")
+end
+
+function lazy_vars.CURRENT_DATE()
+    return os.date("%d")
+end
+
+function lazy_vars.CURRENT_DAY_NAME()
+    return os.date("%A")
+end
+
+function lazy_vars.CURRENT_DAY_NAME_SHORT()
+    return os.date("%a")
+end
+
+function lazy_vars.CURRENT_HOUR()
+    return os.date("%H")
+end
+
+function lazy_vars.CURRENT_MINUTE()
+    return os.date("%M")
+end
+
+function lazy_vars.CURRENT_SECOND()
+    return os.date("%S")
+end
+
+function lazy_vars.CURRENT_SECONDS_UNIX()
+    return tostring(os.time())
+end
+
+-- For inserting random values
+
+math.randomseed(os.time())
+
+function lazy_vars.RANDOM()
+    return string.format("%06d", math.random(999999))
+end
+
+function lazy_vars.RANDOM_HEX()
+    return string.format("%06x", math.random(16777216)) --16^6
+end
+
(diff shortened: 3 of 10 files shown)


L3MON4D3:
Ohhh nice, I'll have a look tomorrow :)

Removing that function makes sense :+1:

L3MON4D3:
Could you rebase? Looks like stylua formats lots of stuff differently (better, I think), but it adds a some noise to the PR (and per-commit-view also has some style-differences :/)

leiserfg:
I rebased indeed, no idea why so many files are getting formated.

leiserfg:
You could review just the first commit, as it has less noise.

L3MON4D3:
Whoops, my bad, I didn't push the format :facepalm:
Rebase again? :sweat_smile:

L3MON4D3:
Nice, ty, much better :+1:

leiserfg:
Replaced the concatenation with a string splice to avoid edge cases.

lua/luasnip/util/environ.lua
```diff
@@ -30,10 +57,28 @@ function Environ:new(pos, o)
 	return o
 end
 
+function Environ.extend_env(name, val)
+	env_namespaces[name] = val
```
L3MON4D3, line 61: What do you think about making the `name may not contain "_"`-condition explicit here? (also, empty name should not cause any issues, right? otherwise that could also be checked)

lua/luasnip/util/environ.lua
```diff
@@ -21,6 +21,33 @@ local function _fill_eager_vars(env, pos)
...
+	end
+	local nmsp = env_namespaces[parts[1]]
+
+	if not nmsp then
+		return nil
+	end
+
+	local varname = full_varname:sub(#parts[1] + 2)
+
+	if type(nmsp) == "function" then
```
L3MON4D3, line 40: Nice, this is better than requiring the user to create a table even if they only need (basically) `__index` :+1:
I think this could be even nicer by making the namespaces exclusively functions (if a table is passed it is wrapped inside a function).

lua/luasnip/util/environ.lua
leiserfg: For  `Environ` it could be empty, but seems like the parser does not like variables starting with `_` so I will add a check for both.

L3MON4D3:
You can also push a commit with that as the message, and squash once at the end :wink: (I'd slightly prefer that, actually, but I won't mind force pushes as well)

L3MON4D3:
Ah, one more thing: What do you think about providing the equivalent of eager variables to the namespaces?
Might be useful to initialize some vars once when the environment is first built, but I can't think of a good example that isn't covered by the existing variables.

lua/luasnip/util/environ.lua
L3MON4D3: Ah, right, the parser :grimacing:
(You're talking about the current one, right? neovims might actually not have a problem with that (except completely emtpy variable names, probably))

leiserfg:
> 

That is messy to implement with an environment "resolver" function, as we don't know all the possible variables beforehand, so we can't evaluate them eagerly. 
A way to implement it is to change how the environment mechanism works, instead of having a dynamic object that resolves the variables once required, we could make some thing like `Environ:instance(list_of_required_vars)` on the snippet creation, that way all the variables will be eagerly evaluated but only those that are required in the snippet, but that requires a change on the parser.

lua/luasnip/util/environ.lua
leiserfg: We can relax it after moving to the new parser

leiserfg:
A similar option could be to expose the list of known vars of a snippet, so they can be passed to `Environ:new` and pre-populated there.

L3MON4D3:
> That is messy to implement with an environment "resolver" function, as we don't know all the possible variables beforehand, so we can't evaluate them eagerly.

True :/

> A way to implement it is to change how the environment mechanism works, instead of having a dynamic object that resolves the variables once required, we could make something like `Environ:instance(list_of_required_vars)` on the snippet expansion, that way all the variables will be eagerly evaluated but only those that are required in the snippet, but that requires a change on the parser.

Interesting, it's a perfect solution for parsed snippets, but I don't think it's possible to do for lua-snippets, since `env` is accessed in functions, where we have no idea what's happening.

How about requiring the names of eager variables in, say, a table passed to `extend_env`?

L3MON4D3:
> A similar option could be to expose the list of known vars of a snippet, so they can be passed to `Environ:new` and pre-populated there.

You mean as a parameter to the snippet's constructor?
Not sure if that would be the way to go about this, I'd categorize eagerness as a property of the variable (`TM_CURRENT_LINE` would always be eager, independently of the snippet)

leiserfg:
> > A similar option could be to expose the list of known vars of a snippet, so they can be passed to `Environ:new` and pre-populated there.
> 
> You mean as a parameter to the snippet's constructor? Not sure if that would be the way to go about this, I'd categorize eagerness as a property of the variable (`TM_CURRENT_LINE` would always be eager, independently of the snippet)



> > A similar option could be to expose the list of known vars of a snippet, so they can be passed to `Environ:new` and pre-populated there.
> 
> You mean as a parameter to the snippet's constructor? Not sure if that would be the way to go about this, I'd categorize eagerness as a property of the variable (`TM_CURRENT_LINE` would always be eager, independently of the snippet)



> > A similar option could be to expose the list of known vars of a snippet, so they can be passed to `Environ:new` and pre-populated there.
> 
> You mean as a parameter to the snippet's constructor? Not sure if that would be the way to go about this, I'd categorize eagerness as a property of the variable (`TM_CURRENT_LINE` would always be eager, independently of the snippet)

No, I mean when you parse the snippet you can know the variables it requires, so during the snippet expansion, they could be all pre-calculated. That way all the required vars are eager for the snippet but lazy in general.

DOC.md
```diff
@@ -1095,6 +1095,31 @@ hitting `<Tab>` while in Visualmode will populate the `*SELECT*`-vars for the ne
...
+* `vars` 
+  * is a function that receives a string and returns a value for the var with that name
+  * or a table from var name to a value 
+    (in this case, if the value is a function it will be executed  lazily once per snippet expansion)
+
+A simple example to make it more clear:
+
+```lua
+local random_lang()
+    return return ({"LUA", "VIML"})[math.floor(math.random()+0.5)]
```
L3MON4D3, line 1110: btw, is it intentional that "VIML" will never be returned here? xD
Honestly, leave it like this, that's amazing :smile:

leiserfg: crap, just noticed the double return :smile:

L3MON4D3:
> No, I mean when you parse the snippet you can know the variables it requires, so during the snippet expansion, they could be all pre-calculated. That way all the required vars are eager for the snippet but lazy in general.

But how'd we do that for lua-snippets, which also access `parent.snippet.env.NS_SOMEVAR` in dynamic/functionNode-functions? Is it possible to introspect functions in this way?

leiserfg:
> > No, I mean when you parse the snippet you can know the variables it requires, so during the snippet expansion, they could be all pre-calculated. That way all the required vars are eager for the snippet but lazy in general.
> 
> But how'd we do that for lua-snippets, which also access `parent.snippet.env.NS_SOMEVAR` in dynamic/functionNode-functions? Is it possible to introspect functions in this way?



> > No, I mean when you parse the snippet you can know the variables it requires, so during the snippet expansion, they could be all pre-calculated. That way all the required vars are eager for the snippet but lazy in general.
> 
> But how'd we do that for lua-snippets, which also access `parent.snippet.env.NS_SOMEVAR` in dynamic/functionNode-functions? Is it possible to introspect functions in this way?



> > No, I mean when you parse the snippet you can know the variables it requires, so during the snippet expansion, they could be all pre-calculated. That way all the required vars are eager for the snippet but lazy in general.
> 
> But how'd we do that for lua-snippets, which also access `parent.snippet.env.NS_SOMEVAR` in dynamic/functionNode-functions? Is it possible to introspect functions in this way?
No, for those there is not much we can do,  given that we can have potentially infinite vars, we can't make them eager.

DOC.md
L3MON4D3: I'm 100% for {"C", "LUA", "VIML9"} with the same `+1.5` :P
Putting some oil on the fire xD

L3MON4D3:
> No, for those there is not much we can do, given that we can have potentially infinite vars, we can't make them eager.

Yeah, that's the problem.
I think the best way out is that every namespace just advertises which variables need to be eagerly evaluated.

leiserfg:
> > No, for those there is not much we can do, given that we can have potentially infinite vars, we can't make them eager.
> 
> Yeah, that's the problem. I think the best way out is that every namespace just advertises which variables need to be eagerly evaluated.



> > No, for those there is not much we can do, given that we can have potentially infinite vars, we can't make them eager.
> 
> Yeah, that's the problem. I think the best way out is that every namespace just advertises which variables need to be eagerly evaluated.



> > No, for those there is not much we can do, given that we can have potentially infinite vars, we can't make them eager.
> 
> Yeah, that's the problem. I think the best way out is that every namespace just advertises which variables need to be eagerly evaluated.

What do you think of passing a 3rd (optional) arg to `extend_env` with a list of names of vars to make them eager?

L3MON4D3:
> What do you think of passing a 3rd (optional) arg to extend_env with a list of names of vars to make them eager?

Sounds good :+1:
Probably best to do the `opts`-"pattern" so there is backwards-compatibility when adding more options.

DOC.md
leiserfg: Rust is missing as well

lua/luasnip/util/parser.lua
```diff
@@ -81,32 +81,24 @@ end
...
+	-- all lines of the variable (important for eg. TM_SELECTED_TEXT).
+	if last_text ~= nil and #last_text.static_text > 1 then
+		local last_line_indent =
+			last_text.static_text[#last_text.static_text]:match("^%s+$")
+		if last_line_indent then
+			f = snipNode.ISN(nil, { f }, "$PARENT_INDENT" .. last_line_indent)
 		end
 	end
-	-- if the function is unknown, just insert an empty text-snippet.
-	return tNode.T({ "" })
```
L3MON4D3, line 109: Could you move the existance-check into `functions.var`? Otherwise we get an error for returning nil from a functionNode.

L3MON4D3:
I'd love to have the default variables resolved through namespaces as well.
How about adding one default-namespace which contains all the current variables (and is searched first, with the full varname)?
(Not necessary for this PR)

lua/luasnip/util/parser.lua
leiserfg: That never happens, if the var does not exists `env[var]` will return ""

leiserfg:
> I'd love to have the default variables resolved through namespaces as well. How about adding one default-namespace which contains all the current variables (and is searched first, with the full varname)? (Not necessary for this PR, just want to hear your thoughts)

It can be done for the lazy ones, but the eager ones we have depend on a destructive operation so they have to be done together.

lua/luasnip/util/parser.lua
L3MON4D3: Ahh okay that's nicer👍

L3MON4D3:
> It can be done for the lazy ones, but the eager ones we have depend on a destructive operation so they have to be done together.

You're referring to the Select-variables right? Maybe we should just give namespaces an `eager_fn`/`init` which initializes some variables at eagerly? Not too outlandish that it might be useful outside the default varies too.

leiserfg:
> > It can be done for the lazy ones, but the eager ones we have depend on a destructive operation so they have to be done together.
> 
> You're referring to the Select-variables right? Maybe we should just give namespaces an `eager_fn`/`init` which initializes some variables at eagerly? Not too outlandish that it might be useful outside the default varies too.

That won't solve anyway the problem of those vars that don't follow the `namespace_name` pattern.

L3MON4D3:
> That won't solve anyway the problem of those vars that don't follow the `namespace_name` pattern.

Thats true, for that we could just split resolving the namespace and resolving the variable into separate functions, our default namespace could be named "" and gets would have to be resolved first

leiserfg:
Then I will separate the functions to add lazy and eager vars, store/initialize them accordingly and set native ones as initial values (so I bypass the checks for empty namespaces)

L3MON4D3:
Okay, yeah, that sounds alright :+1:

leiserfg:
I ended up keeping a single function and passing the type as an arg, as most of the time lazy ones are more useful.

L3MON4D3:
Ahh okay, I see.
Tbh my idea was to have the names of eager variables, and additionally the initializer, so it can be used only for the variables which actually need it. Did I miscommunicate, or does it not work as well as this approach? 😅

leiserfg:
> Ahh okay, I see. Tbh my idea was to have the names of eager variables, and additionally the initializer, so it can be used only for the variables which actually need it. Did I miscommunicate, or does it not work as well as this approach? sweat_smile

That's how I implemented it originally, but problem is that it does not work with eager variables that are interdependent (as all we have builtin).

leiserfg:
Given that the eager and lazy vars are stored apart you can always use the same namespace for both, anyway I think most of the use cases for custom vars require only the lazy ones.

L3MON4D3:
> with eager variables that are interdependent

I don't really get the problem, srry, couldn't the interdependent ones just be initialized by the initializer for that namespace (basically the way it's done on master right now) then the named eager ones, and finally the lazy vars on __index?

L3MON4D3:
> Given that the eager and lazy vars are stored apart you can always use the same namespace for both, anyway I think most of the use cases for custom vars require only the lazy ones.

Okay that's true, at least no separate namespaces for eager and lazy variables

leiserfg:
@L3MON4D3  let me know if something is missing here, and if so, how you will like it to be, I don't want this PR to rot.

L3MON4D3:
Yes, right:
Nothings missing, I think, it's just that I'm not completely happy with the current api (I dislike the split into eager and lazy namespaces, and especially, that eager vars can only be functions, and don't understand why that is necessary)
I'd like:
`extend_env(name, values, opts)`, with `name` and `values` just the way they currently are, and `opts` with
- `init : fn(pos) -> map(string -> string/table)`: `values` is extended with the stuff returned here, during `Environ:new`. If `values` is a function that wouldn't work (for that case: maybe it would be better to make the namespace a table either way, and call `values` in its `__index`, that would be just as uniform + each env has a table then)
- `eager_names : map(string -> bool)`: the variable-names listed here are also evaluated during `Environ:new`. I'm aware that this can also be accomplished in `init`, but I think this is an easier and (if one doesn't need the additional capabilities of `init`) easier approach.

I think especially making the namespaces always tables, with an `__index` if `values` is a function, is an important change that could make all of this clean+readable, but please tell me what you think about this, if I'm missing something that makes this infeasible to implement, or if it has some shortcomings I'm unaware of.

L3MON4D3:
Maybe making the env always tables isn't necessary, the results from `init` could also be put into the main-table, with the appropriate prefix (although I'm a bit worried about name-clashing, maybe we should add an mechanism for that, priority or something like that..)

L3MON4D3:
> there are two tables: one for the lazy and one for the eager, in both cases the key is the name of the namespace.

I don't really understand why they need to be split up (it would be better if the entire namespace can be defined in one call I think), could you elaborate why this is necessary and the approach I outlined [here](https://github.com/L3MON4D3/LuaSnip/pull/479#issuecomment-1181514690) (minus turning the function into `__index`, I can see that that is not necessary) does not work?

> Once the Environ is created, all the eager namespaces functions are executed and the variables (named {namespace}_{name}) are stored in the Environ instance. On variable resolution, those that were not resolved (because they are unknow/non eager) Are then resolved by calling the consequent function, depending on the namespace they belong to.

Yup, that's alright, (except, again, that I dislike that eager variables _have_ to be defined in functions and separate from lazy variables)

leiserfg:
I didn't read your previous comment :smile: , do you think isn't better to pass both `values` and `init` as a table? Like:

```lua
ls.env_namespace(name, {
    values:  Map[str,  str|(()->str)] | (str) -> str | nil ,
    init: (pos)-> Map[str, str] | nil,
    eager_vars: [str] | nil
})

```
Internally we keep my last implementation kinda the same way but store the full namespace in a single table from name to namespace. WDYT?

L3MON4D3:
Ahh true, `values` might not even be necessary anymore, `init` could be all that's needed for a namespace.
Yes, having both optional (but at least one) and not mutually exclusive seems like it should cover all bases, while being pretty minimal :+1:

`eager` should be a `string[]`, right?

> Internally we keep my last implementation kinda the same way but store the full namespace in a single table from name to namespace. WDYT?

That sounds good, but I can't promise I won't find anything to nag about :P

> the name of the namespace can't override a built-in one

I was a bit unsure about this one, but I don't think there is any value in changing the default-implementation of those, so yup, good idea to make sure of that.

It might be a good idea to have some kind of warning when an existing namespace is overridden (I'll have to get started on logging sometime soon)

leiserfg:
> `eager` should be a `string[]`, right?
my `[str] ` is your `string[]` :smile: 
Checking if the namespace is already defined is easy to do but I will prefer not to `print` because the most common case will be a user reloading his config, so if we do wanna show that kind of error it will be better an optional centralized log.

L3MON4D3:
> my `[str] ` is your `string[]` :smile:

Ahhh xD

> Checking if the namespace is already defined is easy to do but I will prefer not to `print` because the most common case will be a user reloading his config, so if we do wanna show that kind of error it will be better an optional centralized log.

:+1:x100
Just wanted to have it out there, in case you had some idea, but yeah, logging it is

leiserfg:
@L3MON4D3 check the PR when you have spare time. I didn't do the logging thingy, I think it's better to add the logger tooling apart.

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,139 @@
...
 		end
-		-- If it can't be extracted from lsp, then we use the file path
-		if not ok and not ws_parts then
-			ws_parts = { vim.fn.expand("%:p:h"), vim.fn.expand("%:p:t") }
+
+		for name, val in pairs(eager_vars) do
+			if ns_name ~= "" then
+				name = ns_name .. "_" .. name
+			end
+			if val and type(val) ~= "string" and not Environ.is_table(name) then
```
L3MON4D3, line 69: Here, `Environ.is_table` should just be the regular `type(val) == "table"`, right? Otherwise this would turn all custom vars which return tables into something like this `"table: 0x7f5a539b5258"` afaict

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,139 @@
...
 			end
 		end
-		-- If it can't be extracted from lsp, then we use the file path
-		if not ok and not ws_parts then
-			ws_parts = { vim.fn.expand("%:p:h"), vim.fn.expand("%:p:t") }
+
+		for name, val in pairs(eager_vars) do
+			if ns_name ~= "" then
+				name = ns_name .. "_" .. name
+			end
```
L3MON4D3, line 68: I'd like this a tiny bit more if a `prefix` was assigned outside the inner loop, and just prependedd to `name`.
(just like to avoid `if` inside `for` :sweat_smile:)
What do you think?

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,139 @@
...
-				ws_parts = { ws, file_path:sub(#ws + 2, -1) }
-				break
+	for ns_name, ns in pairs(namespaces) do
+		local eager_vars = {}
+		if ns.init then
+			eager_vars = ns.init(pos)
+		end
+		for _, eager in ipairs(ns.eager) do
+			if not eager_vars[eager] then
+				eager_vars[eager] = ns.vars(eager)
```
L3MON4D3, line 61: Here it wouldn't make a difference (well, for our builtin vars at least) if the existing `eager_var` would be overwritten, correct? (don't necessarily want to change this, just to be sure)

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,139 @@
...
+		ns and type(ns) == "table",
+		("Your namespace '%s' has to be a table"):format(name)
+	)
+	assert(
+		ns.init or ns.vars,
+		("Your namespace '%s' needs init or vars"):format(name)
+	)
+	assert(
+		not namespace.eager and ns.vars,
+		("Your namespace %s can't set a `eager` fild without the `vars` one"):format(
```
L3MON4D3, line 106: typo: an `eager` field

lua/luasnip/util/_builtin_vars.lua
```diff
@@ -0,0 +1,180 @@
...
+end
+
+local builtin_ns = {}
+
+for name, _ in pairs(lazy_vars) do
+	local parts = vim.split(name, "_")
+	if #parts > 1 then
+		builtin_ns[parts[1]] = true
+	end
+end
```
L3MON4D3, line 169: Not sure if `SELECT` is present here, I don't think it is?

lua/luasnip/util/environ.lua
leiserfg: I thought we were using `is_table` to check when expanding, but now that I check is only for the fake expansion, so I will remove the check.

lua/luasnip/util/environ.lua
leiserfg: Yes, given that `init` and `eager` are potentially capable of providing the same vars and that they don't have to generate the same value, I had to decide to give priority to one of them. If you don't like that choose let me know, but it's one set or the other.

L3MON4D3: Ah, right, I see now that this is just namespace-local.
This should work alright :+1:

lua/luasnip/util/environ.lua
L3MON4D3: Just to expand on that, `fake_expand` (used for docstring, for example) uses a modified `env` which just returns the varname for any variable (`env.TM_SELECTED_TEXT` -> `$TM_SELECTED_TEXT`). Now, for functions to still work correctly we have to return the same type in the fake-expanded snippet.
I don't know how to get that info for custom-env-snippets dynamically, maybe we have to require another `eager`-like-table, which declares which vars are tables

L3MON4D3: Seriously considering requiring that variables are always `string[]` (breaking change, but will make lots of code simpler (no more `if type(var) == "string" then ... else ... end`) and just remove this problem)

lua/luasnip/util/_builtin_vars.lua
```diff
@@ -158,7 +158,7 @@ local function eager_vars(pos)
     return vars
 end
 
-local builtin_ns = {}
+local builtin_ns = {SELECT: true}
```
L3MON4D3, line 161: `:`? :P

lua/luasnip/util/environ.lua
L3MON4D3: > If you wanna be able to always show the type, then namespaces will have to be able to provide the type dynamically as well, so it will have to be a "method" on the namespace, and expose the result via `Environ.is_table`

Oh, that's true, but that's really ugly, isn't it?

> f so wdyt if we do it in a different PR (as it does not break the namespaces nor the fake_expand features)

I'd love to, but `fake_expand` will break if 
1. a custom namespace returns a table for some var
2. and that var is used like `env.MY_CUSTOM_TABLE_VAR[1]` in a function/dynamicNode

(since the `__index` from `fake_expand` will return a string, not a table)

What's your opinion on the breaking change?

L3MON4D3: (okay, not `fake_expand` will break, `get_docstring` might)

leiserfg: True, I forgot that they can be accessed that way too :thinking: but anyway even with a method or a table for doing it, users could always use the namespace wrong, like the creator of the namespace is thinking a var will be used as a string and the user uses it as a table, or the same namespace creator returning the incorrect type, so I'm not sure what to do about it.

leiserfg: What do you think if the `fake_env` vars return always a table with an overridden `__concat` so it behaves as a string? too?

leiserfg: What do you think if the fake_env vars return always a table with an overridden __concat so it behaves as a string too?

L3MON4D3: > What do you think if the `fake_env` vars return always a table with an overridden `__concat` so it behaves as a string too?

That would be pretty cool, overriding `__concat` (and `__tostring`) should cover most bases.
But I'm not sure if it's really the best solution, there probably are still exceptions which, however rare, would break (probably creating some confusion along the way) (I'm thinking of stuff like indexing into a table with the value, that would fail no?)

leiserfg: Well, yes but then there is no way to fix it besides rendering the environment for real, which is not a good solution either, as namespaces can have secondary effects. So I think is better to make it work for the majority of the users at the cost of facing some issues because of missuses. In any case the worst that can happen is that a `f` or a `d` node felling on the preview, but that's something that can happen with any other misuse isn't it?

L3MON4D3: > Well, yes but then there is no way to fix it besides rendering the environment for real, which is not a good solution either, as namespaces can have secondary effects.

We could require the variables to always be `string[]` (`[str]` :P), that would also prevent this  issue.
But shy of that, yes, overriding `__concat` would work just fine.

>  but that's something that can happen with any other misuse isn't it?

Yeh, also true, but this is something we can prevent, so we should I think.

leiserfg: I just added a fake environment that implements `__concat` and `__tostring`

L3MON4D3: Very clean, that should work if we decide to do it that way :+1:
But I'm not sure we should: a clean break, and after it all env-vars are always `string[]`, seems much cleaner than worrying about what is `tostring`d and what isn't.

Also (and I should have realized this sooner), this will only affect the receiving end, we can still allow simple strings as `lazy_vars`, just silently convert it to `string[]` before returning it from `__index` and we should be good to go.

Also (if we do a breaking change), the transition will be very smooth (since it just means some variables need to be indexed with `[1]`), so that should not be a big issue.

Do you see any issues with making the variables uniform? I guess it will make functions that use `env` a _tad_ more verbose, but on the flipside, if `env.SOMEVAR` can be counted on being a table, that could make some stuff easier (if a plugin wants to provide env-variables, to do it properly, there might be a need to list what each var returns (`string` or `string[]`)).
One downside I can see is that there are just very few env-variables which contain multiline text, so the many have to bend to the pressure of the few, which isn't that nice, but I feel like it's the best option we have

leiserfg: My concern with that is that it will break all the lua snippet in the while that are already using the vars as strings in any nonsimple way, and given that there are more string vars than lists vars that will bring way more errors than the other approach I assume.

L3MON4D3: Yeah, it might break a fair bit of snippets (I don't think that many, most usage of `env` is probably through `lsp` and `lambda`), but once the bandaid is off luasnip is in a better state than before.
We could also try a deprecation-period with `__concat`, it's a pretty good solution in 99%, but I think tables-only makes more sense in the long run (simpler on our side, no surprises in dynamic/functionNodes when their docstrings are generated).

L3MON4D3: I'm starting to like the metatable more than every-var-is-a-table, it's more user-friendly (well, to the 99%), and the errors caused by it can be silences via `tostring`. We just need a good way to point users unfamiliar with these errors into the right direction (so probably a big warning somewhere in DOC.md for now, logging the errors later).
One more thing: this can't currently handle `env.somevar:match()` or other string-calls (which, I imagine, also make up a big chunk of what's done with the variables). We could override `__index` and return wrapper-functions around the string-methods, which take our fake_env-var and pass its actual real content to the string-method (and returns the string-method's result)

lua/luasnip/util/environ.lua
```diff
@@ -125,6 +125,19 @@ local fake_var = {
...
+    end,
+    __index = function(obj, key)
+        local wrapped_str = obj[1]
+        local val = wrapped_str[key]
+        if type(val) == "function" then
+            return function (x, ...)
+                if x == obj then
+                    x = wrapped_str
+                end
+                return string[key](x, ...)
```
L3MON4D3: Indexing val instead of immediately `string` is a nice idea :+1:
I think `val` instead of `string[key]` would be slightly more coherent, wdyt?

L3MON4D3: Also +1 for the `x==obj`-check, that might prevent some (okay, unlikely, but better to be safe here) errors.

lua/luasnip/nodes/snippet.lua
```diff
@@ -689,12 +689,7 @@ function Snippet:fake_expand(opts)
 	if opts.env then
 		self.env = opts.env
 	else
-		self.env = {}
-		setmetatable(self.env, {
-			__index = function(_, key)
-				return Environ.is_table(key) and { "$" .. key } or "$" .. key
-			end,
-		})
+		self.env = Environ.fake()
```
L3MON4D3, line 692: Oh right, also good that you moved this into `environ.lua`, much better than creating it in such a disconnected place.

L3MON4D3:
We'll need some tests to assure that docstring-generation still works with custom-namespaces, just assure that concatenation and `env.VAR:somestringfunc` actually work correctly.

Also: my dismissal of using a function for determining which env-variables return tables and which return strings (the function which would be passed to `env_namespace`) was a bit premature, if we allow it to be also table (instead of just function), it would be a good interface.
I feel it's better to leave some way to explicitly tell luasnip which vars are tables than always getting the (potentially unexpectedly behaving) string-simulating-table (which is a good fallback, I'm not suggesting to not use it).
What do you think?

leiserfg:
To be honest, I think that having table variables is indeed the problem, if every var where always a string (maybe with some `\n` and the conversion to a table were done on render time), all these workarounds were unneeded.

L3MON4D3:
Yes, that's true.. Changing that would be a massive breaking change thoigh, I'm not sure we want to do that.
Tbh, I also rather like multiline strings as tables and don't want to remove them, but it's probably the better choice.

Difficult decision, I think I'd want to move that to a later date, when there's a standard for semantic versioning (or we have something that's firmly established, so there's not that many ppl just following master).

But apart from that, do you have doubts about optionally specifying types (string[] vs just string) in `env_namespace`?

leiserfg:
So you mean to add a field `is_table: (str)->bool | [str]` with the default `[]` (all are not tables), and then use them in `Environ.is_table` ? If that's ok I can add it. Regarding the tests do we have a way of testing that does not require having to build nvim locally every time?

leiserfg:
In that case, I will remove the `table-string` hybrid and use the type announced by the matching environment.

L3MON4D3:
My thought was to allow a key `is_table : (str)->bool | [str] | nil` with
* `nil`: fall back to `table-string`-hybrid
* `map`: create function from map, similar to how `lazy_env` is converted to function
* `fn`: just use it
The key being that we don't default to "everything string", but to "table-string-hybrid". That way Sources can still work, even if they return tables.

On the other hand, it would be less complicated (I think for us, and users) if there was just a requirement to announce which variables are tables. For us, since there's much fewer moving parts, for users because there can't be weird errors from "table-string", and it should be more clear what is going on.
Requiring it shouldn't add much boilerplate, in most cases it's going to be an empty table.
If we don't, it can just be ignored and lead to errors.

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,180 @@
...
+	assert(
+		ns and type(ns) == "table",
+		("Your namespace '%s' has to be a table"):format(name)
+	)
+	assert(
+		ns.init or ns.vars,
+		("Your namespace '%s' needs init or vars"):format(name)
+	)
+	assert(
+		not namespace.eager and ns.vars,
```
L3MON4D3, line 109: I think this is a bit hard to decipher (at least for me), what do you think of a (well, slightyl) more explicit `namespace.eager and ns.vars or true` (I wish, for the first time, that "implies" was implemented in more (any?) languages)
(or maybe just add a comment that `not namespace.eager and ns.vars` is `implies`)

leiserfg: > 

That's the same I said 😂

L3MON4D3: Yeah pretty much I know xD
Just wanted to list it all again so it's clear but still show the difference with `nil`

L3MON4D3: (The difference being "no default", but now I doubt it's the best design, a default might be better)

L3MON4D3:
Nice, also allowing false makes sense :+1:
I'd change `is_table` to `table_vars` (or maybe `multiline_vars`, and add in the doc that only those vars may be tables? I think I'd prefer this, wdyt?)

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,197 @@
...
-	local function subs(c)
-		local v = (((c == "x") and random(0, 15)) or random(8, 11))
-		return string.format("%x", v)
+function Environ:__index(key)
+	local nmsp, varname = _resolve_namespace_var(key)
+	---@diagnostic disable-next-line: need-check-nil
+	local val = nmsp.vars(varname)
+
+	if val then
+		val = tostring(val)
```
L3MON4D3, line 163: This could produce unwanted results if `val` is a table ("table 0xsomething"). Is the `tostring` even necessary? Imo it's the namespaces responsibility to actually just return string/string[].

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,197 @@
...
-end
-function lazy_vars.CURRENT_SECONDS_UNIX()
-	return tostring(os.time())
+	return o
 end
 
--- For inserting random values
+local builtin_ns_names = vim.inspect(vim.tbl_keys(builtin_vars.builtin_ns))
+
+function Environ.env_namespace(name, namespace)
```
L3MON4D3, line 85: `namespace` -> `opts`?

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,193 @@
...
+		eager = {},
+		is_table = builtin_vars.is_table,
+	},
+}
 
-function Environ:override(env, new_env)
-	for k, v in pairs(new_env) do
-		env[k] = v
+-- Namespaces allow users to define their own environmet variables
+local function _resolve_namespace_var(full_varname)
```
L3MON4D3, line 25: How are lazy variables from the "" namespace resolved right now? In case they're not, checking them before trying other namespaces seems the right approach.

L3MON4D3:
I'd love if the internal variables also went through the `env_namespace`-function, that might prevent errors later on (for example if the function is changed, one might forget to also update the directly injected namespace + the module can serve as an example).
Maybe add an undocumented key to opts (`__allow_empty_namespace_name`)?

lua/luasnip/util/environ.lua
leiserfg: They are checked after because they can't be overridden so it's safe.

L3MON4D3: Ahh, L37, I'm blind :sweat_smile:

leiserfg:
> I'd love if the internal variables also went through the `env_namespace`-function, that might prevent errors later on (for example if the function is changed, one might forget to also update the directly injected namespace + the module can serve as an example). Maybe add an undocumented key to opts (`__allow_empty_namespace_name`)?

What if I expose a function with checks and use another internally without them ? I hate passing hidden args,they can be found and exploited.

L3MON4D3:
Yeah, that'll work :+1:
(but if there is some undocumented variable starting with a ton of underscores, that's a big sign this isn't meant for the user, so anything going wrong is on them)

I think `environ` should export both versions, so that both definition and addition (? the call to `builtin_namespace_env`) of the builtin vars can be in the same file, we can export the one with checks in eg. `require("luasnip")`

L3MON4D3:
Okay, I don't have any more complaints about the API+implementation, I'll take a look at some minor things.
About tests: (and I completely forgot to answer this) yeah, there's no way around building neovim, we rely on their test-helpers :/
(You can just build it once and remove that part from the makefile (locally) though)

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,193 @@
...
--- Variables defined in https://code.visualstudio.com/docs/editor/userdefinedsnippets#_variables
+	local varname
 
--- Inherited from TextMate
-function lazy_vars.TM_FILENAME()
-	return vim.fn.expand("%:t")
+	if nmsp then
+		varname = full_varname:sub(#parts[1] + 2)
+	else
+		nmsp = namespaces[""]
```
L3MON4D3, line 30: Comment to clarify that it's impossible to produce clashes with builtin variables, and they therefore they can be checked last?

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,193 @@
...
 
-		for _, ws in pairs(vim.lsp.buf.list_workspace_folders()) do
-			if file_path:find(ws, 1, true) == 1 then
-				ws_parts = { ws, file_path:sub(#ws + 2, -1) }
-				break
+		for name, val in pairs(eager_vars) do
+			name = prefix .. name
+			local val_type = type(val)
+			if val and val_type ~= "string" and val_type ~= "table" then
+				val = tostring(val)
```
L3MON4D3, line 68: Can this `tostring` be removed, like the one in `__index`?

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,193 @@
...
 
-function lazy_vars.RELATIVE_FILEPATH() -- The relative (to the opened workspace or folder) file path of the current document
-	return buf_to_ws_part()[2]
-end
+local builtin_ns_names = vim.inspect(vim.tbl_keys(builtin_namespace.builtin_ns))
+
+local function _env_namespace(name, opts)
+	assert(
+		opts and type(opts) == "table",
+		("Your namespace '%s' has to be a table"):format(name)
```
L3MON4D3, line 81: `namespace` -> `opts`? (should be understandable since the second arg is called just that in the documentation)

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,193 @@
...
+	opts.eager = opts.eager or {}
+	local multiline_vars = opts.multiline_vars or false
+
+	local type_of_it = type(multiline_vars)
+
+	assert(
+		type_of_it == "table"
+			or type_of_it == "boolean"
+			or type_of_it == "function",
+		("Your namespace %s can't have `is_table` of type %s"):format(
```
L3MON4D3, line 105: `is_table` -> `multiline_vars`

DOC.md
```diff
@@ -1094,6 +1094,47 @@ To use any `*SELECT*` variable, the `store_selection_keys` must be set via
 hitting `<Tab>` while in Visualmode will populate the `*SELECT*`-vars for the next
 snippet and then clear them.
 
+## Environemt Namespaces
+
+You can also add your own variables by using the `ls.env_namespace(name, opts)` where:
+
+* `name`  is string that does not contains the character "_" and names the namespace.
+* `opts` is a table containing:
+    * `vars`
```
L3MON4D3, line 1103: How about stating the type of these arguments explicitly (+give function-prototype, for functions)? I think that might make it much easier to grasp at a glance

DOC.md
```diff
@@ -1094,6 +1094,47 @@ To use any `*SELECT*` variable, the `store_selection_keys` must be set via
...
+
+The four fields of `opts` are optional but you need to provide either `init` or  `vars`, and `eager` can't be without `vars`.
+Also you can't use namespaces that override default vars.
+
+
+A simple example to make it more clear:
+
+```lua
+local random_lang()
+    return ({"LUA", "VIML", "VIML9"})[math.floor(math.random()/2 + 1.5)]
```
L3MON4D3, line 1122: Oh very nice, I didn't even see it until now xD
Is `math.random` 1-inclusive? Otherwise this would be _very_ unfair towards viml :P

lua/luasnip/util/environ.lua
```diff
@@ -1,190 +1,193 @@
...
+	_env_namespace(name, opts)
 end
 
-function lazy_vars.RANDOM_HEX()
-	return string.format("%06x", math.random(16777216)) --16^6
-end
+function Environ:__index(key)
+	local nmsp, varname = _resolve_namespace_var(key)
+	---@diagnostic disable-next-line: need-check-nil
+	local val = nmsp.vars(varname) or ""
```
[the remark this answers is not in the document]
leiserfg: I think it's safer to return `""` it could hide a wrong variable but will avoid breaking the contract that the variable is a string or a list. One way or the other the value has to be shown as an empty string.

DOC.md
leiserfg: As it is it gives always "LUA"

L3MON4D3: Perfect :ok_hand:

DOC.md
leiserfg: Not sure what you mean by explicitly, type signatures? if is that the case what format should we use?

lua/luasnip/util/environ.lua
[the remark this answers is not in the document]
L3MON4D3: I agree, it's probably safer, but I just read up on this in the lsp-snippet-spec:
"Variables
With `$name` or `${name:default}` you can insert the value of a variable. When a variable isn’t set, its default or the empty string is inserted. When a variable is unknown (that is, its name isn’t defined) the name of the variable is inserted and it is transformed into a placeholder."
(Variable with default: `${VARIABLE: this is the ${1:default}}`)
So we'll need some mechanism to determine whether a variable is not set (is an empty string, according to my short experimentation with vscode) or is just not defined.
I think both a separate function and an overload to return nil are alright (slight tendency toward having `__index` return `string|string[]|nil`, I don't think we have to worry about incorrectly named variables in lua-code)

DOC.md
L3MON4D3: Yeah, type signatures :+1:
So far I've used `fn(arg1name: arg1type, arg2name: arg2type) -> returnType`, different possiblities as `(type1|type2|type3)` (I _think_ I'm consistent with that one)

lua/luasnip/util/environ.lua
L3MON4D3: Hmm I think I looked into that, it has some basic stuff, but we need to have some way to check the screens contents (for integration-tests, which make lots of sense for us), and neovim has some pretty great helpers for that, and I'd like to use those (haven't looked at just extracting them out of neovim, but I'd imagine that's a bit of a hassle)
It's not the most minimal testing-setup, but it works :sweat_smile:

For me it means waiting a minute once in a while, is it more annoying for you?

lua/luasnip/util/environ.lua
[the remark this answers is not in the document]
L3MON4D3: Ah, true, I'll take care of the parser-modifications, no need to do that here :+1:
(And yup, dynamicNode is the way forward for variables)

DOC.md
```diff
@@ -1094,6 +1094,47 @@ To use any `*SELECT*` variable, the `store_selection_keys` must be set via
 hitting `<Tab>` while in Visualmode will populate the `*SELECT*`-vars for the next
 snippet and then clear them.
 
+## Environemt Namespaces
```
L3MON4D3, line 1097: typo, `Environment` (didn't spot this one before :sweat_smile:)

L3MON4D3:
Okay I think apart from tests and that one typo this is good to go :+1:

leiserfg:
> 

Tomorrow I will add the tests.

lua/luasnip/util/environ.lua
leiserfg: To me, the language was not a problem, but the fact that the doc is scattered and the fact that not every package is packaged the same way. So I have to check the package definitions every time I need to customize something, but hey at least I can do that, that's what I like about the distro.

L3MON4D3: Sounds a bit annoying, but probably worth it for all the good stuff nix brings to the table. Probably also just some growing-pains, I hope it becomes more mature quickly, I want to try it sometime

leiserfg:
Done!

L3MON4D3:
Those are great👍
Could you add some more to cover the fake_env and usage of new vars in lsp?

leiserfg:
> Those are great+1 Could you add some more to cover the fake_env and usage of new vars in lsp? I could add for `fake_env` but for the lsp it should be an integration test isn't it?

L3MON4D3:
I think both are good fits for integration tests actually

L3MON4D3:
Nice, what do you think of a dynamicNode which indexes a new table-var, and then calling `get_docstring` (there's a helper for that) for the `fake_env`-test?

leiserfg:
Is there something else missing?

L3MON4D3:
One last look, but I think this is it :tada: :tada:

L3MON4D3:
Okay yup, I've got nothing more about the content, this is really good now, great job :+1::+1::heart:

About the commits: It's alright with me if they stay as-is, with the exception of the "Address Comments"-ones, could you give them commit messages that are also meaningful outside the context of this PR? (You can also just squash a few, or all, if you prefer that, your call)

leiserfg:
I rebased it on top of master and renamed/squashed  some commits

L3MON4D3:
Okay perfect :+1:
Thank you very much for being so patient and sticking through, I know I demand lots of changes :sweat_smile:
But here, I think the results speak for themselves, great work :+1:

PR #502, opened 2022-07-31
feat: add fallback if no paths are currently loaded

pianocomposer321:
This PR adds a fallback if no files are found in the current cache. This allows configs like the following:

```lua
require("luasnip.loaders.from_snipmate").lazy_load()

vim.api.nvim_create_user_command("LuaSnipEdit",
  function()
    require("luasnip.loaders").edit_snippet_files {
      fallback = function(ft, edit)
        if #ft > 0 then
          local filename = "~/.config/nvim/snippets/" .. ft .. ".snippets"
          print("Creating new file: " .. filename)
          edit(filename)
        else
          print("Canceling")
        end
      end
    }
  end, {})
```

Pretty minor change, should be self explanatory - basically it calls the fallback function, passing in the chosen filetype and the edit function. Main use case is to add file for current filetype if it does not already exist, but could be used for other things as well.

I made this because there was no easy way to tell if a file already exists and add it if it doesn't (at least not that I could tell). If I'm mistaken about this, I'd be interested in what the current preferred method is for this.

If I need to add docs or anything, just let me know!

### DIFF
--- a/lua/luasnip/loaders/init.lua
+++ b/lua/luasnip/loaders/init.lua
@@ -33,6 +33,9 @@ function M.edit_snippet_files(opts)
 	opts = opts or {}
 	local format = opts.format or default_format
 	local edit = opts.edit or default_edit
+	local extend = opts.extend or function()
+		return {}
+	end
 
 	local fts = util.get_snippet_filetypes()
 	vim.ui.select(fts, {
@@ -51,6 +54,17 @@ function M.edit_snippet_files(opts)
 						table.insert(items, fmt_name)
 					end
 				end
+			end
+
+			-- extend filetypes with user-defined function.
+			local extended = extend(ft, ft_paths)
+			assert(
+				type(extended) == "table",
+				"You must return a table in extend function"
+			)
+			for _, pair in ipairs(extended) do
+				table.insert(items, pair[1])
+				table.insert(ft_paths, pair[2])
 			end
 
 			-- prompt user again if there are multiple files providing this filetype.
--- a/DOC.md
+++ b/DOC.md
@@ -2041,7 +2041,7 @@ and then (if there are multiple) the associated file to edit.
 
 <!-- panvimdoc-ignore-end -->
 
-`opts` currently only contains one setting:
+`opts` contains three settings:
 
 * `format`: `fn(file:string, source_name:string) -> string|nil`  
   `file` is simply the path to the file, `source_name` is one of `"lua"`,
@@ -2074,6 +2074,32 @@ and then (if there are multiple) the associated file to edit.
   editing. The default is a simple `vim.cmd("edit " .. file)` (replace the
   current buffer), but one could open the file in a split, a tab, or a floating
   window, for example.
+* `extend`: `fn(ft:string, ft_paths:string[]) -> (string,string)[]`  
+  This function can be used to create additional choices for the file-selection.
+
+  * `ft`: The filetype snippet-files are queried for.
+  * `ft_paths`: list of paths to the known snippet files. 
+
+  The function should return a list of `(string,string)`-tuples. The first of
+  each pair is the label that will appear in the selection-prompt, and the
+  second is the path that will be passed to the `edit()` function if that item
+  was selected.
+
+  This can be used to create a new snippet-file for the current filetype:
+```lua
+require("luasnip.loaders").edit_snippet_files {
+  extend = function(ft, paths)
+    if #paths == 0 then
+      return {
+        { "$CONFIG/" .. ft .. ".snippets",
+          string.format("%s/%s.snippets", <PERSONAL_SNIPPETS_FOLDER>, ft) }
+      }
+    end
+
+    return {}
+  end
+}
+```
 
 One comfortable way to call this function is registering it as a command:
 ```vim


L3MON4D3:
Hey, thank you for participating, always happy to see some PRs from new ppl :)

Creating new snippet-files via `:LuaSnipEdit` is a great idea, should improve that workflow :+1:

One downside I can see with this implementation is that it is not possible to add a personal snippet-file if there already is one for this filetype, provided by some collection.

What do you think of, instead of `fallback`, providing a more general `extend`-callback, which gets a list of the files found by luasnip (`ft_paths`), and can return additional ones (as a list of `{item_name, path}`-pairs), which will then also be displayed in the second `vim.ui.select`-dialog?

For creating a new file, one could check the passed paths, if there is no item from the personal snippet-collection, an item for creating that file can be returned, maybe even with a hint that this selection creates a new file.

pianocomposer321:
How does this look?

One thing I would say to note is that line 58 in loaders/init.lua will result in an error if the user-provided function does not return a table (e.g. it returns nil). This could be changed by adding `... or {}` to the end of the line, but I thought it was probably best to leave it as is rather than force ignoring the type-error. But I'm open to your ideas about this.

pianocomposer321:
BTW, this is an example of how I'm using it for reference:

```lua
vim.api.nvim_create_user_command("LuaSnipEdit", function()
  require("luasnip.loaders").edit_snippet_files {
    extend = function(ft, paths)
      if #paths == 0 then
        return {
          { "$CONFIG/" .. ft .. ".snippets", string.format("%s/%s.snippets", pc_path, ft)}
        }
      end

      return {}
    end
  }
end, {})
```

L3MON4D3:
> How does this look?

Looks great, I've no complaints :+1:
Could you add a description of what extend does (and maybe an example of what it can be used for) [here, in `DOC.md`](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#edit_snippets) (oh, and update that line stating that `opts` only contains one setting :sweat_smile:)?

> One thing I would say to note is that line 58 in loaders/init.lua will result in an error if the user-provided function does not return a table (e.g. it returns nil). This could be changed by adding `... or {}` to the end of the line, but I thought it was probably best to leave it as is rather than force ignoring the type-error. But I'm open to your ideas about this.

It's okay, as long as the doc specifies that `extend` should return a table that shouldn't matter. You could add an assert that `extended` actually is "table" though, directly point out misuse.

pianocomposer321:
Sounds good. I won't be able to get to this for a while (probably at least a week and a half), just so you're aware why there isn't any progress.

L3MON4D3:
Take your time :+1:

L3MON4D3:
Hey, will you have time to finish this feature in the near future?
(No worries if not, I'll take over and complete this, I think it's a nice addition)

pianocomposer321:
> Hey, will you have time to finish this feature in the near future? (No worries if not, I'll take over and complete this, I think it's a nice addition)

It's still my intention to get to it eventually, but I'm not sure when. I was on vacation in August, and life's been busy since getting back. TBH I haven't turned my workstation computer on since I got back.

So yes, but I'm not sure I can give you a timeline. I won't be offended if you just take over now, but I'm still willing to get to it eventually if you don't.

Completely up to you.

L3MON4D3:
Ahh, alright :D
I'm in no real hurry to get this in, so you can continue as soon as you find time for it
(just wanted to make sure you're still up for finishing this :sweat_smile:)

pianocomposer321:
Updated docs, added assert for `extended`. Is there anything else?

L3MON4D3:
Ah nice :+1:
I'll have look

lua/luasnip/loaders/init.lua
```diff
@@ -52,6 +55,17 @@ function M.edit_snippet_files(opts)
 				end
 			end
 
+			-- extend filetypes with user-defined function.
+			local extended = extend(ft, ft_paths)
+			assert(
+				type(extended) == "table",
+				"You must return a table in exted function"
```
L3MON4D3, line 62: typo: extend

doc/luasnip.txt
```diff
@@ -1569,6 +1569,23 @@ there are multiple) the associated file to edit.
...
+- `extend`: `fn(ft:string, default_paths:table)` -> `table`
+    Function returning any additional locations in which to look for snippet files. `ft` is the current filetype, `ft_paths` is the list of paths already found. One usecase for this would be in the creation of user-defined snippet files. For example:
+
+
+>
+    require("luasnip.loaders").edit_snippet_files {
+      extend = function(ft, paths)
+        if #paths == 0 then
+          return {
+            { string.format("%s/%s.snippets", PERSONAL_SNIPPETS_FOLDER, ft) }
```
L3MON4D3, line 1581: Don't we expect a `(string,string)[]` as the result of `extend`?

doc/luasnip.txt
```diff
@@ -1569,6 +1569,23 @@ there are multiple) the associated file to edit.
     editing. The default is a simple `vim.cmd("edit " .. file)` (replace the
     current buffer), but one could open the file in a split, a tab, or a floating
     window, for example.
+- `extend`: `fn(ft:string, default_paths:table)` -> `table`
+    Function returning any additional locations in which to look for snippet files. `ft` is the current filetype, `ft_paths` is the list of paths already found. One usecase for this would be in the creation of user-defined snippet files. For example:
```
L3MON4D3, line 1573: current -> selected?

doc/luasnip.txt
```diff
@@ -1569,6 +1569,23 @@ there are multiple) the associated file to edit.
     editing. The default is a simple `vim.cmd("edit " .. file)` (replace the
     current buffer), but one could open the file in a split, a tab, or a floating
     window, for example.
+- `extend`: `fn(ft:string, default_paths:table)` -> `table`
```
L3MON4D3, line 1572: default_paths: `table` -> `string[]`?
And `(string,string)[]` for the return value?

Also, I think mentioning how exactly the return values will be used would be nice
(first -> item-name, second -> item-path basically, more words ofc :D)

pianocomposer321:
How's this?

L3MON4D3:
Very nice, great work :+1:
Could you squash the commits?

L3MON4D3:
(oh, and maybe rebase, that would be nice :D)

pianocomposer321:
Lol, I don't actually know how to do that...I'm still learning how to use git properly. I will google it and get back with you.

Or you're welcome to tell me how to do it too, that way I know for sure I'm doing the right thing. Up to you.

L3MON4D3:
(I'd do that stuff myself, but if I rebase, the commits would seem to be made by me, not you)

pianocomposer321:
Ok, finally got around to this. Sorry for the wait (again 👀).

Did I do everything right? Is there anything else you need before merging?

L3MON4D3:
Oh, I think something went wrong, theres some difference between this branch and master 250 commits ago, which should not be the case.
I can take a look at this tomorrow

leiserfg:
I think that the merge commit is breaking the history, it will be cleaner to remake the branch starting from the current master and cherry-pick the commit with the wanted changes only.

L3MON4D3:
That did it :+1:
I also reworded the doc a bit, @pianocomposer321 if you're happy with that I'll go ahead and squash->merge

pianocomposer321:
Looks good to me 👍

pianocomposer321:
> I think that the merge commit is breaking the history, it will be cleaner to remake the branch starting from the current master and cherry-pick the commit with the wanted changes only.

Also sorry about this. Glad you figured it out ;-).

L3MON4D3:
Wonderful!
Took a while, but we're there now :D
Thank you!

pianocomposer321:
Awesome! Sorry again for the long waiting periods and for the noobishness lol. Thanks for bearing with me...this was the first major project I've contributed to, and it was a great first experience. I learned a lot. Thanks again!

L3MON4D3:
Oh wow, thank you, that's great to hear :D
Don't worry about noobishness (and waiting periods especially, we're all doing this in our free time, some have more, some less), the actual code was pretty much perfect after your first iteration, it's just the organizational (documentation+git) stuff you were "noobish" at all in. I'm certain especially those will get easier with more experience :)
Looking forward to more contributions from you, don't hesitate if you have an idea/are interested in helping out on issues

PR #512, opened 2022-08-08
Handle empty table in inputs

leiserfg:


leiserfg:
Maybe it should go in the `wrap_value` but I'm not sure If it will break something else.

leiserfg:
See  #512

leiserfg:
I checked it breaks all the nodes, I will move the fix to the node itself.

L3MON4D3:
Moving that to `wrap_value` sounds like a good solution, I think it's purpose was handling those functions that can return either `string` or `string[] `.
On the go at the moment, so I'll take a closer look this evening

lua/luasnip/nodes/snippet.lua
```diff
@@ -181,7 +181,7 @@ local function init_snippet_context(context)
 
 	-- context.dscr could be nil, string or table.
 	context.dscr =
-		util.to_line_table(util.wrap_value(context.dscr or context.trigger))
+		util.to_line_table(context.dscr or context.trigger)
```
L3MON4D3: This wasn't supposed to be changed, right?

lua/luasnip/util/util.lua
```diff
@@ -233,12 +233,22 @@ local function put(text, pos)
 	pos[2] = (#text > 1 and 0 or pos[2]) + #text[#text]
 end
 
--- Wrap a value in a table if it isn't one already
+--[[ Wraps the value in a table if it's not one, makes
+  the first element an empty str if the table is empty]]
 local function wrap_value(value)
```
L3MON4D3, line 238: How about renaming this function to just `wrap_string` or `to_string_table` (I think it's only used for that right now either way, but better to be specific)?

lua/luasnip/nodes/snippet.lua
leiserfg: yes `to_line_table` calls wrap_value (now  `to_string_table`) internally already, it was a duplicated call.

L3MON4D3: Ohhh, okay, nvm :+1:

lua/luasnip/util/util.lua
```diff
@@ -233,12 +233,22 @@ local function put(text, pos)
...
-		return value
+--[[ Wraps the value in a table if it's not one, makes
+  the first element an empty str if the table is empty]]
+local function to_string_table(value)
+	if not value then
+		return { "" }
 	end
-	return { value }
+	if type(value) == "string" then
+		return { table }
```
L3MON4D3, line 243: `table` -> `value`? :D
Was just looking into why tests are failing and imagine my surprise when a string was wrapped to a table somehow containing functions xD

leiserfg: Don't program if you are sleepy my friend :facepalm:

L3MON4D3:
Okay, nice, code-wise this is perfect :+1: 
Could you add a test that fails if we regress here?
imo it's enough if there's just one for this specific issue, eg. empty table in insertNode, that should be the most common usage anyway.

L3MON4D3:
Okayyyyy nice, thank you very much :+1:

Issue #482, opened 2022-07-10
Add key to autoexpand snippets instead of having two separate tables


aikow:
I'm currently trying to replicate my latex snippets using lua snip that I've previously had in ultisnips. Ultisnips has the snippet options, where you can simply append an `A` after your snippet to have it automatically expand.

I've been using this for snippets like 
```
snippet lr "left right parenthesis" i
\left( $1 \right)$0
endsnippet

snippet () "left right parenthesis" iA
\left( $1 \right)$0
endsnippet
```
In luasnip, when loading these from a lua file, I have to place them in completely different parts of the document, since they have to be returned in 2 separate tables, which feels kind of clunky.
I could, since the snippets are just lua code, add a bunch of table appending, extending, and manipulation code, but I feel like it shouldn't be to hard to just add it as an option to the snippet node, although I haven't looked at the code behind autosnippets yet.

Thanks for the cool library :grin:.

L3MON4D3:
> I have to place them in completely different parts of the document, since they have to be returned in 2 separate tables, which feels kind of clunky

>  feel like it shouldn't be to hard to just add it as an option to the snippet node

Hmmm, that's true.

The reason they have to be defined in separate tables is just an artifact of the way autosnippets were added to luasnip before `add_snippets`, which was done through a separate table `ls.autosnippets`.
But now, with `add_snippets`, this limitation is gone, and not having any restraints on where snippets are defined is good, so I guess this should be added :+1:
It should be enough to add `autosnippet` as another option in the first table passed to the snippet, similar to `regTrig`, I feel like usage would be a tad nicer with a separate `autosnippet`-function, but that can easily be implemented on top of the option.

aikow:
Yeah, that sounds exactly like what I had in mind :).

Maintaining backwards compatibility should be simple enough as well, if thats desirable, by just adding the `autotrigger = true` to the second table returned by the snippets file.

L3MON4D3:
Right that would work... though I'm not sure the current approach needs to be removed, I don't think the implementation would need to be more complicated for supporting both

atticus-sullivan:
@L3MON4D3 Did you already start working on this?
Right now I've got exams, but after that (takes a while, ~end of the month) I'd look into this, if you're fine with that.
(I know lua well, but not the vim api, but I think this task should be doable after reading your code)

L3MON4D3:
Oh, no I have not :sweat_smile:
[...]

PR #513, opened 2022-08-08
Autotrigger as snippet option

atticus-sullivan:
Allows to specify `autotriggered` in the snippet options, and the snippet is added to the corresponding autosnippet tables. No need anymore to add snippets and autosnippets separately.

TODO:
- [ ] run/modify tests
- [ ] check if [autotable](https://github.com/atticus-sullivan/LuaSnip/tree/autotable_testing) functionality (c.f. [lua-users](https://lua-users.org/wiki/AutomagicTables)) is worth a try (see [this PR](https://github.com/atticus-sullivan/LuaSnip/pull/1) for work on this)
- [ ] add documentation of the new feature

### DIFF
--- a/lua/luasnip/nodes/snippet.lua
+++ b/lua/luasnip/nodes/snippet.lua
@@ -154,6 +154,8 @@ local function init_snippet_opts(opts)
 	-- return sn(t("")) for so-far-undefined keys.
 	in_node.stored = setmetatable(opts.stored or {}, stored_mt)
 
+	in_node.autotriggered = opts.autotriggered or false
+
 	-- wrap non-snippetNode in snippetNode.
 	for key, nodes in pairs(in_node.stored) do
 		in_node.stored[key] = wrap_nodes_in_snippetNode(nodes)
--- a/lua/luasnip/session/snippet_collection.lua
+++ b/lua/luasnip/session/snippet_collection.lua
@@ -202,43 +202,63 @@ local current_id = 0
 -- snippets like {ft1={<snippets>}, ft2={<snippets>}}, opts should be properly
 -- initialized with default values.
 function M.add_snippets(snippets, opts)
-	local prio_snip_table = by_prio[opts.type]
-
 	for ft, ft_snippets in pairs(snippets) do
-		local ft_table = by_ft[opts.type][ft]
+		local prios = {
+			autosnippets = {},
+			snippets = {}
+		}
+		local types = {
+			autosnippets = false,
+			snippets = false
+		}
 
-		if not ft_table then
-			ft_table = {}
-			by_ft[opts.type][ft] = ft_table
-		end
-
-		-- TODO: not the nicest loop, can it be improved? Do table-checks outside
-		-- it, preferably.
+		-- collect which tables should be added
+		-- and do some initialization
 		for _, snip in ipairs(ft_snippets) do
 			snip.priority = opts.override_priority
 				or (snip.priority ~= -1 and snip.priority)
 				or opts.default_priority
 				or 1000
 
-			if not prio_snip_table[snip.priority] then
-				prio_snip_table[snip.priority] = {}
-			end
-
-			local prio_ft_table
-			if not prio_snip_table[snip.priority][ft] then
-				prio_ft_table = {}
-				prio_snip_table[snip.priority][ft] = prio_ft_table
-			else
-				prio_ft_table = prio_snip_table[snip.priority][ft]
-			end
-
-			prio_ft_table[#prio_ft_table + 1] = snip
-
-			ft_table[#ft_table + 1] = snip
+			-- TODO
+			-- prefer more specific option I guess.
+			-- snip.autotriggered may not acutally default to nil, that will
+			-- cause problems with snippetProxy.
+			snip.autotriggered = snip.autotriggered ~= nil and snip.autotriggered or opts.type == "autosnippets"
 
 			snip.id = current_id
-			by_id[current_id] = snip
 			current_id = current_id + 1
+
+			types[snip.autotriggered and "autosnippets" or "snippets"] = true
+			prios[snip.autotriggered and "autosnippets" or "snippets"][snip.priority] = true
+		end
+
+		-- create necessary tables
+		for _, typename in ipairs({"autosnippets", "snippets"}) do
+			-- only create table if there are snippets for it.
+			if types[typename] then
+				if not by_ft[typename][ft] then
+					by_ft[typename][ft] = {}
+				end
+
+				local prio_snippet_table = by_prio[typename]
+				for prio, _ in pairs(prios[typename]) do
+					if not prio_snippet_table[prio] then
+						prio_snippet_table[prio] = {
+							[ft] = {}
+						}
+					elseif not prio_snippet_table[prio][ft] then
+						prio_snippet_table[prio][ft] = {}
+					end
+				end
+			end
+		end
+
+		-- do the actual insertion
+		for _, snip in ipairs(ft_snippets) do
+			table.insert(by_prio[snip.autotriggered and "autosnippets" or "snippets"][snip.priority][ft], snip)
+			table.insert(by_ft[snip.autotriggered and "autosnippets" or "snippets"][ft], snip)
+			by_id[snip.id] = snip
 		end
 	end
 


atticus-sullivan:
Should solve #482 eventually

atticus-sullivan:
@L3MON4D3 regarding the [autotables](https://github.com/atticus-sullivan/LuaSnip/pull/1) (not handled nicely in a separate repo, but I wanted to do this on a new branch and the PR to merge that branch with this one belongs to the other repo).

I've got the issue, that you preallocate the tables with the `1000` index (see e.g. [here](https://github.com/L3MON4D3/LuaSnip/blob/c599c560ed26f04f5bdb7e4498b632dc16fb9209/lua/luasnip/session/snippet_collection.lua#L23)).
Is this really necessary (I see no reason why)? (it prevents me from nicely injecting the automagick metatable)

L3MON4D3:
> I've got the issue, that you preallocate the tables with the 1000 index (see e.g. [here](https://github.com/L3MON4D3/LuaSnip/blob/c599c560ed26f04f5bdb7e4498b632dc16fb9209/lua/luasnip/session/snippet_collection.lua#L23)).
Is this really necessary (I see no reason why)? (it prevents me from nicely injecting the automagick metatable)

Oh, no that's not necessary, it's just that 1000 is the default-priority, so those would probably be created eventually.

atticus-sullivan:
Alright, we should move this sub topic to the other PR (sorry for the confusion with the different branches / PRs)

lua/luasnip/nodes/snippet.lua
```diff
@@ -154,6 +154,8 @@ local function init_snippet_opts(opts)
 	-- return sn(t("")) for so-far-undefined keys.
 	in_node.stored = setmetatable(opts.stored or {}, stored_mt)
 
+	in_node.autotriggered = opts.autotriggered or false
```
L3MON4D3, line 157: I think this option should be in the `context`-table (ie. the first one passed to `s`).
That is populated a few lines below, in `init_snippet_context` (again, docs missing :sweat_smile:)

atticus-sullivan: Ah already was at that point, that I didn't knew what to put in which options table. Yes we definitely can move this option to the `context`-table

atticus-sullivan: Solved in the most recent commit

atticus-sullivan:
We need to talk about how to represent the autotriggered key comes down to `nil/true/false` vs `-1,0,1`. Using the former one seems most logical, on the other hand checking `var == false` is always weird (to differ between `false` and `nil`). The latter one simplifies adding more values (to store how the key was set, even if I don't really can think of a use of this right now).
What is your opinion about this?

L3MON4D3:
I'd also prefer using nil+true/false, but we cannot have any nil-values inside the snippets, due to `snippetProxy`:
It's used to delay parsing of lsp-snippets until they are actually needed (eg. until just before expansion). This is done by parsing it on [`__index`](https://github.com/L3MON4D3/LuaSnip/blob/01d75efdd22e14ceaf57f3c5da568c8af74f8f14/lua/luasnip/nodes/snippetProxy.lua#L70-L84), so as soon as some value that is not available in the `snippetProxy` already is accessed, it's parsed.
We want to prevent parsing them on eg. `add_snippets`, so if `autotriggered` is `nil` at that point, it will trigger parsing of all of those snippets :/
(I hope that made sense :sweat_smile:)

Soooo, we'll have to replace at least `nil` with some other value, I think `-1, 0, 1` makes most sense since `0,1` is synonymous with `false,true`, and I can at least see how `-1` might mean undefined.

atticus-sullivan:
Yes makes kinda sense (but I'm not that into the lsp-snippet stuff.

lua/luasnip/session/snippet_collection.lua
```diff
@@ -202,43 +202,72 @@ local current_id = 0
...
 			snip.priority = opts.override_priority
 				or (snip.priority ~= -1 and snip.priority)
 				or opts.default_priority
 				or 1000
 
-			if not prio_snip_table[snip.priority] then
-				prio_snip_table[snip.priority] = {}
-			end
+			-- TODO
+			-- prefer more specific option I guess.
```
atticus-sullivan, line 224: Not quite sure what you mean by this comment. Does this refer to the name of the option?

L3MON4D3: Oh, I meant that, since autotriggered can be set (or unset!) on the snippet and `add_snippets` (`type = "autosnippets"/type="snippets"`), to potentially conflicting values, we should prefer using the snippet-value since it's more specific.

atticus-sullivan: Ah I understand, that was previously resolved by your (if then else) expression in the next line, and still should be solved by the similar expression (just ported it to `-1`/`0`/`1` values)

L3MON4D3: IMO it would be a tiny bit nicer if the snippet, at the end, had true/false-values again.
But that's really elaborate, I think the proper solution would be making `snippetProxy` not `initialize` on some values, and just return nil if they are nil.
That way we can use more sensible values in prio and autotriggered.

L3MON4D3: I implemented that suggestion in `sProxy_allow_nil_select`, I'll merge that into master later I think.

atticus-sullivan: > at the end, had true/false-values again.

Well, yes this would be nicer, but only if we can use booleans in the first place as well (i.e. never use -1/0/1), as having the same key populated with different data types (int/bool) over time is kinda confusing, I think.

Is this (always using bools like in the beginning of this PR) possible with the changes you've made? (I'm not into `snippetProxy`)

L3MON4D3: Bools + nil is possible, yes👍👍
I'll merge it into master in a bit.

L3MON4D3: (you could also cherry-pick it into this branch, but I think putting it into master + rebase this PR will end up cleaner)

atticus-sullivan: Even tough having the chance of playing with cherry-pick (never used it (never had to) in reality) is tempting, merging the master into this branch is definitely the cleaner way (rebasing/rewriting history isn't that nice since I already "published" the branch here).
So this is done. Thanks for implementing in `snippetProxy` :+1:

atticus-sullivan:
So I think implementing this is done (or anything else open?). Now I'm gonna have a look at the tests and the docs.

After that, I'd plan to keep this PR pending until #514 is merged (this way the adjustments to use auto creating tables with this feature happens in here, I consider this nicer than the other way round).

lua/luasnip/session/snippet_collection.lua
L3MON4D3: Feel free to rebase (in other instances, merge is fine), I can see your point from a technical POV, but imo it's not necessary to limit oneself in this way :sweat_smile:

lua/luasnip/nodes/snippet.lua
```diff
@@ -185,6 +185,11 @@ local function init_snippet_context(context)
 	-- might be nil, but whitelisted in snippetProxy.
 	context.priority = context.priority
 
+	-- keep value, set to nil to signal that this option is unset technically
+	-- this can be removed but having this here to state the options of the
+	-- context-table is nice
+	context.autotriggered = context.autotriggered or nil
```
L3MON4D3, line 191: This might backfire if `context.autotriggered` is set to `false` :D
You make a good point in the comment though, maybe just omit the `or nil` like above, and put a comment on which values are fine.

lua/luasnip/session/snippet_collection.lua
L3MON4D3: (missclick :sweat_smile:)

L3MON4D3:
> Now I'm gonna have a look at the tests and the docs.

The failing test seems to be due to `snippetProxy` getting instantiated again... mhmmmm I'll take a look, that shouldn't happen

L3MON4D3:
Ahhh, autotriggered just has to be whitelisted in `snippetProxy`, just like priority.
(look for `license_to_nil` :P)

L3MON4D3:
> After that, I'd plan to keep this PR pending until #514 is merged (this way the adjustments to use auto creating tables with this feature happens in here, I consider this nicer than the other way round).

Sounds good :+1:

lua/luasnip/nodes/snippet.lua
atticus-sullivan: Well yes this was a mistake. Should be fixed by now

atticus-sullivan:
Speaking of `license`. As this idea comes from [the lua-users wiki](https://lua-users.org/wiki/AutomagicTables), should we add some acknowledgement/appreciation for the authors (mentioned in the article) somewhere? (like I already said, I'm new to open source contributing and don't know exactly the code of conduct)

L3MON4D3:
xD
Yeah, that's true, I think it's okay to add a comment above the metatable-definition, "idea from ..."

atticus-sullivan:
Oups wrong PR, should have been in #514

L3MON4D3:
Yeah, let's go :D

Rebase would be great :+1:
Tests... yeah, just the usual, check that the new functionality actually works (either via integration (`add_snippet_spec.lua`) or unit-test)
Doc, you can append to the existing entry on `SNIPPET`, maybe mention it in `add_snippets` as well, for the key that sets all snippets to autosnippets (the behaviour for conflicting values would be important to mention)

atticus-sullivan:
Hm, I'm still not quite satisfied with the naming, `autotriggered` as option is kinda long and annoying to type. Nevertheless, it describes the option quite well. Do you have any other suggestions? (I just took the name to have something at first, but now I can't think of some other name which still is consistent with the other options)

atticus-sullivan:
Failing tests:
- `../../tests/integration/snippet_basics_spec.lua:353: Row 1 did not match.`: Jumping back into the snippet doesn't seem to be working.
```
Expected:
  |*a[a[]ab^]ab                                        |
Actual:
  |*a[a[]ab]a^b                                        |
```
- `../../tests/integration/dynamic_spec.lua:271: Row 1 did not match.`: To be honest, I've got no Idea on this one
```
 Expected:
  |*c^cc                                               |
Actual:
  |*c^c                                                |
```

L3MON4D3:
Maybe `autosnippet`?
Alternatively, we don't have to use a boolean, we could do something like `type="snippet"/"autosnippet"`, or `snippet_type="regular"/"auto"`, but tbh, bool seems nicer, and I don't think we'll need more types in the future (although this has already bitten us a bit with `regTrig`, which will become such a type-like field)

Maybe recommend using `extend_decorator` (see doc), setting `autotriggered` is a prime-target for that functionality.

L3MON4D3:
> Failing tests:
> 
> * `../../tests/integration/snippet_basics_spec.lua:353: Row 1 did not match.`: Jumping back into the snippet doesn't seem to be working.
> 
> ```
> Expected:
>   |*a[a[]ab^]ab                                        |
>   |{0:~                                                 }|
>   |{2:-- INSERT --}                                      |
> Actual:
>   |*a[a[]ab]a^b                                        |
>   |{0:~                                                 }|
>   |{2:-- INSERT --}                                      |
> ```
> 
> * `../../tests/integration/dynamic_spec.lua:271: Row 1 did not match.`: To be honest, I've got no Idea on this one
> 
> ```
>  Expected:
>   |*c^cc                                               |
>   |{0:~                                                 }|
>   |{2:-- INSERT --}                                      |
> Actual:
>   |*c^c                                                |
>   |{0:~                                                 }|
>   |{2:-- INSERT --}                                      |
> ```

Agh, those are a bit broken, I'm not sure why (at least the upper one I've seen fail/not fail on the same commit)

L3MON4D3:
Yeah, all's well again on a34cc52

Not sure what's causing them, but I suspect the testing-framework

L3MON4D3:
> Ah this would explain why it works for me locally

Yup, don't worry about those :D

atticus-sullivan:
> Maybe recommend using `extend_decorator` (see doc), setting `autotriggered` is a prime-target for that functionality.

Didn't knew about this, but sounds pretty nice (if not already used using `s`/`as` for snippet/auto snippet creation sounds reasonable for "standard setup").

L3MON4D3:
Man, these stupid decisions where the options are soooooo close together :|

`autosnippet` is to-the-point, but `type` is more "neutral" (I hope you know what I mean, if not I'll try my best to explain it xD), and I think I prefer it a tiny bit because of that.
`type` also fits a taaad better with existing api, as you said, but it's also a bit more ambiguous, so highlighting what `type` does in the doc would be important.

I don't get what you meant with "parse something like snippet+otherTypeOption", could you elaborate there?

atticus-sullivan:
Ah I think I understand, `autosnippet` is already kinda "biased" (ans reading quickly over a snippet where the word `autosnippet` occurs in the options (even if there's written `false` afterwards), one might think about the snippet the wrong way (maybe :shrug:))

> I don't get what you meant with "parse something like snippet+otherTypeOption", could you elaborate there?

If other "types of types" come up which claim the same option (like using the `type` option to configure `autosnippets` and `regTrig`s), one could of course merge these in on `type` option by using something like `snippet+regTrig` or `autosnippet+noRegTrig`, something like that.
So the argument was not to use the `type` option to prevent any clashes like this in the first place.

On the other hand you're right, `type` just fits in so nicely and the way this plugin is written its really normal `snippets` and `autosnippets` as two basic types. Every other type-like stuff will always be just a "sub-type" (hierarchically speaking below the (auto)snippet typing).

So I think we're closing in on the `type` option after all (you've got me convinced ^^).

L3MON4D3:
> Ah I think I understand, `autosnippet` is already kinda "biased" (ans reading quickly over a snippet where the word `autosnippet` occurs in the options (even if there's written `false` afterwards), one might think about the snippet the wrong way (maybe shrug))

Yeeah, exactly. Why call it `autosnippet`, and not `snippet`? That whole conundrum is avoided with `type`

> If other "types of types" come up which claim the same option (like using the `type` option to configure `autosnippets` and `regTrig`s), one could of course merge these in on `type` option by using something like `snippet+regTrig` or `autosnippet+noRegTrig`, something like that. So the argument was not to use the `type` option to prevent any clashes like this in the first place.
> On the other hand you're right, `type` just fits in so nicely and the way this plugin is written its really normal `snippets` and `autosnippets` as two basic types. Every other type-like stuff will always be just a "sub-type" (hierarchically speaking below the (auto)snippet typing).

Mhmmm okay, we'll have to use a different name for the trigger-type (probably `trigType`). Now, I'm not sure which is nicer (in the future), `snippetType` and `trigType` (autotrigger and trigger-type are orthogonal), or `type` and `trigType` (trigger-type is a subtype to autotrigger, as you suggested).

For this one, I really don't care, just wanted to mention the differences from my POV.

(collapsing the type-options into one string is interesting, it's close to what ultisnips is doing. Might be cool to implement an option that takes a string like that, for quick compatibility :D)

> So I think we're closing in on the `type` option after all (you've got me convinced ^^).

Ayyyy :D

atticus-sullivan:
> (collapsing the type-options into one string is interesting, it's close to what ultisnips is doing. Might be cool to implement an option that takes a string like that, for quick compatibility :D)

To be honest, not a big fan of this one, as the mapping which setting goes to which option has to be positional (at least if it's the way I think of it right now) so introducing default values or giving just one setting seems pretty ugly. But that should be part of another issue/PR, not this one.
Just mentioning it again, since if you already are thinking about this the `type` key is the only logical key for such a general option. Thus we really should "reserve" it for that purpose and go with `snippetType` (sorry for the many changes of mind on my 

So I'll quickly rename the keys in the code and in the docs. Not sure if you want to have a final look at the change before I squash the commits and merging afterwards.

L3MON4D3:
> To be honest, not a big fan of this one, as the mapping which setting goes to which option has to be positional (at least if it's the way I think of it right now) so introducing default values or giving just one setting seems pretty ugly. But that should be part of another issue/PR, not this one.

Yeah, not in this PR, I just have to immediately write down ideas I have :sweat_smile:
(just to quickly explain: we accept a string, if it contains eg. `A`, the snippet is a autosnippet, if not a regular one. `b` for expanding at beginning of line only, `r` for regex-triggers. Could be an alternative (no, extension! merging options could be fiddly, but doable) to the rather "long" options we have right now)

> Thus we really should "reserve" it for that purpose and go with snippetType (sorry for the many changes of mind)

definitely no worries at all, it doesn't hurt to consider multiple options (tbh, I also think `snippetType` is the safer option)

Just squash, I'll check the result :D

tests/integration/add_snippets_spec.lua
```diff
@@ -193,4 +193,73 @@ describe("add_snippets", function()
...
+				ls.snippet({trig="triB", autotriggered=false}, {ls.text_node("helloBworld")}, {})
+			}, {
+				key = "b"
+			} )
+		]])
+		exec_lua([[
+			ls.add_snippets("all", {
+				ls.snippet({trig="triC", autotriggered=nil}, {ls.text_node("helloCworld")}, {})
+			}, {
+				key = "c"
```
L3MON4D3, line 217: Tests look good :+1:
I think it would be a bit nicer to add `type="snippets"` explicitly in this call, and one more tests with `autotriggered`/(`snippetType`)`=nil` and add-snippets-`type` `"autosnippet"` (to make sure the `nil` is correctly handled for both `type`s)

L3MON4D3: Oh, right, and one test to ensure we keep behaviour for conflicting (snippet)types

atticus-sullivan:
Oh and just a sort note. I'll just use `snippetType` for indexing the `by_...` tables (if the strings are already presents, why not). I just have to switch from the singular form (much more comprehensive in configuring `s({snippetType="autosnippets"})` is kinda weird) to the plural form (otherwise direct indexing doesn't work). (of course using `assert`s to ensure only valid values are present).
And due to string interning (at least I hope luajit does this (PUC lua does it to my knowledge)) this shouldn't pose a problem

tests/integration/add_snippets_spec.lua
atticus-sullivan: Ah alright :+1:

atticus-sullivan:
Should be finally finished. If you still want to have a look at the single commits, you can look at [this branch](https://github.com/atticus-sullivan/LuaSnip/tree/dev_autotrigger_as_snippet_opt) (I like to at least keep the atomic commits somewhere around for a while).

---

So this is the end of the development of this feature (all in all it took quite some time in the end, even when looking at the code its pretty few changes). It was fun working with you and discussing things (maybe at some points to much in detail ^^) (I'm not really used to projects which I don't do completely on my own).
And thanks for your input on the design decisions :+1:

L3MON4D3:
> Oh and just a sort note. I'll just use `snippetType` for indexing the `by_...` tables (if the strings are already presents, why not). I just have to switch from the singular form (much more comprehensive in configuring `s({snippetType="autosnippets"})` is kinda weird) to the plural form (otherwise direct indexing doesn't work). (of course using `assert`s to ensure only valid values are present). And due to string interning (at least I hope luajit does this (PUC lua does it to my knowledge)) this shouldn't pose a problem

Oh yeah, that's nice :+1:

I like the additional asserts, we should be much more generous with stuff like that :D

> Should be finally finished (2nd force push was the rebase on master). 

Yessss looks good :tada::tada:
It really took quite some time :sweat_smile: But I think this is a really nice addition, so it was worth it (& I'm very happy with the interface we now have to it, Plus: autotables make adding snippets much easier) (Also, I guess it took a bit longer than usual just because of the autotables, which are a finnicky and require a lot of attention to detail)

You're right, we did discuss some things a bit too much, but I really think that leads to better results (even though it takes a bit longer to get there)

> And thanks for your input on the design decisions :+1:

Always :P

Thanks a lot for your work, and your patience especially :D

PR #514, opened 2022-08-08
Make use of auto creating tables

atticus-sullivan:
Makes inserting into the snippet tables easier as there is no need to manually create tables along the way.

TODOs:
- [x] Benchmark if this is worth it and check if creating new functions for each created table is nice or if depth should be checked via a table member
- [x] check at which places now `rawget()` has to be used (it this nice code style?) to check if a key exists
- [x] tests? see [this comment](https://github.com/L3MON4D3/LuaSnip/pull/514#issuecomment-1209298574)

#513 works in the same code area, I just wanted to have this ready for testing. If we merge this PR, then make adjustments to #513 and merge that PR after that, or if we simply close this PR, we will see (mainly wanted to have a place to discuss which is bundled with the rest of the project, not in some other repo)

_Help wanted:_ Ideas how to benchmark all this? (will most probably be done with [this approach](https://github.com/L3MON4D3/LuaSnip/pull/315#issuecomment-1047661997) but we need test scenarios to benchmark, I think)

### DIFF
--- a/lua/luasnip/session/snippet_collection.lua
+++ b/lua/luasnip/session/snippet_collection.lua
@@ -85,10 +85,10 @@ local by_prio_snippets_mt = {
 -- metatable for the by_prio table used when by_prio.type[prio] is reset
 -- create here so that it can be shared and only has to be created once
 local prio_mt2 = {
-	__index = function(s,k)
+	__index = function(s, k)
 		-- make new tables as they are indexed
 		return auto_creating_tables(s, k, 2)
-	end
+	end,
 }
 
 setmetatable(by_prio.snippets, by_prio_snippets_mt)
--- a/lua/luasnip/session/snippet_collection.lua
+++ b/lua/luasnip/session/snippet_collection.lua
@@ -10,6 +10,28 @@ local M = {
 	invalidated_count = 0,
 }
 
+-- depth specifies how many levels under this table should be allowed to index
+-- throug this metamethod
+-- set depth to 0 to disable checking
+-- Acknowledgment: This is (maybe more than) inspired by
+-- https://lua-users.org/wiki/AutomagicTables so special thanks to
+-- Thomas Wrensch and Rici Lake for sharing their ideas on this topic.
+local function auto_creating_tables(self, key, depth)
+	local t = {}
+	assert(depth ~= 1, "don't index at that level")
+	setmetatable(t, {
+		-- creating a new function on each time (could be shared) isn't that
+		-- nice. Nonetheless this shouldn't be too bad, as these are only
+		-- created twice (auto+snippet) per ft and twice for each prio,ft
+		-- combination
+		__index = function(s, k)
+			return auto_creating_tables(s, k, depth - 1)
+		end,
+	})
+	self[key] = t
+	return t
+end
+
 local by_key = {}
 
 -- stores snippets/autosnippets by priority.
@@ -17,20 +39,10 @@ local by_prio = {
 	snippets = {
 		-- stores sorted keys, eg 1=1000, 2=1010, 3=1020,..., used for
 		-- quick iterating.
-		order = {
-			1000,
-		},
-		[1000] = {
-			all = {},
-		},
+		order = {},
 	},
 	autosnippets = {
-		order = {
-			1000,
-		},
-		[1000] = {
-			all = {},
-		},
+		order = {},
 	},
 }
 
@@ -59,16 +71,28 @@ local function insert_sorted_unique(t, k)
 	t[i] = k
 end
 
-local sort_mt = {
+local by_prio_snippets_mt = {
+	__index = function(s, k)
+		-- make new tables as they are indexed
+		return auto_creating_tables(s, k, 3)
+	end,
 	__newindex = function(t, k, v)
 		-- update priority-order as well.
 		insert_sorted_unique(t.order, k)
 		rawset(t, k, v)
 	end,
 }
+-- metatable for the by_prio table used when by_prio.type[prio] is reset
+-- create here so that it can be shared and only has to be created once
+local prio_mt2 = {
+	__index = function(s,k)
+		-- make new tables as they are indexed
+		return auto_creating_tables(s, k, 2)
+	end
+}
 
-setmetatable(by_prio.snippets, sort_mt)
-setmetatable(by_prio.autosnippets, sort_mt)
+setmetatable(by_prio.snippets, by_prio_snippets_mt)
+setmetatable(by_prio.autosnippets, by_prio_snippets_mt)
 
 -- iterate priorities, high to low.
 local function prio_iter(type)
@@ -88,6 +112,14 @@ local by_ft = {
 	snippets = {},
 	autosnippets = {},
 }
+
+local by_ft_snippets_mt = {
+	__index = function(s, k)
+		return auto_creating_tables(s, k, 2)
+	end,
+}
+setmetatable(by_ft.snippets, by_ft_snippets_mt)
+setmetatable(by_ft.autosnippets, by_ft_snippets_mt)
 
 local by_id = setmetatable({}, {
 	-- make by_id-table weak (v).
@@ -118,14 +150,18 @@ function M.clear_snippets(ft)
 		-- remove all (auto)snippets for all priorities.
 		for _, prio in ipairs(by_prio.snippets.order) do
 			by_prio.snippets[prio] = {}
+			setmetatable(by_prio.snippets[prio], prio_mt2)
 		end
 		for _, prio in ipairs(by_prio.autosnippets.order) do
 			by_prio.autosnippets[prio] = {}
+			setmetatable(by_prio.autosnippets[prio], prio_mt2)
 		end
 
-		by_ft.snippets = {}
-		by_ft.autosnippets = {}
 		by_key = {}
+		by_ft.snippets = {}
+		setmetatable(by_ft.snippets, by_ft_snippets_mt)
+		by_ft.autosnippets = {}
+		setmetatable(by_ft.autosnippets, by_ft_snippets_mt)
 	end
 end
 
@@ -174,6 +210,7 @@ function M.clean_invalidated(opts)
 		for key, prio_snippets in pairs(type_snippets) do
 			if key ~= "order" then
 				type_snippets[key] = without_invalidated(prio_snippets)
+				setmetatable(type_snippets[key], prio_mt2)
 			end
 		end
 	end
@@ -202,8 +239,6 @@ local current_id = 0
 -- snippets like {ft1={<snippets>}, ft2={<snippets>}}, opts should be properly
 -- initialized with default values.
 function M.add_snippets(snippets, opts)
-	local prio_snip_table = by_prio[opts.type]
-
 	for ft, ft_snippets in pairs(snippets) do
 		local ft_table = by_ft[opts.type][ft]
 
@@ -212,33 +247,19 @@ function M.add_snippets(snippets, opts)
 			by_ft[opts.type][ft] = ft_table
 		end
 
-		-- TODO: not the nicest loop, can it be improved? Do table-checks outside
-		-- it, preferably.
 		for _, snip in ipairs(ft_snippets) do
 			snip.priority = opts.override_priority
 				or (snip.priority and snip.priority)
 				or opts.default_priority
 				or 1000
 
-			if not prio_snip_table[snip.priority] then
-				prio_snip_table[snip.priority] = {}
-			end
-
-			local prio_ft_table
-			if not prio_snip_table[snip.priority][ft] then
-				prio_ft_table = {}
-				prio_snip_table[snip.priority][ft] = prio_ft_table
-			else
-				prio_ft_table = prio_snip_table[snip.priority][ft]
-			end
-
-			prio_ft_table[#prio_ft_table + 1] = snip
-
-			ft_table[#ft_table + 1] = snip
-
 			snip.id = current_id
-			by_id[current_id] = snip
 			current_id = current_id + 1
+
+			-- do the insertion
+			table.insert(by_prio[opts.type][snip.priority][ft], snip)
+			table.insert(by_ft[opts.type][ft], snip)
+			by_id[snip.id] = snip
 		end
 	end
 
@@ -250,12 +271,29 @@ function M.add_snippets(snippets, opts)
 	end
 end
 
+-- specialized copy functions to not loose performance on ifs when copying
+-- and to be able to specify when pairs or ipairs is used
+local function copy_by_ft_type_ft(tab)
+	local r = {}
+	for k, v in ipairs(tab) do
+		r[k] = v
+	end
+	return r
+end
+local function copy_by_ft_type(tab)
+	local r = {}
+	for k, v in pairs(tab) do
+		r[k] = copy_by_ft_type_ft(v)
+	end
+	return r
+end
+
 -- ft may be nil, type not.
 function M.get_snippets(ft, type)
 	if ft then
-		return by_ft[type][ft]
+		return copy_by_ft_type_ft(by_ft[type][ft])
 	else
-		return by_ft[type]
+		return copy_by_ft_type(by_ft[type])
 	end
 end
 
--- a/tests/unit/snippet_collection_spec.lua
+++ b/tests/unit/snippet_collection_spec.lua
@@ -1,0 +1,48 @@
+local helpers = require("test.functional.helpers")(after_each)
+
+local works = function(snippets, opts) end
+
+describe("snippet_collection.add/get", function()
+	-- apparently clear() needs to run before anything else...
+	helpers.clear()
+	helpers.exec("set rtp+=" .. os.getenv("LUASNIP_SOURCE"))
+
+	it("get_id", function()
+		local function foo()
+			return helpers.exec_lua([[
+				local s,t = require("luasnip").snippet, require("luasnip").text_node
+				local collection = require("luasnip.session.snippet_collection")
+				collection.clear_snippets()
+				local s1,s2 = s("trig1", t("snippet1")), s("trig2", t("snippet2"))
+				local opts = {type="snippets", default_priority=1000}
+				collection.add_snippets({["txt"]={s1,s2}}, opts)
+				return collection.get_id_snippet(s1.id) == s1
+				]])
+		end
+		assert.has_no.errors(foo)
+		assert.is_true(foo())
+	end)
+
+	it("get_snippets", function()
+		local function foo()
+			return helpers.exec_lua([[
+				local s,t = require("luasnip").snippet, require("luasnip").text_node
+				local collection = require("luasnip.session.snippet_collection")
+				collection.clear_snippets()
+				local s1,s2 = s("trig1", t("snippet1")), s("trig2", t("snippet2"))
+				local opts = {type="snippets", default_priority=1000}
+				collection.add_snippets({["txt"]={s1,s2}}, opts)
+				local r = collection.get_snippets(nil, "snippets")
+				assert(getmetatable(r) == nil)
+				for _,t in pairs(r) do
+					assert(getmetatable(t) == nil)
+				end
+				r = collection.get_snippets("txt", "snippets")
+				assert(getmetatable(r) == nil)
+				return #r == 2 and ((r[1] == s1 and r[2] == s2) or (r[1] == s2 and r[2] == s1))
+				]])
+		end
+		assert.has_no.errors(foo)
+		assert.is_true(foo())
+	end)
+end)


atticus-sullivan:
>> I've got the issue, that you preallocate the tables with the 1000 index (see e.g. [here](https://github.com/L3MON4D3/LuaSnip/blob/c599c560ed26f04f5bdb7e4498b632dc16fb9209/lua/luasnip/session/snippet_collection.lua#L23)).
Is this really necessary (I see no reason why)? (it prevents me from nicely injecting the automagick metatable)

>Oh, no that's not necessary, it's just that 1000 is the default-priority, so those would probably be created eventually.

Ah, ok. So I think we either do it the way I'm doing it right now (originally thought as a quick and dirty fix), or just remove this preallocation and rely on it being created when the snippets get inserted :shrug:

L3MON4D3:
TY for moving that discussion here :+1:

> or just remove this preallocation and rely on it being created when the snippets get inserted

Yeah, that's cleaner, go ahead :)

Benchmarking: I think the quickest way would be just `load()`ing (not lazy_load, there the snippets are only added once they're really required) all of `vim-snipmate`, and comparing before-after flamegraphs (at least that's what I've done so far, seems to work pretty well :D ).
I don't think we need to very much care about priority-snippets, that should only be a very small amount.
A way to test them would be just adding snippets with different priorities in a loop.

I think checking the depth is not necessary, we only ever `ipairs` or set a new key in the "depth=1"-tables.
Also, the tables are only accessed directly in `snippet_collection`, and I can't see any place where we'd actually need the `rawget`, so we're in luck there :)

L3MON4D3:
Oh, one problem maybe: in `M.get_snippets` we need to return a list of snippets for some filetype.
I'm not sure if we want to return tables with (maybe) unexpected metatables, so we should probably do a shallow copy there.
Alternatively, we could specify that those tables may only be iterated via `ipairs`, but that's a bit unconventional :/

atticus-sullivan:
Small note on `rawset` and `tab[i]` regarding performance: (but doesn't concern us)
Just read on a lua discord (tbh, didn't check it on my own)
> PUC Lua: rawset is ~3x slower than just setting the key,
Luajit: about the same (its implemented as a fast function)

But this shouldn't be a problem as nvim uses luajit.

L3MON4D3:
> Small note on `rawset` and `tab[i]` regarding performance: (but doesn't concern us) Just read on a lua discord (tbh, didn't check it on my own)
> 
> > PUC Lua: rawset is ~3x slower than just setting the key,
> > Luajit: about the same (its implemented as a fast function)
> 
> But this shouldn't be a problem as nvim uses luajit.

Ouh, good to know, I would've expected it to be faster, if anything..
Also, just so you're aware, and I've just recently learned this myself, nvim might fall back to regular lua if luajit is not available, so there are some platforms where PUC lua is used.
But I'll assume that's a diminishingly small number of users, so your point about not worrying about it stands IMO

leiserfg:
They are not the same, in metatables it's safer to use `rawset` as it bypasses the metatable.

L3MON4D3:
Yeah, we'll have to careful around those tables from now on, eg. should have some tests that cover a good portion of that code. I think we don't have any tests for `get_snippets`.
(Just throwing it out there, let's wait with adding those until the groundwork is actually done :D)

lua/luasnip/session/snippet_collection.lua
```diff
@@ -250,12 +256,24 @@ function M.add_snippets(snippets, opts)
 	end
 end
 
+-- TODO benchmark without doing a shallow copy (is it worth to be
+-- unconventional and return tables with weird metatables to the outside?)
+-- TODO shallow copy functionality in nvim (I found nothing)?
+-- if not this should at least be moved to some utils file
```
L3MON4D3, line 262: I agree, `luasnip.util.util`'s the place :D
Otherwise there's `vim.list_slice`, but that name isn't very nice.

Maybe specify this function is for lists, that way `ipairs` can be used instead of `pairs`.

atticus-sullivan: `ipairs` is only possible in the former case (`by_ft[type][ft]`), in the latter one (`by_ft[type]`) the returned table isn't a list but a dictionary.
As we already differ the two cases, we can of course use different copy functions.

L3MON4D3: True!
Ah, also: if the whole table should be returned, a second copy of the snippets for each filetype becomes necessary, if we really don't want to leak the metatables.

atticus-sullivan: Oh that's right as well :see_no_evil:
In that case we need different copy functions either way, I think.

atticus-sullivan:
Alright, implementing should be done by now so far. Next step should be the benchmarking. L3MON4D3 once suggested doing so as described [here](https://github.com/L3MON4D3/LuaSnip/pull/315#issuecomment-1047661997).

Tried this, but the `jit.p` module couldn't be found. Anything to install before being able to used this?
Note: I'm using
```
$ nvim --version
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"
```
(so the requirement luajit 2.1 stated in the [comment](https://github.com/L3MON4D3/LuaSnip/pull/315#issuecomment-1047661997)) should be fulfilled)
Sorry I have to ask, but I also cannot find a real documentation on the `jit.p` module (would be interesting tough to understand the passed arguments (only found some source code on this, but no documentation/reference)). @leiserfg, maybe you can help with this.

atticus-sullivan:
And on what to benchmark, as far as I understood L3MON4D3, loading (no lazy-load) the snippets from `tests/data/snipmate-snippets/` served well so far, right?

L3MON4D3:
> Sorry I have to ask, but I also cannot find a real documentation on the jit.p module (would be interesting tough to understand the passed arguments (only found some source code on this, but no documentation/reference)).

There's some doc [here](https://blast.hk/moonloader/luajit/ext_profiler.html), but I'm stumped as to why it doesn't work on your machine o.O
Maybe LUA_(C)PATH again?

> And on what to benchmark, as far as I understood L3MON4D3, loading (no lazy-load) the snippets from tests/data/snipmate-snippets/ served well so far, right?

Ah, no, I meant the snippets provided by [`vim-snippets`](https://github.com/honza/vim-snippets), that should be a good number of `add_snippets`-calls (at least one per file, with quite some snippets each).

leiserfg:
If your luajit for a packaging reason does not comes with `jit.p`, you can just copy it from [here](https://github.com/luapower/luajit/blob/master/jit/p.lua), it's just a set of wrappers to `jit.profile`.

atticus-sullivan:
> Maybe LUA_(C)PATH again?
Yes this was the problem (again).
Thanks for the reminder (this gets a bit annoying, I should think about some nice way to do this automatically).

atticus-sullivan:
> Ah, no, I meant the snippets provided by [vim-snippets](https://github.com/honza/vim-snippets), that should be a good number of add_snippets-calls (at least one per file, with quite some snippets each).

Ah nice yes this works. I get these two graphs for the current `master` branch and this one (also attached the data files).
To be honest I never worked with flamegraphs before and while just reading one graph, I've troubles comparing them properly.

[master.log](https://github.com/L3MON4D3/LuaSnip/files/9349486/master.log)
master
[selfPop.log](https://github.com/L3MON4D3/LuaSnip/files/9349487/selfPop.log)
selfPop

Generated via
```lua
require("jit.p").start("10,i1,s,m0,G", "/tmp/output.log")
require("luasnip.loaders.from_snipmate").load{
    paths = "~/coding/vim-snippets/snippets"
}
require'jit.p'.stop()
```

atticus-sullivan:
I just had a look at [this talk](https://www.youtube.com/watch?v=D53T1Ejig1Q) (of the author of the flamegraph tooling) and it's amazing, you can view the svg in the browser and you'll be able to hover over the bars (click and zoom) and get more information.
Turns out that the flamegraphs I generated are hard to compare, since there are many one sample bars (some functions may be fast enough so that they only appear by chance, generating "noise").
Best would be if we could get a higher resolution, generating more samples. As I'm afraid this is not possible with `jit.p`, I guess we have to cope with having 1 sample bars which may or may not appear (so these shouldn't be weighted so strong).

leiserfg:
> Best would be if we could get a higher resolution, generating more samples. As I'm afraid this is not possible with `jit.p`, I guess we have to cope with having 1 sample bars which may or may not appear (so these shouldn't be weighted so strong).

You can, just change the 10 to another number.

atticus-sullivan:
Ah nice, thanks (based on the docs I thought the `10` would be the stack dump depth, not the sampling rate, which already should have been on `1ms` due to the `i1` argument)

leiserfg:
> Ah nice, thanks (based on the docs I thought the `10` would be the stack dump depth, not the sampling rate, which already should have been on `1ms` due to the `i1` argument)

you are right, I was abusing my memory :facepalm:

L3MON4D3:
Well, I think in this case it's alright to just look at the total time elapsed, it's just nice to have the flamegraphs ready, then one can directly see what takes too long.
In the two graphs you posted, IIUC, master takes 87ms, this branch 107ms, but `add_snippets` does not contribute to this total at all.
Is that difference consistent? (Probably not, you later said that this branch performed better, so it's probably fluctuating)

Maybe it would be best to write a benchmark that really just calls `add_snippets` in a loop, with large numbers of snippets. Then there would be no parse or deepcopy polluting the flamegraph.

> loss of performance we certainly get due to copying tables instead of just returning them in get_snippets

yup, that's true... We'd need another benchmark for this.

atticus-sullivan:
> Probably not, you later said that this branch performed better, so it's probably fluctuating

Ah yes, that statement was based on another graph then the one I posted. I've got 91 samples for selfPop and 96 for master.
I think I'll play with the snipmate loader to return a list of the nodes to be added instead of directly adding them, so I can add them manually afterwards. (Changes just for benchmarking)

L3MON4D3:
> I think I'll play with the snipmate loader to return a list of the nodes to be added instead of directly adding them, so I can add them manually afterwards. (Changes just for benchmarking)

Oh yeah, good idea, that's more realistic than my suggestion :+1:

L3MON4D3:
(Not sure how you're approaching this, easiest might be appending them to some global table instead of `add_snippets`)

atticus-sullivan:
>> This idea comes from [the lua-users wiki](https://lua-users.org/wiki/AutomagicTables), should we add some acknowledgement/appreciation for the authors (mentioned in the article) somewhere? (like I already said, I'm new to open source contributing and don't know exactly the code of conduct)
>
> xD
Yeah, that's true, I think it's okay to add a comment above the metatable-definition, "idea from ..."

(posted in the wrong PR, sorry)

L3MON4D3:
Ahh, `ss` is the arg-tuple to `add_snippets`? Cool :+1:

> (should not be catched by caching, as it should invalidate the snippets on each iteration and insert them once again, right?)

Might be that parsing is skipped, but that doesn't affect us here, I think.

Are you sure it's just 346 snippets? I could've sworn C alone has like 100 or so..

Okay it for sure looks like the performance-difference is negligible, and dwarfed by stuff like parsing, so that's pretty good :+1:
Could you do one more test for `get_snippets`?

atticus-sullivan:
> Might be that parsing is skipped, but that doesn't affect us here, I think.

parsing (as it takes place in the load function) is outside the loop, I think.

> Are you sure it's just 346 snippets? I could've sworn C alone has like 100 or so..

At least that's what `print(#ss)` returned (I didn't doublecheck it).

>  Could you do one more test for `get_snippets`?

Yes I'll do (later today/tomorrow I think). Any remarks on the benchmark? (a simple repeated `get_snippets` (is pretty far from reality and) will of course show a massive slowdown as the difference between `return copy` vs `return` is pretty obvious)

L3MON4D3:
Ahh, 346 is just the number of `add_snippets`-calls, not the number of snippets :facepalm:
Okay that makes more sense :+1:

`get_snippets` is pretty much always used from `cmp_luasnip`. It would be interesting to see the difference of getting just all snippets of a big collection, if both are not measurable, we good :D
I don't think the slowdown will be massive tbh, even if it is, we could just cache the results and invalidate on `add_snippets`.

atticus-sullivan:
Seams that you're right. Well actually indeed we only do a shallow copy so copying takes some time (especially on if many many snippets are registered for one filetype) but its only "pointers" we have to copy.

Results of
```lua
	local foo = function(...) end
	require("jit.p").start("1,i1,s,m0,G", "/tmp/output.log")
	print(os.time())
	-- for _=1,100000000 do
	for _=1,1 do
		local i = ls.get_id_snippet(10)
		local fs = ls.get_snippets("c")
		for _,f in ipairs(fs) do
			i = ls.get_id_snippet(f.id)
		end
		print(#ls.get_id_snippet(fs[#fs].id))
		print(#i, #fs)
		foo(i, #i, fs, #fs)
	end
	print(os.time())
	print("hi")
	require'jit.p'.stop()
```
(Remark:  `get_id_snippet` didn't always returned an empty table (but snippet expansion worked), I wonder is this is intended event if the ids gathered from the `fs` snippets are taken)

[master.log](https://github.com/L3MON4D3/LuaSnip/files/9366259/master.log)
master
[selfPop.log](https://github.com/L3MON4D3/LuaSnip/files/9366261/selfPop.log)
selfPop

atticus-sullivan:
If you've got any ideas about how to benchmark otherwise (or know why getting by id doesn't work), just let me know

L3MON4D3:
I don't understand what the benchmark is supposed to do :sweat_smile:
Wouldn't a simple `get_snippets("<ft>")` do it? (maybe repeated to actually see the differences)

> (Remark: get_id_snippet didn't always returned an empty table (but snippet expansion worked), I wonder is this is intended event if the ids gathered from the fs snippets are taken)

`get_id_snippet(id)` should return nil (invalid id) or a snippet (valid id), not sure where empty tables are coming from :grimacing:

atticus-sullivan:
> I don't understand what the benchmark is supposed to do sweat_smile
Wouldn't a simple `get_snippets("<ft>")` do it? (maybe repeated to actually see the differences)

Oh that's approximately what I'm doing (just added some stuff to try to find a snippet `by_id` as well).

> not sure where empty tables are coming from

Hm, that's my bad, I keep forgetting that `#table` only works for tables with array part to check if the table is empty :see_no_evil:

atticus-sullivan:
Just ran a new test with
```lua
	local foo = function(...) end
	require("jit.p").start("1,i1,s,m0,G", "/tmp/output.log")
	print(os.time())
	for _=1,10000000 do
		local i = ls.get_id_snippet(10)
		local fs = ls.get_snippets("c")
		foo(i, #i, fs, #fs)
	end
	print(os.time())
	print("hi")
	require'jit.p'.stop()
```
and hm this runs really fast on the master (even no samples in the log) and takes ~ 2.000 samples (i.e. ~ 20 seconds) on this branch. With these results, I think the previous test was messed up (not sure anymore if I had the print statements included, which would make the benchmark false. I have no other explanation how the last test could possibly take so long on the master if now it's so fast).

So I'm afraid we have quite a performance problem (~20 seconds vs ~0 seconds) here.

> I don't think the slowdown will be massive tbh, even if it is, we could just cache the results and invalidate on add_snippets.

I don't have a real relation on how big the tables usually are, but sure we should cache them all? (thereby create a complete copy of all snippets EDIT1: only creates a copy of the lookup table -> only a copy of "pointers" )
EDIT2: `by_ft` should usually be always accessed with the same key -> only a copy of one table (+ `by_ft.all`)

Opinions? (in my opinion with EDIT2, it should be ok to do this caching)

L3MON4D3:
Not sure if it's such a bad performance-problem, I mean 20,000ms/10,000,000calls < 1ms/call for each `get_snippets` seems very okay to me (could you check if there is a difference for just one single call to `get_snippets`?)

If we want a cache:
In the current usage (by `cmp_luasnip`), it's pretty likely that each added snippet will be queried at some point in time, so it should make most sense to cache the entire `by_ft`-structure without metatables.
I wouldn't update the cache on `add_snippets` or `clean_invalidated` (not sure if that is its name), since many of `add_snippets` will be called in short succession on startup, probably without a `get_snippets` inbetween, so only the last would actually be used.
Sooo, If we want a cache, best set/update some identifier for the current snippet-collection in `add_snippets`, and update the cache if there is none/it's identifier no longer matches that of the current collection

atticus-sullivan:
Hm ok, relatively it is a huge performance difference between the two branches, but on a pretty low level, that's right.

> could you check if there is a difference for just one single call to `get_snippets`

Not sure how to get the time with more precision than with `os.time` (or if it is possible). And since a single call gives me no samples (too fast) for the flamegraph, this isn't an option as well. If you have ideas, let me know.

L3MON4D3:
> since a single call gives me no samples (too fast)

I think that's all the relevant info right there :D
I'm guessing the sample rate is <10ms, so if the new version is below that, that's very good.

If we get performance-issues later on we can still slap the cache on it, but I don't think that will be necessary at all

atticus-sullivan:
Alright, so I think we should leave a comment there that this could be subject for caching.
After that the benchmarking part is done, I think.
Remaining: Writing some tests

lua/luasnip/session/snippet_collection.lua
```diff
@@ -10,6 +10,22 @@ local M = {
...
+	local t = {}
+	if depth ~= 1 then
+		setmetatable(t, {
+			-- TODO not sure if this is that nice, creating a new function on
+			-- each time (lua-users does this by a seperate mamber)
+			__index = function(s, k)
+				return auto_creating_tables(s, k, depth - 1)
+			end,
+		})
+	end
```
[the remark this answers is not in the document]
atticus-sullivan: Well yes we could of course drop this restriction (as we kinda remove the metatable when we pass the tables to the outside this should be no problem). We just need to be careful in the `snippet_collection.lua` when working with these tables (e.g. `by_ft.snippets["xyz"][number]` as well as `by_prio.snippets[number]` might return unexpected empty tables (instead of `nil` or a valid snippet)).
Just noticed in fact, the depth restriction isn't in place for `by_ft` (using `0` as initial depth).

But I don't understand what you want to assert?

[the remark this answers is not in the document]
L3MON4D3: > We just need to be careful in the snippet_collection.lua when working with these tables

+1, yes exactly
My idea was to `assert` this, eg. check that `depth >= 1` whenever a new `auto_creating_table` is created.
(Obviously the result is very similar to the current code, and there's no performance-benefit since we can't produce a "release"-version of the repo without asserts (though it would be interesting to look into accomplishing just that...))

L3MON4D3:
> Writing some tests

Yup, `get_snippets`, and that we don't create unwanted metatables (this would already be included in the current tests if we `assert` it, otherwise I'm unsure how to test it :grimacing:)

lua/luasnip/session/snippet_collection.lua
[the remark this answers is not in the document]
atticus-sullivan: `depth >= 1` changed to `depth ~= 1` to be able to disable checking/asserting by setting depth to `0`.

atticus-sullivan:
> (Obviously the result is very similar to the current code, and there's no performance-benefit since we can't produce a "release"-version of the repo without asserts (though it would be interesting to look into accomplishing just that...))

I work a bit on [this repo](https://github.com/minoki/cluttex/blob/master/build.lua) (on rare occasions ^^). There they created a `build.lua` script to produce a release-version (in their case to get a single file containing all modules and differentiate between linux and windows). So that's something some people actually do. 
Not quite sure though how to exactly do it for this project (but probably the best way would be to introduce some markers similar to the `%-%- TEST CODE\n(.-)%-%- END TEST CODE\n` in `cluttex`).

Take this just as a small brain teaser, if we're gonna discuss this further, we most probable should do it in a new issue (not clutter this PR any further ^^)

atticus-sullivan:
> Yup, `get_snippets`, and that we don't create unwanted metatables (this would already be included in the current tests if we `assert` it, otherwise I'm unsure how to test it :grimacing:)

Well there is `getmetatable` (returns `nil` if no metatable present according to the [reference](https://www.lua.org/manual/5.1/manual.html#pdf-getmetatable)).

---
*Note:* We currently create possibly unwanted metatables (for depths as well where `__index` always will fail due to `assert`), as I changed the `if` to an `assert`. I'm fine with this, what do you think? (more concrete example: The table `by_ft.snippets` has a metatable with "working" `__index`. The table `by_ft.snippets["lua"]` has a metatable with "asserting" `__index` as it is only supposed to act as array containing snippets but the metatable is still present)

atticus-sullivan:
Regarding tests, the files in `tests/*/*.lua` are executed in a `busted` environment (see [this repo](https://github.com/lunarmodules/busted))? And these tests would be unit tests right?

lua/luasnip/session/snippet_collection.lua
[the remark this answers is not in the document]
L3MON4D3: Sounds good, that might be useful in other places :+1:

L3MON4D3:
> > Yup, `get_snippets`, and that we don't create unwanted metatables (this would already be included in the current tests if we `assert` it, otherwise I'm unsure how to test it :grimacing:)
> 
> Well there is `getmetatable` (returns `nil` if no metatable present according to the [reference](https://www.lua.org/manual/5.1/manual.html#pdf-getmetatable)).

Right, what I meant there was that, since we verify/test that some `__index` are not called during regular usage via the `assert`, and the tests simulate regular usage (except `get_snippets`), we automatically test that no unwanted metatables are created.

> _Note:_ We currently create possibly unwanted metatables (for depths as well where `__index` always will fail due to `assert`), as I changed the `if` to an `assert`. I'm fine with this, what do you think? (more concrete example: The table `by_ft.snippets` has a metatable with "working" `__index`. The table `by_ft.snippets["lua"]` has a metatable with "asserting" `__index` as it is only supposed to act as array containing snippets but the metatable is still present)

That sounds perfect, I think we can't avoid the metatable if we want to assert that some tables aren't `__index`d



> Regarding tests, the files in `tests/*/*.lua` are executed in a `busted` environment (see [this repo](https://github.com/lunarmodules/busted))? And these tests would be unit tests right?

I think so, pretty much. We use neovims internal testing-engine, so the test-files have to be in the `tests`-directory and match `*_spec.lua`.
For `get_snippets`, unit seems appropriate :+1:

lua/luasnip/session/snippet_collection.lua
```diff
@@ -20,17 +39,11 @@ local by_prio = {
 		order = {
 			1000,
```
L3MON4D3, line 40: Just noticed: these can also be removed, now that the tables with prio 1000 are no longer present from the beginning

atticus-sullivan:
> > > Yup, `get_snippets`, and that we don't create unwanted metatables (this would already be included in the current tests if we `assert` it, otherwise I'm unsure how to test it grimacing)
> > 
> > 
> > Well there is `getmetatable` (returns `nil` if no metatable present according to the [reference](https://www.lua.org/manual/5.1/manual.html#pdf-getmetatable)).
> 
> Right, what I meant there was that, since we verify/test that some `__index` are not called during regular usage via the `assert`, and the tests simulate regular usage (except `get_snippets`), we automatically test that no unwanted metatables are created.
> 
> > _Note:_ We currently create possibly unwanted metatables (for depths as well where `__index` always will fail due to `assert`), as I changed the `if` to an `assert`. I'm fine with this, what do you think? (more concrete example: The table `by_ft.snippets` has a metatable with "working" `__index`. The table `by_ft.snippets["lua"]` has a metatable with "asserting" `__index` as it is only supposed to act as array containing snippets but the metatable is still present)
> 
> That sounds perfect, I think we can't avoid the metatable if we want to assert that some tables aren't `__index`d

Alright nice.

atticus-sullivan:
> > Regarding tests, the files in `tests/*/*.lua` are executed in a `busted` environment (see [this repo](https://github.com/lunarmodules/busted))? And these tests would be unit tests right?
> 
> I think so, pretty much. We use neovims internal testing-engine, so the test-files have to be in the `tests`-directory and match `*_spec.lua`. For `get_snippets`, unit seems appropriate :+1:

Hm, I'm still struggling with the test structuring. Is it one file per testcase-type, one file per module or something else? (just because e.g. there is `split_lines_spec.lua` (which is only the name of the function) but there is also `utils_spec.lua` (which is the name of the module))

lua/luasnip/session/snippet_collection.lua
atticus-sullivan: Ah yes, to be honest, I didn't digged into how this order/store sorted worked (isn't that relevant for this PR, the only point of contact was removing the initial population of the tables though).
Should be fine by now (please check again).
(I'd like to keep the `order` member in the initial tables to indicate to any reader that this member exists (don't overuse the autocreating feature where there's no real benefit (`oder` will be created anyways soon) as it just makes reading/understanding much harder))

L3MON4D3:
Testfile-names are a bit all over the place😅
Tests are roughly split up by unit vs integration tests, for the files it makes more sense to use the modulename I'd say, so `unit/snippet_collection_spec.lua` would fine with me

lua/luasnip/session/snippet_collection.lua
L3MON4D3: `order` is just for quickly iterating the priorities in order.
Looks good, I definitely agree with keeping the empty tables 👍👍

atticus-sullivan:
Written some simple tests. If you have any ideas for more tests, let me know.

Regarding the last `TODO`: I think we should rename it to something like `copy_by_ft_type` and ``copy_by_ft_type_ft` since the functions only work on (a table containing) arrays.
We could of course use `pairs` in both cases and use just one generic function applicable for all depths, but I'm afraid that would hit the performance on copying (not sure if it is neglectable :shrug:).

After that TODO and if we don't need additional tests (or have no ideas on this), I'll update this branch with master (merge master into this one) and then I think we're finally ready to merge this PR.

lua/luasnip/session/snippet_collection.lua
```diff
@@ -59,16 +68,26 @@ local function insert_sorted_unique(t, k)
...
+		-- make new tables as they are indexed
+		return auto_creating_tables(s, k, 3)
+	end,
 	__newindex = function(t, k, v)
 		-- update priority-order as well.
 		insert_sorted_unique(t.order, k)
 		rawset(t, k, v)
 	end,
 }
+local prio_mt2 = {
```
L3MON4D3, line 82: A comment for usage of this table? Since it's only used much later (declaring it here is good though :+1:)

leiserfg:
> Merging leads to messed-up history

Not only that, it makes it painful to revert changes.

atticus-sullivan:
> Didn't know about `has_error`

Just looked it up as well ^^ (just guessed it should exist as lua already provides these `pcall` stuff). Didn't like it throwing errors which are used to indicate that the test failed but don't catch them.

(Addition: just added the acknowledgement to lua-users we talked about)

So everything should be done by now. Of course I can rebase ^^ (I definitely don't dislike rebasing, I normally just fully agree with the "rule" never rebase when the branch is out in the open and others might already have checked it out by now)

---

Off-topic: (things about rebasing)
nvm I'll move it to the [discussions part of the repo](https://github.com/L3MON4D3/LuaSnip/discussions/581) (there shouldn't be off-topic in a PR ^^)

lua/luasnip/session/snippet_collection.lua
L3MON4D3: Nice!

L3MON4D3:
Ah, yeah I think we're done here :+1:

(Should've included this in my request to rebase :sweat_smile::) Could you squash all the commits? I think 1-2 should suffice here.
(If you want to be extra fance, you can exclude the changes to `luasnip.txt`, I like that file exclusively being touched by the "Auto generate docs"-commits, but feel very free to ignore this)

atticus-sullivan:
Ah yes. Maybe
commit 1 -> normal changes
commit 2 -> adding tests
something like that.

Excluding `luasnip.txt`: Yes I didn't modify something on my own so just removing the auto generated commits should work, I guess.

In general I'd simply do an interactive rebase (removing commits and squashing the other ones), or do think of another (maybe better) way to do this?

L3MON4D3:
Mhm, you could soft-reset the entire branch and then make two new commits, excluding the changes to luasnip.txt.
But effort for both seems about equal, and the result is the same

atticus-sullivan:
Squashed and removed `doc/luasnip.txt` (the github action will add the `doc/luasnip.txt` again nonetheless).

Issue #518, opened 2022-08-12
Hot reloading: handle symlinks?


helixbass:
Hi, I just figured out (with the help of https://github.com/L3MON4D3/LuaSnip/issues/490#issuecomment-1186827332, thanks) that why hot reloading of a Lua snippets file wasn't working for me is that I was editing the snippets file via a symlink'd parent directory (specifically I symlink `~/.config/nvim` to a subfolder of my dotfiles repo ie `~/prj/dot/nvim`). It worked when instead I opened the snippets file specifying its "true" (non-symlink'd) file path

To be specific, in my config where I specify loading the snippets file I'm providing the symlink'd path, not the true path:
```
require('luasnip.loaders.from_lua').load({paths = '~/.config/nvim/snippets'})
```

So I wonder if it's possible to handle detecting editing of snippets files even if opened via a symlink'd file path?

L3MON4D3:
Ouhh I see, I guess the autocommand `BufWritePost` is only fired for exactly the name of the buffer..
So we have
1. load real directory
2. Add autocommand on BufWritePost for real path
3. Edit symlinked file
4. BufWritePost fires for symlinked, but not the real path

I'm afraid we can't do anything about that, we don't know the name of the symlinks to that file :/
We might have to switch to some other way to register snippetfile-changes(not sure if there are other autocommands for this) to correctly handle this case.
For now, you could use `:LuaSnipEdit` (search for that command in the DOC if it doesn't tell you anything😅) to edit the snippet-files.

L3MON4D3:
Mhmm, perhaps we should, instead of registering autocommands, collect all interesting files in some central table.
We could then have one autocommand, which, on BufWritePost, resolves the real path of the written buffer and then updates the snippets if it was a snippet-file

That would also remove some other rather brittle parts of the current procedure.. tempting :D

leiserfg:
I just read the code o the auto-reload, and yes it should be a single `autocmd` per loader instead of one per file, registering autocmds is very slow.

L3MON4D3:
Ouh, I did not know that.
Do you think it would make more sense to store the reload-function as a closure (simpler, a bit more uniform, maybe slower) or just as a loader-specific tuple of parameters to that function (easier to debug, since the parameters are right there)?

leiserfg:
I don't think they're very different in speed but storing only the data will take less space and will be easier to debug.

L3MON4D3:
Perfect👍
[...]

PR #526, opened 2022-08-14
Handle symlinks in loaders.

L3MON4D3:
Problem: See #518 
Solution: homogenize the reload-functions, store normalized paths (eg. without symlinks) and pass normalized filenames when reloading (which now needs just one autocommand, yay)

@leiserfg Could you take a look at this PR? I like having a second set of eyes on my code :D

### DIFF
--- a/tests/integration/loaders_spec.lua
+++ b/tests/integration/loaders_spec.lua
@@ -146,17 +146,21 @@ describe("loaders:", function()
 			feed(edit_keys)
 
 			exec_lua("ls.expand()")
-			screen:expect({
-				grid = [[
+
+			-- undo changes to snippet-file before checking results.
+			feed("<Esc><C-I>u:w<Cr><C-O>")
+
+			-- re-enter current placeholder
+			exec_lua("ls.jump(-1)")
+			exec_lua("ls.jump(1)")
+
+			screen:expect({grid=[[
 				replaces? jumps? ^  !                              |
 				{0:~                                                 }|
 				{0:~                                                 }|
 				{0:~                                                 }|
 				{2:-- INSERT --}                                      |]],
 			})
-
-			-- undo changes to snippet-file.
-			feed("<Esc><C-I>u:w<Cr>")
 		end)
 	end
 


lua/luasnip/loaders/init.lua
```diff
@@ -96,4 +89,15 @@ function M.load_lazy_loaded(fts)
 	end
 end
 
+vim.cmd([[
+		augroup luasnip_watch_reload
+		autocmd BufWritePost * lua require("luasnip.loaders").reload_file(vim.fn.expand("<afile>"))
+		augroup END
+	]])
+function M.reload_file(filename)
```
leiserfg, line 97: I think the normalization of the path should happen here instead of in each loader module.

L3MON4D3: Mhh yeah, that might be better.
I put them into each loader since it makes for a nicer public interface (whatever kind of path is passed to it, it works), but `reload_file` isn't even public.
I'll make that a bit more obvious (prefix with __ or something), and do the normalization outside those functions. TY👍

L3MON4D3:
I'll add some tests for reloading symlinked dirs later, too

L3MON4D3:
Concerning the failing test, it looks like `scriptnames` in `nvim -u NONE` now prints nothing.
Any ideas what to use as the default now?

(Didn't spot this locally since I always run tests from within nvim, where the variable is ofc. set)

leiserfg:
Crap, that's a breaking change, let's complain to @clason, if they can't give us luajit they can't take scriptnames from us without notification :smile:

leiserfg:
I think the issue is not `-u NONE` as it 's always empty in that case, the problem is that when using  `-u something.lua`  the first script is not the user config anymore, isn't?

clason:
You can do `nvim --clean -u myinit.lua` (order matters).

leiserfg:
I never understood why using `-u` does not set `$MYSCRIPT`.  if `scriptnames` does not work, we can only use `vim.v.vargs` and extract the value of `-u` from it.

leiserfg:
I'll make a PR to implement it.

L3MON4D3:
> > Concerning the failing test, it looks like scriptnames in nvim -u NONE now prints nothing.
> > Any ideas what to use as the default now?
> 
> `nvim --clean` (since a long time already -- `NONE` really means "none", not even builtin scripts!)

Weird that it only occurred just now, maybe there's more to it than just "scriptnames is empty".
I'll take another look at what went wrong later.

L3MON4D3:
> I'll make a PR to implement it.

👍👍

leiserfg:
https://github.com/L3MON4D3/LuaSnip/pull/528

Issue #81, opened 2021-07-23
Breaking Changes


L3MON4D3:
Watch this issue to be notified of any breaking changes.

thanks to @clason for recommending this

L3MON4D3:
## 23.07.2021
09716b1

 Changed default for `wordTrig` to true. If you have any snippets that rely on the default being `false`, `wordTrig` must now be explicitly set to `false`, eg.
```lua
s({trig = "trigger anywhere"}, {...})
```
or
```lua
s("trigger anywhere", {...})
```
to
```lua
s({trig = "trigger anywhere", wordTrig=false}, {...})
```

L3MON4D3:
## 24.07.2021

e9f4d03

Move pseudo-nodes (`rep` and `partial`) from `luasnip.util.functions` to `luasnip.extras`.

These functions are probably imported once:
```lua
local r = require("luasnip.util.functions").rep
local p = require("luasnip.util.functions").partial
```
Those lines must be changed to 
```lua
local r = require("luasnip.extras").rep
local p = require("luasnip.extras").partial
```

L3MON4D3:
## 02.08.2021

eb4680f

Pass function/dynamicNodes their arg-text without snippet-indent, prepend snippet-indent to returned text.

Should affect very few setups, if a functionNode relies on the previous behaviour, they can now be simplified.

L3MON4D3:
## 30.08.2021

d2aa43c

Fixes parsing of `{`,`}` in lsp-snippets, only closing brackets have to be escaped, opening ones don't.
In fixing, the parser also got more strict, previously `"\\begin{$1}"` would've parsed just fine, now the closing `}` has to be escaped(`"\begin{$1\}"`).

L3MON4D3:
## 05.09.2021

8361b35

The functions used in `matchNode` are no longer only passed the `'\n'`-joined text from the node. Instead, they now get the same object as the function passed to `functionNode` ie. `{{line1, line2}, snip}`.

Changing
```lua
function(text)
	-- do something with text.
end
```

to

```lua
function(args)
	local text = table.concat(args[1], "\n")
	-- do the same thing with text.
end
```

should do the trick.

L3MON4D3:
## 21.09.2021

cb4fe49

Changes the signature of the function passed to function/dynamicNodes from
```lua
-- snippet inside the table of argnodes' text.
function({argnode_text, snip}, ...) end
```

to
```lua
[...]

PR #528, opened 2022-08-15
Extract config root from pwd instead of using scriptnames

leiserfg:


L3MON4D3:
Makes sense👍
Could you apply this commit to the first PR and see whether tests still pass?
What do you think about doing a different kind of fallback? I mean, if the config is not sourced, there just is no CONFIG_ROOT. It might be more sensible to fallback to.. maybe just HOME

leiserfg:
This allows to copy your full config to another folder (like  `.config/nvim/`  to `/tmp/nvim`)  do some changes and be able to use  `nvim -u /tmp/nvim/init.lua`

leiserfg:
I rebased your branch on top of this one and the tests pass.

L3MON4D3:
> This allows to copy your full config to another folder (like `.config/nvim/` to `/tmp/nvim`) do some changes and be able to use `nvim -u /tmp/nvim/init.lua`

Okay, yeah that makes some sense, but wouldn't unsourced `plugin`-files cause issues then?

L3MON4D3:
> I rebased your branch on top of this one and the tests pass.

Ahh, great👍

leiserfg:
> > This allows to copy your full config to another folder (like `.config/nvim/` to `/tmp/nvim`) do some changes and be able to use `nvim -u /tmp/nvim/init.lua`
> 
> Okay, yeah that makes some sense, but wouldn't unsourced `plugin`-files cause issues then?

Other plugins if they depend on the path (most o the time they don't), we do because we (potentially) read files relative to the config.

L3MON4D3:
No, I mean if you have some stuff in your local `ftplugin`, for example. That wouldn't be sourced, right?
Like, is it an usual workflow to just copy everything?

The reason I'm bringing this up is because, beside `-u something`, `--clean` also leads to an empty `$MYVIMRC`, and I'm not sure those two are all cases, so handling each differently by checking commandline-arguments seems brittle.
(I think a simple "if not set then set to ..." might make more sense, since it's predictable what `./` means (maybe just current directory might be best, even), but if you have another solution, I'm all ears)

leiserfg:
I'm just going for "keeping the previous behavior", because the only use case I know for checking a nonexisting `$MYCONFIG` is when testing configs. I don't think the `--clean` only case is important as in that case the plugin won't even load.

L3MON4D3:
Keeping the existing behaviour makes sense, that's right..
But even in `--clean`, the plugin might be loaded manually in some way.

Any lua-config loaded via `-u lua_config.lua` is probably one crafted specifically for this occasion (MWE, probably). As such, we don't need to ensure compatibility with whatever setup exists otherwise (so it doesn't matter if a `load`-path has to be changed).

Setting `$MYCONFIG_ROOT` to the cwd seems the most sensible thing for MWEs, especially since then, we can just state in the docs that, if `$MYVIMRC` is not set, `$MYCONFIG_ROOT` resolves to cwd, which is precise(as opposed to manually handling commandline-arguments) and simple.

leiserfg:
Then it is a breaking change, I don't think many will suffer with it but still one so it should be in [Breaking Changes](https://github.com/L3MON4D3/LuaSnip/issues/81). Fortunately, it was a non-documented feature.

L3MON4D3:
> Then it is a breaking change, I don't think many will suffer with it but still one so it should be in [Breaking Changes](https://github.com/L3MON4D3/LuaSnip/issues/81). Fortunately, it was a non-documented feature.

no doc -> no breaking changes, ezpz :P

I'll add it there before merging :+1:

lua/luasnip/util/path.lua
```diff
@@ -51,18 +51,15 @@ function Path.read_file(path)
 	return buf
 end
 
-local MYCONFIG_ROOT = vim.env.MYVIMRC
--- if MYVIMRC is not set then it means nvim was called with -u
--- therefore the first script is the configuration
--- in case of calling -u NONE the plugin won't be loaded so we don't
--- have to handle that
+local MYCONFIG_ROOT = vim.env.MYCONFIG_ROOT
```
L3MON4D3, line 54: This is here so it can be overridden from the outside?

lua/luasnip/util/path.lua
```diff
@@ -51,18 +51,15 @@ function Path.read_file(path)
...
 
 if not MYCONFIG_ROOT then
-	MYCONFIG_ROOT = vim.fn.execute("scriptnames"):match("1: ([^\n]+)")
+	if vim.env.MYVIMRC then
+		MYCONFIG_ROOT = vim.env.MYVIMRC:gsub(("%s[^%s]+$"):format(sep, sep), "")
```
L3MON4D3, line 58: What do you think of using `fnamemodify` here?
Would make this a bit more readable

lua/luasnip/util/path.lua
leiserfg: Yes, if you don't like it I can remove it, I thought it was what you meant in your last comment.

leiserfg:
> > Then it is a breaking change, I don't think many will suffer with it but still one so it should be in [Breaking Changes](https://github.com/L3MON4D3/LuaSnip/issues/81). Fortunately, it was a non-documented feature.
> 
> no doc -> no breaking changes, ezpz :P
> 
> I'll add it there before merging +1

https://xkcd.com/1172/

lua/luasnip/util/path.lua
L3MON4D3: Ohhh, no I didn't mean that. Though I can see I didn't word that all that well.
I don't think we need that mechanism.

L3MON4D3:
> > > Then it is a breaking change, I don't think many will suffer with it but still one so it should be in [Breaking Changes](https://github.com/L3MON4D3/LuaSnip/issues/81). Fortunately, it was a non-documented feature.
> > 
> > 
> > no doc -> no breaking changes, ezpz :P
> > I'll add it there before merging +1
> 
> https://xkcd.com/1172/

Ahhh that's a good one xD

L3MON4D3:
Okay, code's good, could you rebase so this is just one commit? (best include `Breaking Change` in the first line of the message).

L3MON4D3:
Almost forgot about this one :sweat_smile:

PR #530, opened 2022-08-15
Add Setting: Square-Branckets is the default-delimiters

yanskun:
`fmt` for `{}`.
`fmta` for `<>`.
each with their default delimited.

When I am making snippets for jsx, both `{}` and `<>` are used a lot, but `[]` is rarely used.
Therefore, I added `fmts` (an acronym for square) in addition to `fmta` because I thought it would be more convenient to have it without having to change the repeat setting.

thank you.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1128,8 +1128,9 @@ ls.add_snippets("all", {
   * `dedent`: remove indent common to all lines in `format`. Again, makes
   	passing multiline-strings a bit nicer (default true).
 
-There is also `require("luasnip.extras.fmt").fmta`. This only differs from `fmt`
-by using angle-brackets (`<>`) as the default-delimiter.
+Also `require("luasnip.extras.fmt").fmta` and `require("luasnip.extras.fmt").fmts`. This is
+fmt`, using the default delimiters of the angle-brackets (`<>`) for fmta and the
+square-brackets (`[]`) for fmts, respectively.
 
 
 ## On The Fly snippets
--- a/DOC.md
+++ b/DOC.md
@@ -1128,8 +1128,8 @@ ls.add_snippets("all", {
   * `dedent`: remove indent common to all lines in `format`. Again, makes
   	passing multiline-strings a bit nicer (default true).
 
-Also `require("luasnip.extras.fmt").fmta` and `require("luasnip.extras.fmt").fmts`. This is
-fmt`, using the default-delimiters of the angle-brackets (`<>`) for fmta and the
+There are also `require("luasnip.extras.fmt").fmta` and `require("luasnip.extras.fmt").fmts`.
+This is `fmt`,　using the default-delimiters of the angle-brackets (`<>`) for fmta and the
 square-brackets (`[]`) for fmts, respectively.
 
 
--- a/DOC.md
+++ b/DOC.md
@@ -1129,7 +1129,7 @@ ls.add_snippets("all", {
   	passing multiline-strings a bit nicer (default true).
 
 Also `require("luasnip.extras.fmt").fmta` and `require("luasnip.extras.fmt").fmts`. This is
-fmt`, using the default delimiters of the angle-brackets (`<>`) for fmta and the
+fmt`, using the default-delimiters of the angle-brackets (`<>`) for fmta and the
 square-brackets (`[]`) for fmts, respectively.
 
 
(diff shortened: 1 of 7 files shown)


L3MON4D3:
Mhmmm, I can see your point.
I'm a bit conflicted: This is easily achieved in one's own config (by just wrapping `fmt` like you have done), but it might come in handy if it's already available in luasnip.

But I think there's not really any downside to just including this :)

Examples/snippets.lua
```diff
@@ -279,12 +280,14 @@ ls.add_snippets("all", {
 		})
 	),
 	-- The delimiters can be changed from the default `{}` to something else.
-	s("fmt4", fmt("foo() { return []; }", i(1, "x"), { delimiters = "[]" })),
+	s("fmt4", fmt("foo() { return (); }", i(1, "x"), { delimiters = "()" })),
```
L3MON4D3, line 283: I don't really like this change, in almost all languages, it would be very annoying to use `()`, `[]` just make more sense since they are more rarely used (as you have realised :D)

What do you think of, instead, amending the comment above, stating that `[]` is also available, directly, via `fmts`?

DOC.md
```diff
@@ -1128,8 +1128,9 @@ ls.add_snippets("all", {
   * `dedent`: remove indent common to all lines in `format`. Again, makes
   	passing multiline-strings a bit nicer (default true).
 
-There is also `require("luasnip.extras.fmt").fmta`. This only differs from `fmt`
-by using angle-brackets (`<>`) as the default-delimiter.
+Also `require("luasnip.extras.fmt").fmta` and `require("luasnip.extras.fmt").fmts`. This is
```
L3MON4D3, line 1131: Looks like the first sentence is missing some words? :sweat_smile:
(and a small mishap with `` ` `` in the line below)

L3MON4D3:
No complaints about the code :D

leiserfg:
IMHO instead of having a lot of custom `fmt` we should just have the default one (the one with `{}`) and a  factory function to create custom ones with whatever delimiter one wanna use. Like 
```lua
local fmta = fmt_factory{del={"<", ">"}}
local fmt_jsx = fmt_factory{del={"{%", "%}"}}
```

L3MON4D3:
Oh, that's the correct approach (simplify the personalize-approach :+1::+1:), and a really cool pattern.

It could even come in handy in other places, like having a dedicated function for creating snippets with `regTrig` set.
Maybe we create one generic version that is capable of overriding a key in any parameter-position, and then more specific ones for `fmt`, `snippet`, ...

Could live as a module in extras.

Examples/snippets.lua
yanskun: Certainly this is not good because it is not common.
I changed it because I didn't want to duplicate fmts, but I'll change it back.

DOC.md
yanskun: Huh? Sorry, this is a boring mistake.

yanskun: I was totally unaware of it. Thank you very much.

yanskun:
@leiserfg @L3MON4D3 
Indeed, I understand that if we were to make them one by one, there would be no end to it.
However, I still think it would be worth it.

If you decide you don't want it, please CLOSE it!
It's a lot of fun to make suggestions and get feedback!

leiserfg:
I will actually prefer it if you take over and implement the factory one, so we keep the API  flexible but with simple defaults. WDYT?

yanskun:
I understand that simple is beautiful. I completely agree!😁

However, I looked at the documentation and found a customized fmta, so I guess I can grow fmts. So I made it.

yanskun:
I had never heard of the slang 'WDYT' before!

L3MON4D3:
> Indeed, I understand that if we were to make them one by one, there would be no end to it.
However, I still think it would be worth it.

Dammit, this feels like https://xkcd.com/1801, or https://xkcd.com/1445..
Okay I think it makes _most_ sense to just have the one `fmt` and a good framework for extending it, and _more_ sense to accept that we have `fmta`, for some reason, but not other fmt-presets.

That way, we can use `fmts` as a nice example for the generic case, once we get to implementing that :P

> However, I looked at the documentation and found a customized fmta, so I guess I can grow fmts. So I made it.

And I applaud the initiative, I'm sorry we can't make use of your work :/

yanskun:
@leiserfg @L3MON4D3

Thank you both!

L3MON4D3:
Working on the generic implementation in #535

PR #531, opened 2022-08-16
unset needs to be in the same subshell as the make command

atticus-sullivan:
Since `make` by default (can be changed with [`.ONESHELL`](https://www.gnu.org/software/make/manual/html_node/One-Shell.html)) creates a new subshell for each new command, the current way of unsetting does not work.
We need to either use `.ONESHELL` or (in my opinion the nicer way) do `unset LUA_PATH LUA_CPATH ; ...`


L3MON4D3:
Ohh, I was not aware of this!
TY!

PR #535, opened 2022-08-17
Make it easy to create functions that override the defaults of functions.

L3MON4D3:
Sounds a bit cryptic, the general idea is to allow stuff like
```lua
-- this somewhere above: local override_fn = require("luasnip.extras.override_fn")
local fmts = override_fn.fmt({delimiters = "[]"})
```
This makes it easy to create functions with different defaults, like above for `fmt`, or maybe a `ls.snippet`-function that always sets `regTrig` or some combination of parameters. 

This needs some more work:
- [x] snippet does not really work yet, if the first arg is a string, we're in trouble.
- [x] add more functions beside `fmt`
- [x] think about exposing and documenting the generic function (`override_any`, currently) so it can be used to craft yet more of these override-functions (although that is really quick)
- [x] think of a better name, `override_fn` isn't very self-explanatory

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1219,6 +1219,78 @@ Contains some utility-functions that can be passed to the `ft_func` or
   })
   ```
 
+# EXTEND_DECORATOR
+
+Most of luasnip's functions have some arguments to control their behaviour.  
+Examples include `s`, where `wordTrig`, `regTrig`, ... can be set in the first
+argument to the function, or `fmt`, where the delimiter can be set in the third
+argument.  
+This is all good and well, but if these functions are often used with
+non-default settings, it can become cumbersome to always explicitly set them.
+
+This is where the `extend_decorator` comes in:  
+It can be used to create decorated functions which always extend the arguments
+passed directly with other, previously defined ones.  
+An example:
+```lua
+local fmt = require("luasnip.extras.fmt").fmt
+
+fmt("{}", {i(1)}) -- -> list of nodes, containing just the i(1).
+
+-- when authoring snippets for some filetype where `{` and `}` are common, they
+-- would always have to be escaped in the format-string. It might be preferable
+-- to use other delimiters, like `<` and `>`.
+
+fmt("<>", {i(1)}, {delimiters = "<>"}) -- -> same as above.
+
+-- but it's quite annoying to always pass the `{delimiters = "<>"}`.
+
+-- with extend_decorator:
+local fmt_angle = ls.extend_decorator.apply(fmt, {delimiters = "<>"})
+fmt_angle("<>", {i(1)}) -- -> same as above.
+
+-- the same also works with other functions provided by luasnip, for example all
+-- node/snippet-constructors and `parse_snippet`.
+```
+
+`extend_decorator.apply(fn, ...)` requires that `fn` is previously registered
+via `extend_decorator.register`.  
+(This is not limited to luasnip's functions!)  
+(although, for usage outside of luasnip, best copy the source-file
+`/lua/luasnip/util/extend_decorator.lua`).
+
+`register(fn, ...)`:
+* `fn`: the function.
+* `...`: any number of tables. Each specifies how to extend an argument of `fn`.
+  The tables accept:
+  * arg_indx, `number` (required): the position of the parameter to override.
+  * extend, `fn(arg, extend_value) -> effective_arg` (optional): this function
+    is used to extend the args passed to the decorated function.
+    It defaults to a function which just extends the the arg-table with the
+    extend-table (accepts `nil`).
+    This extend-behaviour is adaptable to accomodate `s`, where the first
+    argument may be string or table.
+
+`apply(fn, ...) -> decorated_fn`:
+* `fn`: the function to decorate.
+* `...`: The values to extend with. These should match the descriptions passed
+  in `register` (the argument first passed to `register` will be extended with
+  the first value passed here).
+
+One more example for registering a new function:
+```lua
+local function somefn(arg1, arg2, opts1, opts2)
+	... -- not important
+end
+
+-- note the reversed arg_indx!!
+extend_decorator.register(somefn, {arg_indx=4}, {arg_indx=3})
+local extended = extend_decorator.apply(somefn,
+	{key = "opts2 is extended with this"},
+	{key = "and opts1 with this"})
+extended(...)
+```
+
 # LSP-SNIPPETS
 
 Luasnip is capable of parsing lsp-style snippets using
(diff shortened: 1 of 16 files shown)


leiserfg:
WTYT if instead of the `override_fn.func` api you just expose `override_with(func, opt_overrides)` in `luasnip.utils`?

L3MON4D3:
The issue with that is that we don't know which position the `opts` are in, or how many there are (think `ls.s`, one might like to override both the first or third parameter).
So we'd best provide presets for common functions (like `fmt`, `s`, `add_snippets`), that's why I put that mechanism into its own module for now.

> The same function could be used for creating auto-expand snippets by wrapping `add_snippets` with the auto_expand (is that the name?).

Yeah right, via `type = "autosnippets"`

L3MON4D3:
Did a small rewrite, now it's possible to override the snippet-parameters.
Also, extend instead of override would make more sense I think.
Example:
```lua
ls.setup_snip_env()

local override_fn = require("luasnip.extras.override_fn")
local expand_conditions = require("luasnip.extras.expand_conditions")

local fmts = override_fn.fmt({delimiters = "[]"})
-- equivalent to ultisnips br (well, bp for pattern :D)
local br_snip = override_fn.snippet(
	{regTrig = true},
	{condition = expand_conditions.line_begin} )

ls.add_snippets("all", {
	br_snip("%d", {
		i(1, "asdf")
	}),
	s("trig", fmts("[]", {i(1, "asdf")} ) )
}, {
	key = "asd"
})
```

I think this is really sick, great idea @leiserfg!!

leiserfg:
I was riding my :bike: and an idea came to my mind. What if we add a `fun._opt_pos` to every function that can be overridden, so you don't have to import everything inside of `luasnip.extras.override_fn`? That way you only expose `override_fn` and it uses `._opt_pos` to know where to override. The decorated function should get the same `_opt_pos as the original one, that way a function can be wrapped more than once. The default value should be 2, maybe it can be also a 3rd parameter of the function, in case someone wanna override a function that is not shipped by luasnip.

L3MON4D3:
Mhmmm interesting idea :+1:
I definitely like defining properties of the function (position of the `opts`-arg, how can it be overridden) close to the function, as opposed to in a completely different file.

To implement this, we'd have to turn all outward-facing functions into tables with `__call`, right? (Don't like this very much, although we could hide this behind a nice constructor, so at least the code doesn't look completely weird)
(Can you think of a better approach for the implementation?)

Communicating which functions are overridable would be a bit harder this way, by putting all of the overridable functions into one module it's very clearly defined.
(We don't have this problem if we make all functions overridable though xD)
(OTOH, we can just return a readable error in the function that's responsible for overriding, yeeah that works)

Yeah, I think this is even better :+1::+1:

leiserfg:
> To implement this, we'd have to turn all outward-facing functions into tables with __call, right? 
Crap, I forgot functions are not indexable in lua :face_exhaling:

leiserfg:
But also, it does not need to make every  function a table, you can just expose a function to annotate where the opt goes, like so:
```lua
local opt_pos = setmetatable({}, {__mode = "kv"})
function M.overridable(fn, pos)  
   opt_pos[fn] = pos
   return fn
end

local function override(fn, opt, maybe_pos)
-- if maybe_pos use it other way check the the table, if none is there, consider it's 2 or fail, the one you prefer
-- call  `overridable` on the decorated function to set the same position as the used 
end
```

L3MON4D3:
> > To implement this, we'd have to turn all outward-facing functions into tables with __call, right?
> 
> Crap, I forgot functions are not indexable in lua :face_exhaling:

xD
What language is that possible in?

> But also, it does not need to make every function a table, you can just expose a function to annotate where the opt goes, like so:

Oh, yeah I like that approach :+1:

Sidenote:
Making the keys weak is a nice touch, but I don't get why the values are weak too, aren't they only stored in this table?

leiserfg:
> xD
> What language is that possible in?

I use python 90% of the time :smile:, there you can.

L3MON4D3:
Oho, I use Python sometimes but have never seen that, cool

leiserfg:
> Making the keys weak is a nice touch, but I don't get why the values are weak too, aren't they only stored in this table?

Yes, `kv` is not needed, `k` is good enough.

L3MON4D3:
Okay, works pretty well, I've renamed everything to reflect that we're extending, not overriding the passed args (not sure what's easier to understand, overriding defaults or extending passed arguments...)
Here's what usage looks like now:
```lua
ls.setup_snip_env()

local ed = require("luasnip.util.extend_decorator")
local expand_conditions = require("luasnip.extras.expand_conditions")

local fmts = ed.apply(fmt, {delimiters = "[]"})
-- can also apply to already-decorated functions.
local fmtd = ed.apply(require("luasnip.extras.fmt").fmta, {dedent = false})

-- equivalent to ultisnips br (well, bp for pattern :D)
local br_snip = ed.apply(s,
	{regTrig = true},
	{condition = expand_conditions.line_begin} )

ls.add_snippets("all", {
	br_snip("%d", {
		i(1, "asdf")
	}),
	s("trig", fmts("\t\t\t[]", {i(1, "asdf")} ) ),
	s("triga", fmtd("\t\t\t<>", {i(1, "asdf")} ) )
}, {
	key = "asd"
})
```
Also moved it to `util`, since we now rely on it internally.
It's also pretty straightforward to register personal functions, I think we should properly expose this.
I've decided not to set default-values for the `opts`-parameter, too easy to introduce subtle errors imo. Maybe I'll also remove the defaulting for the `extend`-function and just export the default-extend from the module, I think that's okay. Maybe export it from util, under a fitting name.

leiserfg:
Yes, it makes sense to re-export it (as we did for the function to add environments) and only document that exposed part.

L3MON4D3:
Ahh, yeah, that file is so small it's fine to re-export :+1:
Can you think of useful functions to extend?
I guess
* node-constructors (not thaaaaaaat useful, but one can more easily create highlighted nodes that way, like `red_insertNode` xD)
* `parse`: once the parser-update is merged, it'll also have `dedent` and `trim_empty` like `fmt`, but default false to not mess up snippets from lsp. So overriding that is an easy target.
* ???

will come in handy, but those are the only ones that come to mind directly.

L3MON4D3:
@leiserfg Would you take a look?

leiserfg:
> @leiserfg Would you take a look?

Sure, tonight I will review it.

leiserfg:
Looks awesome :tada:

L3MON4D3:
Hell yeah, that's what I wanna hear :D
Thank you!

PR #539, opened 2022-08-18
doc: update readme.md install instructions

jrowlingson:


gegoune (commented):
Neither of code snippets will work as they are. Is it really necessary? shouldn't be plugin's responsibility to teach users how to use plugin manager they chose in the first place.

L3MON4D3:
I agree with @gegoune, showing that one line does not make a difference.
You can add the link to packer though, it's a better recommendation than `vim-plug`

jrowlingson:
Definitely not necessary - just a convenience I've seen [elsewhere](https://github.com/nvim-telescope/telescope.nvim#installation). Agree that anyone landing here is probably already going to be familiar with using a plugin manager.

L3MON4D3:
This looks good, ty!
Could you squash the commits into one?

L3MON4D3:
Nice, thank you!!

PR #540, opened 2022-08-18
doc: doc.md alternate function node example

jrowlingson:
I don't know if I made this example more intense than it needs to be, but, writing it out this way and running it helped my understanding. The existing example does not match the gif. Double spaced to squeeze the comments in. Let me know and I'll add the gif.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -297,19 +297,29 @@ reasons, check out Luasnip#110).
 
 Function Nodes insert text based on the content of other nodes using a
 user-defined function:
+
 ```lua
- s("trig", {
- 	i(1),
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Yeah, looks good, makes sense to explain what each arg does in small comments :+1:
(+ the old one was plain incorrect with the `i(0)`, so thank you for fixing that up).
With the updated gif this is good to go :)

jrowlingson:
Shaved the yak a bit to get this gif :laughing: (had to update ubuntu). I see that the doc generation step is adding an extra commit - shall I squash? (there's also the squash and merge option if you're using the github web client).

L3MON4D3:
TY!!

L3MON4D3:
(missclick :grimacing:)

PR #543, opened 2022-08-20
Move setup to root module

leiserfg:
This does not break old configs as we still keep `setup` and `set_config` in the `luasnip.config` module.


L3MON4D3:
Niice, yup :+1:
Could you amend that one occurrence in `Examples/snippets.lua` too?

L3MON4D3:
Clean clean clean, thank you :+1:

Issue #548, opened 2022-08-22
worry number of arguments to 'insert' (Path.normalize failed on Windows)


Verf:
Error Message: `Error running config for LuaSnip: ..packer\start\LuaSnip\lua\luasnip\loaders\from_vscode.lua:142: wrong number of arguments to 'insert'`

I found the last commit 2368d62 use Path.normalize to normlize path, but this function failed on Windows system:
```lua
Path.normalize(Path.join(root, snippet_entry.path))
-- root -> C:\Users\fht\AppData\Local\nvim\vsnip
-- snippet_entry.path -> ./global.json
-- Path.join(root, snippet_entry.path) -> C:\Users\fht\AppData\Local\nvim\vsnip\./global.json
-- Path.normalize(Path.join(root, snippet_entry.path)) -> nil
```

A quick and dirty fix is :
```lua
local filename = string.gsub(snippet_entry.path, "^%.[%/%\\]", "")
table.insert(
    ft_files[ft],
    Path.normalize(Path.join(root, filename)) 
) 
```

c3n21:
I'm getting the same error on Linux

```
Error in packer_compiled: ...packer/start/LuaSnip/lua/luasnip/loaders/from_vscode.lua:142: wrong number of arguments to 'insert'
```

if I hard reset to this hash

```
fc25e77d74efb06033355b2175e2f5b78d15e7df
```

the error disappears

leiserfg:
@c3n21 can you share how are you using the loader? To try to reproduce the error.

leiserfg:
@Verf can you check if this (no so dirty fix)  works in windows? 
function Path.join(...)
	return vim.fs.normalize(table.concat({ ... }, sep))
end

Verf:
> @Verf can you check if this (no so dirty fix) works in windows?
> 
> ```lua
> function Path.join(...)
> 	return vim.fs.normalize(table.concat({ ... }, [[/]]))
> end
> ```

It can only be worked on the nightly build, but can not work on the latest stable build (nvim v0.7.2), because vim.fs does't exist.

leiserfg:
Well, but if it works then we could vendorize it.

Verf:
That's it.

c3n21:
> @c3n21 can you share how are you using the loader? To try to reproduce the error.

```lua
local luasnip = require('luasnip')
local types = require('luasnip.util.types')

luasnip.config.set_config({
    history = true,
[...]

PR #549, opened 2022-08-22
Use unix separators

leiserfg:


leiserfg:
Should address #548 , blocked by someone testing it on windows   ( @Verf )

L3MON4D3:
Looks good, gj with the quick fix :+1:

leiserfg:
Crap, this depends on nightly, I will have to "vendorize" normalize.

lua/luasnip/util/path.lua
```diff
@@ -2,20 +2,13 @@ local Path = {}
...
-			return "/"
-		else
-			return "\\"
-		end
-	end
-	return package.config:sub(1, 1)
-end)()
+-- In windows, unix style of paths is also valid
+function Path.unixfy(path)
+	return path:gsub("\\", "/")
```
L3MON4D3, line 7: A unix path could contain `\` naturally :/

leiserfg: Didn't though about that (neither neovim devs :smile: )

L3MON4D3: :grimacing::grimacing:

leiserfg: I'll use the `cappital_letter:` pattern to detect if it's a windows path.

L3MON4D3: Are we sure relative paths won't be passed to unixify?

leiserfg: We are the only users of it,  it's used in `join` and we only call join starting from absolute paths, the other usage is `expand` but that one resolves to an absolute path before calling it.

L3MON4D3: Mhmm, okay it would work for now, but just thinking of the unexpected errors and debugging _if_ a relative path somehow makes its' way there.. :grimacing:
(we could rename all functions to reflect that they're supposed to handle only relative paths, but that's hardly a good solution)

I think the proper way to tackle this issue is to improve the normalization of paths, so always changing `\` to `/`, or, if they're relative, remove leading `./` (for this problem, for example, it would be enough to remove the `./` from paths in `package.json`).

But, I'd wait until we understand the full behaviour of `fs_realpath` on windows, then we can make up for corner-cases not covered by it.

PR #551, opened 2022-08-26
doc: improve section on fmt.

L3MON4D3:


leiserfg (approved):

L3MON4D3:
Thank you for already taking a look @leiserfg :)
I think I'll throw the other changes ("jump index" everywhere) mentioned in #550 in here too.

leiserfg:
Maybe add a Notes section and there explain it and link it in the entries referring to it? So you don't have to put the same in several places?

L3MON4D3:
Do you mean for explaining `jump-index`? Right, that's actually missing :sweat_smile:
But I don't really get what's duplicated right now

leiserfg:
It is not, just  if you will document the behaviour in other places you could just link it.

L3MON4D3:
Makes sense :+1:
I've though of doing something like that for the argnodes passed to function/dynamicNode, since they can be 
* just an integer for the jump-index
* `absolute_indexer`
* or even just a node
(or, I guess just refer to them by a proper name as well.. maybe "node-reference")

L3MON4D3:
Though that is a pretty big number of links.. But you're right, it will come in handy.
Maybe a glossary, preferably at the end?

DOC.md
```diff
@@ -977,10 +979,13 @@ is only a short outline, their usage is shown more expansively in
...
-manipulation. For example, to replace all occurences of "a" in the nth insert
-with "e", one could use `lambda(lambda._1:gsub("a", "e"), n)` (signature is
-similar to that of `functionNode`).
-If a node has multiple lines, they will be concatenated using "\n".
+  manipulation.  
+  For example, to generate text where all occurences of "a" in the text of the
+  node with jump-index `n` are replaced with "e", one could use
+  `lambda(lambda._1:gsub("a", "e"), n)` (signature is similar to that of
+  `functionNode`).  
+  `lambda._n` returns the text of the `n`th argnode. If this text has multiple
```
danielo515, line 987: does argnode here refer to jump-index too?

DOC.md
```diff
@@ -228,18 +228,18 @@ s("trigger", {
...
-1. After expansion, we will be at InsertNode 1.
-2. After jumping forward, we will be at InsertNode 2.
-3. After jumping forward again, we will be at InsertNode 0.
+1. After expansion, the cursor is at InsertNode 1,
+2. after jumping forward once at InsertNode 2,
+3. and after jumping forward again at InsertNode 0.
 
 If no 0-th InsertNode is found in a snippet, one is automatically inserted
-after all other nodes.
+behind other nodes.
```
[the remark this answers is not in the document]
L3MON4D3: Okay, I'll change it back :+1:

DOC.md
```diff
@@ -707,7 +709,7 @@ user-input.
 The prototype for the dynamicNodes' constructor is 
 `d(position:int, function, argnodes:table of nodes, opts: table)`:
 
-1. `position`: just like all jumpable nodes, when this node will be jumped into.
+1. `jump_index`: just like all jumpable nodes, its' position in the jump-list.
```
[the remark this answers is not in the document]
L3MON4D3: That's the change you recommended, right?

DOC.md
L3MON4D3: Oh, this is a bit subtle:
`l(l._1 .. l._2, {2,1})` (`l` as `lambda`), would return the text from the node with jump-index **2** followed by that of **1**.
So the order of argnodes only depends on the order of that list (second argument), not their jump-index.
But this could definitely be more clear, I'll reference the args passed to function/dynamicNode I think.

DOC.md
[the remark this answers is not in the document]
danielo515: Yes, exactly. Sorry, I added this comment first then I realized I can suggest changes to unchanged lines too.

DOC.md
danielo515: Oh, I see what you mean. So this actually refers to the table indexes within the user_arguments argument passed to the lambda (which is the third argument in a normal function). Definitely very specific

L3MON4D3: Ah, not `user_arguments`, that's for more easily re-using similar functions.
`functionNode(fn, argnode-references, opts)`, and `fn` is `function(argnode-text, snippet, user_args)`
(where `argnode_references` could be a list of `jump-indices` (or just a single one)), and `l(lambda-expression, argnode-references)`

DOC.md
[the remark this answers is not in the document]
L3MON4D3: Just making sure I didn't miss something :)

L3MON4D3:
I've opted for working the sections that explain those concepts into the text, feels more natural than a big glossary at the end IMO, but happy to hear reasons for doing it differently :D

danielo515:
That sounds awesome!

L3MON4D3:
Okay I think this is done, I'll hold off from merging until #612 is in, don't wanna complicate that.

I've though about turning list-entries in `API-REFERENCE` to subsections (+: can link directly to those, -: it's pretty unnecessary for short stuff, which is most of these), but don't think it's worth it

L3MON4D3:
Oookay, gonna have one last look tomorrow when I don't have past-midnight-brain, then merge

doc/luasnip.txt
```diff
@@ -270,11 +318,14 @@ lines rather than a string:
 <
 
 
+`t(text, node_opts)`: - `text`: `string` or `string[]` - `node_opts`: `table`,
```
bew, line 321: same here

doc/luasnip.txt
```diff
@@ -386,204 +436,111 @@ user-defined function:
...
-        }),
-        f(reused_func, {}, {
-            user_args = {"different text"}
-        }),
-    })
-<
-
-
-Examples: Use captures from the regex-trigger using a functionNode:
+`f(fn, argnode_references, node_opts)`: - `fn`: `function(argnode_text, parent,
```
bew, line 439: here as well

doc/luasnip.txt
```diff
@@ -386,204 +436,111 @@ user-defined function:
...
-                          }
-                        }
-            }
-<
-
+NODE_REFERENCE                                        *luasnip-node_reference*
+
+Node-references are used to refer to other nodes in various parts of
+luasnip’s API. For example, argnodes in functionNode, dynamicNode or lambda
+are node-references. These references can be either of: - `number`: the
```
bew, line 531: here also? for `these references can be either of ....`

doc/luasnip.txt
```diff
@@ -596,29 +553,27 @@ ChoiceNodes allow choosing between multiple nodes.
...
-    being edited also appears in the switched-to choice (can be the case if a
-    `restoreNode` is present in both choice) the cursor is restored relative to
-    that node.
-    The default is `false` as enabling might lead to worse performance. It’s
-    possible to override the default by wrapping the `choiceNode`-constructor
-    in another function that sets `opts.restore_cursor` to `true` and then using
-    that to construct `choiceNode`s:
-    `lua     local function restore_cursor_choice(pos, choices, opts)         if opts then             opts.restore_cursor = true         else             opts = {restore_cursor = true}         end         return c(pos, choices, opts)     end`
-
+`c(jump_index, choices, node_opts)` - |luasnip-`jump_index`|: `number`, since
```
bew, line 556: here (I think searching for ` - ` (a dash surrounded by spaces) can find you all inlined lists)

doc/luasnip.txt
```diff
@@ -964,120 +951,150 @@ to be accessed as `ai[restoreNodeIndx][0][1]`.
...
+|luasnip-`node-references`|, just like in function- and dynamicNode.
+
+There are many examples for `lamda` in `Examples/snippets.lua`
+
+MATCH                                                          *luasnip-match*
+
+`match` can insert text based on a predicate (again, a shorthand for
+`functionNode`).
+
+`match(argnodes, condition, then, else)`, where * `argnode`: A single
```
bew, line 974: another list, using `*` instead of `-`

L3MON4D3:
Mhmm, looks like the helptext-generator is messing those up.
Check `DOC.md`, there they actually are lists

L3MON4D3:
pushed a fix, 4b8d67817

L3MON4D3:
Thank you for bringing them to my attention, I very rarely check the help-files :sweat_smile:

PR #571, opened 2022-09-02
Add _CAPn and _TRIG environemnt vars

leiserfg:
- Add _CAPn and _TRIG environment variables
- Auto generate docs

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1399,10 +1399,13 @@ If `jsregexp` is not available, transformation are replaced by a simple copy.
 # VARIABLES
 
 All `TM_something`-variables are supported with two additions:
-`SELECT_RAW` and `SELECT_DEDENT`. These were introduced because
+`LS_SELECT_RAW` and `LS_SELECT_DEDENT`. These were introduced because
 `TM_SELECTED_TEXT` is designed to be compatible with vscodes' behavior, which
 can be counterintuitive when the snippet can be expanded at places other than
 the point where selection started (or when doing transformations on selected text).
+Besides those we also provide `LS_TRIGGER` which contains the trigger of the snippet,
+and  `LS_CAPTURE_n` (where n is a positive integer) that contains the n-th capture
+when using a regex with capture groups as `trig` in the snippet definition.
 
 All variables can be used outside of lsp-parsed snippets as their values are
 stored in a snippets' `snip.env`-table:
@@ -1410,7 +1413,7 @@ stored in a snippets' `snip.env`-table:
 s("selected_text", f(function(args, snip)
   local res, env = {}, snip.env
   table.insert(res, "Selected Text (current line is " .. env.TM_LINE_NUMBER .. "):")
-  for _, ele in ipairs(env.SELECT_RAW) do table.insert(res, ele) end
+  for _, ele in ipairs(env.LS_SELECT_RAW) do table.insert(res, ele) end
   return res
 end, {}))
 ```
@@ -1436,9 +1439,11 @@ You can also add your own variables by using the `ls.env_namespace(name, opts)` 
     Is a function that receives a string and returns a value for the var with that name
     or a table from var name to a value
     (in this case, if the value is a function it will be executed  lazily once per snippet expansion).
-    * `init`: `fn(pos: pair[int])->map[string, EnvVal]` Takes the 0-based position of the cursor and returns
+    * `init`: `fn(info: table)->map[string, EnvVal]`  Returns
         a table of variables that will set to the environment of the snippet on expansion,
         use this for vars that have to be calculated in that moment or that depend on each other.
+        The `info` table argument contains `pos` (0-based position of the cursor on expansion),
+        the `trigger` of the snippet and the `captures` list.
     * `eager`: `list[string]` names of variables that will be taken from `vars` and appended eagerly (like those in init)
     * `multiline_vars`: `(fn(name:string)->bool)|map[sting, bool]|bool|string[]` Says if certain vars are a table or just a string,
         can be a function that get's the name of the var and returns true if the var is a key,
@@ -1464,7 +1469,7 @@ ls.env_namespace("SYS", {vars=os.getenv, eager={"HOME"}})
 
 -- then you can use  $SYS_HOME which was eagerly initialized but also $SYS_USER (or any other system environment var) in your snippets
 
-lsp.env_namespace("POS", {init=function(pos) return {"VAL": vim.inspect(pos)}} end)
+lsp.env_namespace("POS", {init=function(info) return {"VAL": vim.inspect(info.pos)}} end)
 
 -- then you can use  $POS_VAL in your snippets
 
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*            For NVIM v0.5.0           Last change: 2022 August 30
+*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 02
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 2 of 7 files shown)


L3MON4D3:
Ah, nice :+1:

TBH, I'm not a big fan of calling the variables `_CAP` and `_TRIG`, IMO we should just take the breaking change (affecting probably nobody) for nicer variable names.

lua/luasnip/util/environ.lua
```diff
@@ -39,14 +39,14 @@ function Environ.is_table(var_fullname)
...
-function Environ:new(pos, o)
+function Environ:new(meta, o)
 	o = o or {}
 	setmetatable(o, self)
 
 	for ns_name, ns in pairs(namespaces) do
 		local eager_vars = {}
 		if ns.init then
-			eager_vars = ns.init(pos)
+			eager_vars = ns.init(meta)
```
L3MON4D3, line 49: Mhmm, what do you think of changing the table passed to `init` to
```lua
{
  pos = {row,col},
  trigger = ...
  captures = {...}
}
```
?
We could set `meta[1], meta[2]` (btw what's the mnemonic here? metadata?) to `row`, `col` to stay backward compatible

leiserfg:
> Ah, nice +1
> 
> TBH, I'm not a big fan of calling the variables `_CAP` and `_TRIG`, IMO we should just take the breaking change (affecting probably nobody) for nicer variable names.

this one is up to you

lua/luasnip/util/environ.lua
leiserfg: > btw what's the mnemonic here? metadata? 
It's either that or a reference to the effects of  Meta over the environment :interrobang: 




> We could set `meta[1], meta[2]`

Do you think then I should do that and keep `meta[1] & meta[2]` just as an undocumented non-breaking change protection?

L3MON4D3: > > btw what's the mnemonic here? metadata?
> 
> It's either that or a reference to the effects of Meta over the environment :interrobang:

:laughing:
Mhmm, what `data` is this the `metadata` to though?
(Just to be clear, I'm not worried about the name of the variable itself, more about what we call it in the documentation. I think `meta` is a bit confusing (+negatively connotated :P), just `info` maybe?)

> > We could set `meta[1], meta[2]`
> 
> Do you think then I should do that and keep `meta[1] & meta[2]` just as an undocumented non-breaking change protection?

Yes, undocumented :+1::+1:

L3MON4D3:
Hmmmm yeah, a namespace for our variables is a good idea, a breaking change for each variable we add would be annoying.

TBH, I'd prefer the longer versions `LS_CAPTUREn`, `LS_TRIGGER`, feels more in line with other variables.

I'm thinking; should we add older variables we introduced (`SELECT_RAW`, `SELECT_DEDENT`) to this namespace too?
Upside, it would be more coherent, downside, breaking change (and this one would actually affect ppl)
=> I think we shouldn't.

leiserfg:
> I'm thinking; should we add older variables we introduced (`SELECT_RAW`, `SELECT_DEDENT`) to this namespace too?

I will do both, so in the doc we can just talk about `LS` but we don't break old users' configs.

leiserfg:
> > > I'm thinking; should we add older variables we introduced (`SELECT_RAW`, `SELECT_DEDENT`) to this namespace too?
> > 
> > 
> > I will do both, so in the doc we can just talk about `LS` but we don't break old users' configs.
> 
> True, that sounds like the best solution, as long as it's not too complex to maintain (but doesn't seem to be) +1

In case you didn't notice, it's done already.

L3MON4D3:
Yup :+1:
Mhmmm I think it's okay to just leave lambdas implementation of `l.CAPTUREn` alone, it works, it'll continue to work.

Would you add tests for the new stuff passed to `init`, and `LS_`-variables?

Apart from that, this is pretty clean, nice job :+1::+1:

L3MON4D3:
Nice, looks good :+1:
Wanna squash them, 1 commit?

DOC.md
```diff
@@ -1436,9 +1436,11 @@ You can also add your own variables by using the `ls.env_namespace(name, opts)`
     Is a function that receives a string and returns a value for the var with that name
     or a table from var name to a value
     (in this case, if the value is a function it will be executed  lazily once per snippet expansion).
-    * `init`: `fn(pos: pair[int])->map[string, EnvVal]` Takes the 0-based position of the cursor and returns
+    * `init`: `fn(info: pair[int])->map[string, EnvVal]`  Returns
```
bew, line 1439: note: `info` is not really a `pair[int]` anymore 'officially'

tests/unit/environ_spec.lua
```diff
@@ -180,4 +187,26 @@ describe("luasnip.util.environ", function()
...
+						var_name,
+						test
+					)
+				)
+			)
+		end)
+	end
+
+	check_builtin("LS_TRIGGER", [[function(r) return r == "trigg" end]])
+	check_builtin("LS_CAPTURE1", [[function(r) return r == "one" end]])
```
bew, line 211: if we're going for long name, for capture group, it could be named `LS_CAPTURE_1` ?

leiserfg: [«Is mir egal»](https://www.youtube.com/watch?v=YEYim54pJ00) @L3MON4D3  should decide.

L3MON4D3: :D
I agree, it's easier to read :+1:

DOC.md
```diff
@@ -1399,7 +1399,7 @@ If `jsregexp` is not available, transformation are replaced by a simple copy.
 # VARIABLES
 
 All `TM_something`-variables are supported with two additions:
-`SELECT_RAW` and `SELECT_DEDENT`. These were introduced because
+`LS_SELECT_RAW` and `LS_SELECT_DEDENT`. These were introduced because
```
L3MON4D3, line 1402: Ah, could you add the new vars here too?
Best make it a list for readability

leiserfg: I think we should (not in this PR) list the TM_ ones too.

leiserfg: A link is ok for the online doc, but for the .txt I think it's better to have it in place :thinking:

L3MON4D3:
(I think test failure in d69e20b is just due to the testing-framework, seen it happen a few times)

doc/luasnip.txt
```diff
@@ -1345,6 +1345,9 @@ and `LS_SELECT_DEDENT`. These were introduced because `TM_SELECTED_TEXT` is
 designed to be compatible with vscodes’ behavior, which can be
 counterintuitive when the snippet can be expanded at places other than the
 point where selection started (or when doing transformations on selected text).
+Besides those we also provide `LS_TRIGGER` which contains the trigger of the
+snippet, and `LS_CAPTURE_n` (where n is a positive integer) that contains the
+n-th capture of the when using a regex as `trig` in the snippet definition.
```
bew, line 1350:
```suggestion
n-th capture when using a regex with capture groups as `trig` in the snippet definition.
```

leiserfg: So I can just use your review directly :smile:

leiserfg:
If everyone is happy now I will squash this.

leiserfg:
Squashed and re-named

L3MON4D3:
Ah, could you add a notice of the `LS`-breaking change to the commitmessage?

leiserfg:
> Ah, could you add a notice of the `LS`-breaking change to the commitmessage?

I will but if someone is using it it will appear anyway because of the asserts in environ.

DOC.md
L3MON4D3: Hmm yes, good point... I don't think `panvimdoc` allows .txt-specific sections though.
OTOH, just making a big list should be okay for the markdown as well, then we could even mention details of our implementation.

L3MON4D3:
True, but don't wanna be the bad guy if someone updates their config and it suddenly throws a big scary _red_ (:P) error

leiserfg:
> True, but don't wanna be the bad guy if someone updates their config and it suddenly throws a big scary _red_ (:P) error

You mean "again" :smile:

L3MON4D3:
Fine work @leiserfg :ok_hand:

PR #577, opened 2022-09-04
Add interface for applying snippetTextEdits

L3MON4D3:
This can be useful for `rust-analyzer` and other lsp-servers implementing their [extension](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#snippet-textedit).
Not 100% sure luasnip is really the best place for this, but it makes more sense than each languageserver-extension wanting to expand these snippets implementing it themselves.
(nvim also doesn't really fit since it's just an extension to lsp, and not official)

### DIFF
--- a/lua/luasnip/extras/lsp.lua
+++ b/lua/luasnip/extras/lsp.lua
@@ -1,0 +1,63 @@
+local luasnip_ns_id = require("luasnip.session").ns_id
+local ls = require("luasnip")
+
+local M = {}
+
+---Apply text/snippetTextEdits (at most one snippetText though).
+---@param snippet_or_text_edits `(snippetTextEdit|textEdit)[]`
+--- snippetTextEdit as defined in https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#snippet-textedit)
+---@param bufnr number, buffer where the snippet should be expanded.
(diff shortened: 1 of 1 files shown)


leiserfg:
Should not this also expose the companion lsp-capability that should be passed to the lsp-server?

L3MON4D3:
Oh, yeah that makes sense.
Is there some standard for how the capabilities are updated? Or would one have to call `update_capabilities(caps)` manually?

leiserfg:
> Oh, yeah that makes sense. Is there some standard for how the capabilities are updated? Or would one have to call `update_capabilities(caps)` manually?

Yes, it has to be manually, but it's something that those developing plugins that require it could do, like in this case the guys of  [rust-tools](https://github.com/simrat39/rust-tools.nvim).

leiserfg (approved):

L3MON4D3:
Thanks for taking a look, I had some more stuff to add though :D
Applying snippetTextEdits should work pretty well now, but we have some more complexity too:
There's a new option for `snip_expand`, `jumplist_insert_func`, which can link up snippets in a different way. Problem is, this is subject to some restrictions, and should be done carefully, so I'll have to add some pretty thorough documentation for it (or not document it at all, which would go a bit against having everything in `extras` not depend on internals)
In a similar vein, there's a lot of stuff that needs API for this to be really well-made, so far it's more of a POC

PR #578, opened 2022-09-04
Logging.

L3MON4D3:
Logging.

Current behaviour:
- will always append, even when a new session is started
  +: older logs still accessible
  -: log could grow too large
- can log either info,warn,debug,error. We'll probably want a way to turn off certain logging, based either on the severity or the source (I think turning off some sources could be pretty useful, or setting severities separately (`marks`="info", `loaders`="debug", something like that) otherwise I'd worry about drowning the user in uninteresting information (although, what do we have grep for :D)))

### DIFF
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -5,6 +5,7 @@ local str_util = require("luasnip.util.str")
 local loader_util = require("luasnip.loaders.util")
 local Path = require("luasnip.util.path")
 local sp = require("luasnip.nodes.snippetProxy")
+local log = require("luasnip.util.log").new("vscode-loader")
 
 local function json_decode(data)
 	local status, result = pcall(util.json_decode, data)
@@ -19,10 +20,15 @@ local function get_file_snippets(file)
 	local lang_snips = {}
 	local auto_lang_snips = {}
 
-	local data = Path.read_file(file)
+	local data_ok, data = pcall(Path.read_file, file)
+	if not data_ok then
+		log.error("Could not read file %s", file)
+		return {}, {}
+	end
 	local snippet_set_data = json_decode(data)
 	if snippet_set_data == nil then
-		return
+		log.error("Could not parse json in %s", file)
+		return {}, {}
 	end
 
 	for name, parts in pairs(snippet_set_data) do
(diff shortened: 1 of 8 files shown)


lua/luasnip/loaders/from_vscode.lua
```diff
@@ -113,15 +122,20 @@ end
 local function package_files(root, filter)
 	local package = Path.join(root, "package.json")
 	local data = Path.read_file(package)
+	-- if root doesn't contain a package.json, or it contributes no snippets,
+	-- return no snippets.
+	if not data then
+		log.warn("Tried reading package %s, but it does not exist", package)
+	end
```
bew, line 129: you forgot the `return {}` here?

L3MON4D3: Ah yes, ty!

lua/luasnip/util/log.lua
```diff
@@ -2,14 +2,19 @@
...
+	log_location,
 	-- only append.
 	"a",
 	-- 420 = 0644
 	420)
 
-
+local logsize = vim.loop.fs_fstat(luasnip_log_fd).size
+if logsize > 50*2^20 then
+	print("Luasnip's log now takes up more than 50MiB. Consider deleting it at " .. log_location)
```
L3MON4D3: Not sure about this, it could be a bit annoying..
Opinions?

leiserfg:
@L3MON4D3  had you tested it with two instances running in parallel? (I haven't just been curious about the concurrent appends).

L3MON4D3:
I think (hope, at least linux is capable of this afaik) the lines should stay intact, but they will be interleaved for sure.
I've considered prepending a unique identifier for the session to each message, but am not convinced it's really useful and doesn't just clutter up the log (I guess it might be useful for taking a look at older logs, then one could simply filter based on that identifier. Mhmmmm...)

lua/luasnip/util/log.lua
leiserfg: Maybe it could be rotated on startup? Like after 1M make a new one and rename the others?

L3MON4D3: Interesting.. maybe one main log, one old log, and when the main log reaches some threshold, delete old, copy main to old, and clear main?
That way we shouldn't delete any important older data, and sessions running concurrent with the rotation will also continue to write their logs.
+ we don't ever consume way too much storage
(had a nvim-dap-log with like 900Mb once)

PR #580, opened 2022-09-05
Implement semantic versioning.

L3MON4D3:
Hight time we do this :D

I'm thinking of doing releases regularly, using github's Release-feature for this seems to be the most straightforward way.

@leiserfg WDYT, any ideas to make this as easy as possible?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,20 @@ Neovim >= 0.5 (extmarks)
 
 # Setup
 ## Install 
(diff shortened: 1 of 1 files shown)


leiserfg:
Maybe something like this is helpful https://github.com/semantic-release/semantic-release

L3MON4D3:
(At least I think I'll start out with that, maybe it becomes annoying as time goes on :D)

gegoune:
There also is https://github.com/marketplace/actions/release-please-action.

L3MON4D3:
Oh, the release-PR seems like a nice way to keep track of changes.. I'll definitely keep that in mind :+1:

PR #597, opened 2022-09-14
feat: add new built-in condition for line ending

### LINKED ISSUE #592: Regular expressions with end-of-line match
Hey 👋🏾 

I'm currently trying to improve my snippets as some of them expand automatically too often when it doesn't fit. Therefore I want to check if there is nothing in the line after the trigger. In theory this should be a simple end-of-line match in the trigger. But that doesn't work. Adding this makes the snippet never expand.

Here is an example. The trigger I was using so far was ` if[%( ]`. So I simply added `$` at the end of the trigger (see below). That doesn't work. To simplify the problem I also tried ` if $`, but doesn't work neither.
Using `^` to match the start of the line works fine for me. No I'm unsure if this has something to do with how regular expressions are matched in the context of LuaSnip or if I don't understand a special case of the Lua expression syntax.

```lua
snippet(
  { trig = ' if[%( ]$', regTrig = true },
  if_node, -- ignore here to a simple example
}
```

Thanks for your help!

weilbith:
Closes #592

### DIFF
--- a/Examples/snippets.lua
+++ b/Examples/snippets.lua
@@ -326,6 +326,11 @@ ls.add_snippets("all", {
 	}, {
 		condition = conds.line_begin,
 	}),
+	s("cond3", {
+		t("will only expand at the end of the line"),
+	}, {
+		condition = conds.line_end,
+	}),
 	-- The last entry of args passed to the user-function is the surrounding snippet.
 	s(
 		{ trig = "a%d", regTrig = true },
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 08
+*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 14
 
 ==============================================================================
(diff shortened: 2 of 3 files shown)


L3MON4D3:
Hey, nice :+1:
TY!

atticus-sullivan:
Working on #612, I've looked at this code again. Is there any reason why you're just comparing the amount of characters in the string? (comparing strings in lua should be as cheap as a pointer comparison afaik)

L3MON4D3:
Mhmm, well number of characters works because both strings are the same line, so same lengths <=> same string.
I hadn't considered that just comparing them directly could be faster, but reading up on interning, that does seem to be the case (very interesting, I would've thought looking up the result after every string-operation leads to overall worse performance, but it seems to be more nuanced that that).
I'm definitely open to making this a direct comparison, best with a comment so it's clear that this might be/is faster than comparing lengths.

atticus-sullivan:
Alright, then I guess I'll do this change in #612 (or new PR?)

L3MON4D3:
Amend it, different commit, same PR is alright

PR #599, opened 2022-09-15
Hotfix nil in get snippets + adds test for error

### LINKED ISSUE #591: Error detected while processing TextChangedI Autocommands for "*"
Hi I just updated neovim and luasnip today, and it throws this error below everytime I start typing in a new line:

```
Error detected while processing TextChangedI Autocommands for "*":                                                                                                                               
Error executing lua callback: ...start/LuaSnip/lua/luasnip/session/snippet_collection.lua:278: bad argument #1 to 'ipairs' (table expected, got nil)                                             
stack traceback:                                                                                                                                                                                 
        [C]: in function 'ipairs'                                                                                                                                                                
        ...start/LuaSnip/lua/luasnip/session/snippet_collection.lua:278: in function 'get_snippets'                                                                                              
        ...nvim/site/pack/packer/start/LuaSnip/lua/luasnip/init.lua:51: in function 'get_snippets'                                                                                               
        ...e/pack/packer/start/cmp_luasnip/lua/cmp_luasnip/init.lua:74: in function 'complete'                                                                                                   
        .../nvim/site/pack/packer/start/nvim-cmp/lua/cmp/source.lua:311: in function 'complete'                                                                                                  
        ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:286: in function 'complete'                                                                                                  
        ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:168: in function 'callback'                                                                                                  
        ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:218: in function 'autoindent'                                                                                                
        ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:160: in function 'on_change'                                                                                                 
        ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/init.lua:310: in function 'callback'                                                                                                  
        ...ite/pack/packer/start/nvim-cmp/lua/cmp/utils/autocmd.lua:49: in function 'emit'                                                                                                       
        ...ite/pack/packer/start/nvim-cmp/lua/cmp/utils/autocmd.lua:23: in function <...ite/pack/packer/start/nvim-cmp/lua/cmp/utils/autocmd.lua:22>  
```
My neovim version:
```
:version                                                                                                                                                                                      
NVIM v0.8.0-dev-1071-gf98cff957                                                                                                                                                               
Build type: RelWithDebInfo                                                                                                                                                                    
LuaJIT 2.1.0-beta3                                                                                                                                                                            
Compilation: /usr/bin/cc -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-
prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-c
ommon -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/ub/programs/neovim/build/cmake.
config -I/home/ub/programs/neovim/src -I/home/ub/programs/neovim/.deps/usr/include -I/usr/include -I/home/ub/programs/neovim/build/src/nvim/auto -I/home/ub/programs/neovim/build/include     
Compiled by root@zen                                                                                                                                                                          
                                                                                                                                                                                              
Features: +acl +iconv +tui                                                                                                                                                                    
See ":help feature-compile"                                                                                                                                                                   
                                                                                                                                                                                              
   system vimrc file: "$VIM/sysinit.vim"                                                                                                                                                      
  fall-back for $VIM: "/usr/local/share/nvim"                                                                                                                                                 
                                                                                                                                                                                              
Run :checkhealth for more info    
```
My luasnip config file is the same as that in examples with some custom python snippets I have added (it still throws error if i delete my python snippets from config). Thanks in advance!

```
local ls = require("luasnip")
-- some shorthands...
local s = ls.snippet
local sn = ls.snippet_node
local t = ls.text_node
local i = ls.insert_node
local f = ls.function_node
local c = ls.choice_node
local d = ls.dynamic_node
local r = ls.restore_node
local l = require("luasnip.extras").lambda
local p = require("luasnip.extras").partial
local m = require("luasnip.extras").match
local n = require("luasnip.extras").nonempty
local dl = require("luasnip.extras").dynamic_lambda
local fmt = require("luasnip.extras.fmt").fmt
local fmta = require("luasnip.extras.fmt").fmta
local types = require("luasnip.util.types")
local conds = require("luasnip.extras.expand_conditions")

-- If you're reading this file for the first time, best skip to around line 190
-- where the actual snippet-definitions start.

-- Every unspecified option will be set to the default.
ls.setup({
	history = true,
	-- Update more often, :h events for more info.
	update_events = "TextChanged,TextChangedI",
	-- Snippets aren't automatically removed if their text is deleted.
	-- `delete_check_events` determines on which events (:h events) a check for
	-- deleted snippets is performed.
	-- This can be especially useful when `history` is enabled.
	delete_check_events = "TextChanged",
	ext_opts = {
		[types.choiceNode] = {
			active = {
				virt_text = { { "choiceNode", "Comment" } },
			},
		},
	},
	-- treesitter-hl has 100, use something higher (default is 200).
	ext_base_prio = 300,
	-- minimal increase in priority.
	ext_prio_increase = 1,
	enable_autosnippets = true,
	-- mapping for cutting selected text so it's usable as SELECT_DEDENT,
	-- SELECT_RAW or TM_SELECTED_TEXT (mapped via xmap).
	store_selection_keys = "<Tab>",
	-- luasnip uses this function to get the currently active filetype. This
	-- is the (rather uninteresting) default, but it's possible to use
	-- eg. treesitter for getting the current filetype by setting ft_func to
	-- require("luasnip.extras.filetype_functions").from_cursor (requires
	-- `nvim-treesitter/nvim-treesitter`). This allows correctly resolving
	-- the current filetype in eg. a markdown-code block or `vim.cmd()`.
	ft_func = function()
		return vim.split(vim.bo.filetype, ".", true)
	end,
})

-- args is a table, where 1 is the text in Placeholder 1, 2 the text in
-- placeholder 2,...
local function copy(args)
	return args[1]
end

-- 'recursive' dynamic snippet. Expands to some text followed by itself.
local rec_ls
rec_ls = function()
	return sn(
		nil,
		c(1, {
			-- Order is important, sn(...) first would cause infinite loop of expansion.
			t(""),
			sn(nil, { t({ "", "\t\\item " }), i(1), d(2, rec_ls, {}) }),
		})
	)
end

-- complicated function for dynamicNode.
local function jdocsnip(args, _, old_state)
	-- !!! old_state is used to preserve user-input here. DON'T DO IT THAT WAY!
	-- Using a restoreNode instead is much easier.
	-- View this only as an example on how old_state functions.
	local nodes = {
		t({ "/**", " * " }),
		i(1, "A short Description"),
		t({ "", "" }),
	}

	-- These will be merged with the snippet; that way, should the snippet be updated,
	-- some user input eg. text can be referred to in the new snippet.
	local param_nodes = {}

	if old_state then
		nodes[2] = i(1, old_state.descr:get_text())
	end
	param_nodes.descr = nodes[2]

	-- At least one param.
	if string.find(args[2][1], ", ") then
		vim.list_extend(nodes, { t({ " * ", "" }) })
	end

	local insert = 2
	for indx, arg in ipairs(vim.split(args[2][1], ", ", true)) do
		-- Get actual name parameter.
		arg = vim.split(arg, " ", true)[2]
		if arg then
			local inode
			-- if there was some text in this parameter, use it as static_text for this new snippet.
			if old_state and old_state[arg] then
				inode = i(insert, old_state["arg" .. arg]:get_text())
			else
				inode = i(insert)
			end
			vim.list_extend(
				nodes,
				{ t({ " * @param " .. arg .. " " }), inode, t({ "", "" }) }
			)
			param_nodes["arg" .. arg] = inode

			insert = insert + 1
		end
	end

	if args[1][1] ~= "void" then
		local inode
		if old_state and old_state.ret then
			inode = i(insert, old_state.ret:get_text())
		else
			inode = i(insert)
		end

		vim.list_extend(
			nodes,
			{ t({ " * ", " * @return " }), inode, t({ "", "" }) }
		)
		param_nodes.ret = inode
		insert = insert + 1
	end

	if vim.tbl_count(args[3]) ~= 1 then
		local exc = string.gsub(args[3][2], " throws ", "")
		local ins
		if old_state and old_state.ex then
			ins = i(insert, old_state.ex:get_text())
		else
			ins = i(insert)
		end
		vim.list_extend(
			nodes,
			{ t({ " * ", " * @throws " .. exc .. " " }), ins, t({ "", "" }) }
		)
		param_nodes.ex = ins
		insert = insert + 1
	end

	vim.list_extend(nodes, { t({ " */" }) })

	local snip = sn(nil, nodes)
	-- Error on attempting overwrite.
	snip.old_state = param_nodes
	return snip
end

-- Make sure to not pass an invalid command, as io.popen() may write over nvim-text.
local function bash(_, _, command)
	local file = io.popen(command, "r")
	local res = {}
	for line in file:lines() do
		table.insert(res, line)
	end
	return res
end

-- Returns a snippet_node wrapped around an insert_node whose initial
-- text value is set to the current date in the desired format.
local date_input = function(args, snip, old_state, fmt)
	local fmt = fmt or "%Y-%m-%d"
	return sn(nil, i(1, os.date(fmt)))
end

-- snippets are added via ls.add_snippets(filetype, snippets[, opts]), where
-- opts may specify the `type` of the snippets ("snippets" or "autosnippets",
-- for snippets that should expand directly after the trigger is typed).
--
-- opts can also specify a key. By passing an unique key to each add_snippets, it's possible to reload snippets by
-- re-`:luafile`ing the file in which they are defined (eg. this one).
ls.add_snippets("all", {
	-- trigger is `fn`, second argument to snippet-constructor are the nodes to insert into the buffer on expansion.
	s("fn", {
		-- Simple static text.
		t("//Parameters: "),
		-- function, first parameter is the function, second the Placeholders
		-- whose text it gets as input.
		f(copy, 2),
		t({ "", "function " }),
		-- Placeholder/Insert.
		i(1),
		t("("),
		-- Placeholder with initial text.
		i(2, "int foo"),
		-- Linebreak
		t({ ") {", "\t" }),
		-- Last Placeholder, exit Point of the snippet.
		i(0),
		t({ "", "}" }),
	}),
	s("class", {
		-- Choice: Switch between two different Nodes, first parameter is its position, second a list of nodes.
		c(1, {
			t("public "),
			t("private "),
		}),
		t("class "),
		i(2),
		t(" "),
		c(3, {
			t("{"),
			-- sn: Nested Snippet. Instead of a trigger, it has a position, just like insert-nodes. !!! These don't expect a 0-node!!!!
			-- Inside Choices, Nodes don't need a position as the choice node is the one being jumped to.
			sn(nil, {
				t("extends "),
				-- restoreNode: stores and restores nodes.
				-- pass position, store-key and nodes.
				r(1, "other_class", i(1)),
				t(" {"),
			}),
			sn(nil, {
				t("implements "),
				-- no need to define the nodes for a given key a second time.
				r(1, "other_class"),
				t(" {"),
			}),
		}),
		t({ "", "\t" }),
		i(0),
		t({ "", "}" }),
	}),
	-- Alternative printf-like notation for defining snippets. It uses format
	-- string with placeholders similar to the ones used with Python's .format().
	s(
		"fmt1",
		fmt("To {title} {} {}.", {
			i(2, "Name"),
			i(3, "Surname"),
			title = c(1, { t("Mr."), t("Ms.") }),
		})
	),
	-- To escape delimiters use double them, e.g. `{}` -> `{{}}`.
	-- Multi-line format strings by default have empty first/last line removed.
	-- Indent common to all lines is also removed. Use the third `opts` argument
	-- to control this behaviour.
	s(
		"fmt2",
		fmt(
			[[
		foo({1}, {3}) {{
			return {2} * {4}
		}}
		]],
			{
				i(1, "x"),
				rep(1),
				i(2, "y"),
				rep(2),
			}
		)
	),
	-- Empty placeholders are numbered automatically starting from 1 or the last
	-- value of a numbered placeholder. Named placeholders do not affect numbering.
	s(
		"fmt3",
		fmt("{} {a} {} {1} {}", {
			t("1"),
			t("2"),
			a = t("A"),
		})
	),
	-- The delimiters can be changed from the default `{}` to something else.
	s("fmt4", fmt("foo() { return []; }", i(1, "x"), { delimiters = "[]" })),
	-- `fmta` is a convenient wrapper that uses `<>` instead of `{}`.
	s("fmt5", fmta("foo() { return <>; }", i(1, "x"))),
	-- By default all args must be used. Use strict=false to disable the check
	s(
		"fmt6",
		fmt("use {} only", { t("this"), t("not this") }, { strict = false })
	),
	-- Use a dynamic_node to interpolate the output of a
	-- function (see date_input above) into the initial
	-- value of an insert_node.
	s("novel", {
		t("It was a dark and stormy night on "),
		d(1, date_input, {}, { user_args = { "%A, %B %d of %Y" } }),
		t(" and the clocks were striking thirteen."),
	}),
	-- Parsing snippets: First parameter: Snippet-Trigger, Second: Snippet body.
	-- Placeholders are parsed into choices with 1. the placeholder text(as a snippet) and 2. an empty string.
	-- This means they are not SELECTed like in other editors/Snippet engines.
	ls.parser.parse_snippet(
		"lspsyn",
		"Wow! This ${1:Stuff} really ${2:works. ${3:Well, a bit.}}"
	),

	-- When wordTrig is set to false, snippets may also expand inside other words.
	ls.parser.parse_snippet(
		{ trig = "te", wordTrig = false },
		"${1:cond} ? ${2:true} : ${3:false}"
	),

	-- When regTrig is set, trig is treated like a pattern, this snippet will expand after any number.
	ls.parser.parse_snippet({ trig = "%d", regTrig = true }, "A Number!!"),
	-- Using the condition, it's possible to allow expansion only in specific cases.
	s("cond", {
		t("will only expand in c-style comments"),
	}, {
		condition = function(line_to_cursor, matched_trigger, captures)
			-- optional whitespace followed by //
			return line_to_cursor:match("%s*//")
		end,
	}),
	-- there's some built-in conditions in "luasnip.extras.expand_conditions".
	s("cond2", {
		t("will only expand at the beginning of the line"),
	}, {
		condition = conds.line_begin,
	}),
	-- The last entry of args passed to the user-function is the surrounding snippet.
	s(
		{ trig = "a%d", regTrig = true },
		f(function(_, snip)
			return "Triggered with " .. snip.trigger .. "."
		end, {})
	),
	-- It's possible to use capture-groups inside regex-triggers.
	s(
		{ trig = "b(%d)", regTrig = true },
		f(function(_, snip)
			return "Captured Text: " .. snip.captures[1] .. "."
		end, {})
	),
	s({ trig = "c(%d+)", regTrig = true }, {
		t("will only expand for even numbers"),
	}, {
		condition = function(line_to_cursor, matched_trigger, captures)
			return tonumber(captures[1]) % 2 == 0
		end,
	}),
	-- Use a function to execute any shell command and print its text.
	s("bash", f(bash, {}, { user_args = { "ls" } })),
	-- Short version for applying String transformations using function nodes.
	s("transform", {
		i(1, "initial text"),
		t({ "", "" }),
		-- lambda nodes accept an l._1,2,3,4,5, which in turn accept any string transformations.
		-- This list will be applied in order to the first node given in the second argument.
		l(l._1:match("[^i]*$"):gsub("i", "o"):gsub(" ", "_"):upper(), 1),
	}),

	s("transform2", {
		i(1, "initial text"),
		t("::"),
		i(2, "replacement for e"),
		t({ "", "" }),
		-- Lambdas can also apply transforms USING the text of other nodes:
		l(l._1:gsub("e", l._2), { 1, 2 }),
	}),
	s({ trig = "trafo(%d+)", regTrig = true }, {
		-- env-variables and captures can also be used:
		l(l.CAPTURE1:gsub("1", l.TM_FILENAME), {}),
	}),
	-- Set store_selection_keys = "<Tab>" (for example) in your
	-- luasnip.config.setup() call to populate
	-- TM_SELECTED_TEXT/SELECT_RAW/SELECT_DEDENT.
	-- In this case: select a URL, hit Tab, then expand this snippet.
	s("link_url", {
		t('<a href="'),
		f(function(_, snip)
			-- TM_SELECTED_TEXT is a table to account for multiline-selections.
			-- In this case only the first line is inserted.
			return snip.env.TM_SELECTED_TEXT[1] or {}
		end, {}),
		t('">'),
		i(1),
		t("</a>"),
		i(0),
	}),
	-- Shorthand for repeating the text in a given node.
	s("repeat", { i(1, "text"), t({ "", "" }), rep(1) }),
	-- Directly insert the ouput from a function evaluated at runtime.
	s("part", p(os.date, "%Y")),
	-- use matchNodes (`m(argnode, condition, then, else)`) to insert text
	-- based on a pattern/function/lambda-evaluation.
	-- It's basically a shortcut for simple functionNodes:
	s("mat", {
		i(1, { "sample_text" }),
		t(": "),
		m(1, "%d", "contains a number", "no number :("),
	}),
	-- The `then`-text defaults to the first capture group/the entire
	-- match if there are none.
	s("mat2", {
		i(1, { "sample_text" }),
		t(": "),
		m(1, "[abc][abc][abc]"),
	}),
	-- It is even possible to apply gsubs' or other transformations
	-- before matching.
	s("mat3", {
		i(1, { "sample_text" }),
		t(": "),
		m(
			1,
			l._1:gsub("[123]", ""):match("%d"),
			"contains a number that isn't 1, 2 or 3!"
		),
	}),
	-- `match` also accepts a function in place of the condition, which in
	-- turn accepts the usual functionNode-args.
	-- The condition is considered true if the function returns any
	-- non-nil/false-value.
	-- If that value is a string, it is used as the `if`-text if no if is explicitly given.
	s("mat4", {
		i(1, { "sample_text" }),
		t(": "),
		m(1, function(args)
			-- args is a table of multiline-strings (as usual).
			return (#args[1][1] % 2 == 0 and args[1]) or nil
		end),
	}),
	-- The nonempty-node inserts text depending on whether the arg-node is
	-- empty.
	s("nempty", {
		i(1, "sample_text"),
		n(1, "i(1) is not empty!"),
	}),
	-- dynamic lambdas work exactly like regular lambdas, except that they
	-- don't return a textNode, but a dynamicNode containing one insertNode.
	-- This makes it easier to dynamically set preset-text for insertNodes.
	s("dl1", {
		i(1, "sample_text"),
		t({ ":", "" }),
		dl(2, l._1, 1),
	}),
	-- Obviously, it's also possible to apply transformations, just like lambdas.
	s("dl2", {
		i(1, "sample_text"),
		i(2, "sample_text_2"),
		t({ "", "" }),
		dl(3, l._1:gsub("\n", " linebreak ") .. l._2, { 1, 2 }),
	}),
}, {
	key = "all",
})


  ls.add_snippets("python" ,
  {
      s({
        trig="mark",
        name="Markdown-Jupyter",
        dscr="Add markdown field"
    },
      {
     t({"######################################################################################","# %% [markdown]","## "}), i(1),
    	t({"","# %%",""}), i(0),
    } ),
      s({
        trig="fr",
        name="Import"
    },
      { t("from "),
        i(1),t({" import "}),
        i(0)

      }
    ),
    s(
      {
        trig="'''",
        name="docstring",
      },
      {
        t({"'''",""}),i(1),
        t({"","'''",""}),i(0)
      }
    ),
    s(
      {
        trig = "ipdb",
        name= "Set trace"
      },
      {
        t({"import ipdb","tr = ipdb.set_trace",""}),
      }
    ),
    s(
    {
        trig = "autoreload",
        name = "default variables"
      },
    {

    t(
        {

          "if 'get_ipython' in globals():",
          "\timport os; os.chdir('/home/ub/Dropbox/code/unet_segmentation')",
          "\tprint('setting autoreload')",
          "\tfrom IPython import get_ipython",
          "\tipython = get_ipython()",
          "\tipython.magic('load_ext autoreload')",
          "\tipython.magic('autoreload 2')",
          }
        ),
      }
    )
    },
 {
	key = "python",
}
)



ls.add_snippets("java", {
	-- Very long example for a java class.
	s("fn", {
		d(6, jdocsnip, { 2, 4, 5 }),
		t({ "", "" }),
		c(1, {
			t("public "),
			t("private "),
		}),
		c(2, {
			t("void"),
			t("String"),
			t("char"),
			t("int"),
			t("double"),
			t("boolean"),
			i(nil, ""),
		}),
		t(" "),
		i(3, "myFunc"),
		t("("),
		i(4),
		t(")"),
		c(5, {
			t(""),
			sn(nil, {
				t({ "", " throws " }),
				i(1),
			}),
		}),
		t({ " {", "\t" }),
		i(0),
		t({ "", "}" }),
	}),
}, {
	key = "java",
})

ls.add_snippets("tex", {
	-- rec_ls is self-referencing. That makes this snippet 'infinite' eg. have as many
	-- \item as necessary by utilizing a choiceNode.
	s("ls", {
		t({ "\\begin{itemize}", "\t\\item " }),
		i(1),
		d(2, rec_ls, {}),
		t({ "", "\\end{itemize}" }),
	}),
}, {
	key = "tex",
})

-- set type to "autosnippets" for adding autotriggered snippets.
ls.add_snippets("all", {
	s("autotrigger", {
		t("autosnippet"),
	}),
}, {
	type = "autosnippets",
	key = "all_auto",
})

-- in a lua file: search lua-, then c-, then all-snippets.
ls.filetype_extend("lua", { "c" })
-- in a cpp file: search c-snippets, then all-snippets only (no cpp-snippets!!).
ls.filetype_set("cpp", { "c" })

-- Beside defining your own snippets you can also load snippets from "vscode-like" packages
-- that expose snippets in json files, for example <https://github.com/rafamadriz/friendly-snippets>.

require("luasnip.loaders.from_vscode").load({ include = { "python" } }) -- Load only python snippets

-- The directories will have to be structured like eg. <https://github.com/rafamadriz/friendly-snippets> (include
-- a similar `package.json`)
require("luasnip.loaders.from_vscode").load({ paths = { "./my-snippets" } }) -- Load snippets from my-snippets folder

-- You can also use lazy loading so snippets are loaded on-demand, not all at once (may interfere with lazy-loading luasnip itself).
require("luasnip.loaders.from_vscode").lazy_load() -- You can pass { paths = "./my-snippets/"} as well

-- You can also use snippets in snipmate format, for example <https://github.com/honza/vim-snippets>.
-- The usage is similar to vscode.

-- One peculiarity of honza/vim-snippets is that the file containing global
-- snippets is _.snippets, so we need to tell luasnip that the filetype "_"
-- contains global snippets:
ls.filetype_extend("all", { "_" })

-- see DOC.md/LUA SNIPPETS LOADER for some details.
require("luasnip.loaders.from_lua").load({ include = { "c" } })
require("luasnip.loaders.from_lua").lazy_load({ include = { "all", "cpp" } })
```

atticus-sullivan:
Fixes #591. We've forgotten to add the metatable once again in `clean_invalidated` (should be the last occurrence where it was missing, checked (hopefully) all the other ones)

Please mention if you've got another idea how to test this by not simply checking for a thrown error.

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 14
+*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 15
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
--- a/lua/luasnip/session/snippet_collection.lua
+++ b/lua/luasnip/session/snippet_collection.lua
@@ -217,6 +217,7 @@ function M.clean_invalidated(opts)
 
 	for type, type_snippets in pairs(by_ft) do
 		by_ft[type] = without_invalidated(type_snippets)
+		setmetatable(by_ft[type], by_ft_snippets_mt)
 	end
 
 	for key, key_snippets in pairs(by_key) do
--- a/tests/unit/snippet_collection_spec.lua
+++ b/tests/unit/snippet_collection_spec.lua
@@ -54,7 +54,7 @@ describe("add_snippets invalidation", function()
 
 	it("", function()
 		local function foo()
-				return helpers.exec_lua([[
+			return helpers.exec_lua([[
 					local s,t = require("luasnip").snippet, require("luasnip").text_node
 					local collection = require("luasnip.session.snippet_collection")
 					collection.clear_snippets()
--- a/tests/unit/snippet_collection_spec.lua
+++ b/tests/unit/snippet_collection_spec.lua
@@ -46,3 +46,25 @@ describe("snippet_collection.add/get", function()
 		assert.is_true(foo())
 	end)
 end)
+
+describe("add_snippets invalidation", function()
+	-- apparently clear() needs to run before anything else...
+	helpers.clear()
+	helpers.exec("set rtp+=" .. os.getenv("LUASNIP_SOURCE"))
+
+	it("", function()
+		local function foo()
+				return helpers.exec_lua([[
+					local s,t = require("luasnip").snippet, require("luasnip").text_node
+					local collection = require("luasnip.session.snippet_collection")
+					collection.clear_snippets()
+					local s1,s2 = s("trig1", t("snippet1")), s("trig2", t("snippet2"))
+					local opts = {type="snippets", default_priority=1000, key="abc"}
+					collection.add_snippets({["txt"]={s1,s2}}, opts)
+					collection.clean_invalidated({ inv_limit = -1 })
+					local r = collection.get_snippets("nonExistantFT", "snippets")
+					]])
+		end
+		assert.has_no.errors(foo)
+	end)
+end)


L3MON4D3:
Nice, that was quick :+1:

About the test: you can call `clean_invalidated` from the outside, and even set the number of snippets needed to actually trigger it (for the test, just 0 I guess)
Checking for an error seems just fine :+1:

atticus-sullivan:
Ah right, had my eyes only on `without_invalidated` ^^. calling `clean_invalidated` directly is much better as this way we don't get any trouble if we decide to change the threshold in `invalidate_snippets`.

atticus-sullivan:
Should be done.

> Nice, that was quick :+1:

Well, I was already on it, and then things are usually quick (and didn't wanted to keep people waiting ^^)

PR #603, opened 2022-09-17
Remove \r from strings before parsing

L3MON4D3:
I guess this is the correct thing to do here, seems like jdtls also separates (not terminates) lines with `\r\n` (at least).
(also means we can't re-use the function for reading an entire file from `loaders.util`, where line-terminators are assumed).


L3MON4D3:
merged via cherry-pick.

PR #604, opened 2022-09-19
fix(snipmate): correctly escape backtick

danilshvalov:
Let's have the following snippet:
```snippets
snippet src
	\`\`\`$1
	$0
	\`\`\`
```

Without this patch, it will expand into:
````
\`\`\`

\`\`\`
````

After this fix, it will expand into what is expected:
````
```

```
````

### DIFF
--- a/lua/luasnip/util/parser/init.lua
+++ b/lua/luasnip/util/parser/init.lua
@@ -116,7 +116,7 @@ local function backticks_to_variable(body)
 	end
 
 	-- append remaining characters.
-	var_string = var_string .. body:sub(processed_to, -1)
+	var_string = var_string .. body:sub(processed_to, -1):gsub("\\`", "`")
 
 	return var_map, var_string
 end
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 18
+*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 20
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: lua/luasnip/util/parser/init.lua [hunk 1]
```diff
@@ -117,6 +117,7 @@ local function backticks_to_variable(body)
 
 	-- append remaining characters.
 	var_string = var_string .. body:sub(processed_to, -1)
+	var_string = var_string:gsub("\\`", "`")
```
L3MON4D3, line 120: I think we'd only want to `gsub` occurences in the remaining characters, the escaped ``` ` ``` inside unescaped backticks are already replaced once.
(``` `echo '\\\` lel'` ``` would become ``` echo '`lel' ```, instead of ``` echo '\`lel' ``` (I think so at least, please verify :sweat_smile:))

So basically, `body:sub(...):gsub("\\`", "`")` should work instead.

L3MON4D3:
Thank you for making a PR for this :)

Are you up for adding a test to ensure this doesn't regress?
You should be able to run current test via just `make test`, and can append a new test for this in `tests/integration/parser_spec.lua`

lua/luasnip/util/parser/init.lua
danilshvalov: Done

L3MON4D3:
Nice :)
One more test for the edge-case I mentioned? (`` \\\` `` inside regular backticks)
(And could you squash the commits after?)

danilshvalov:
> Nice :) One more test for the edge-case I mentioned? (`` \\\` `` inside regular backticks) (And could you squash the commits after?)

Done

L3MON4D3:
Yup, ty!

PR #609, opened 2022-09-28
doc: fix tiny typo its' -> its

ejmastnak:
Completely trivial, but thought I'd fix it. Thanks for the great plugin!


L3MON4D3:
Ah, I'll never get used to those, damn exceptions :P
Thanks for the fix!

ejmastnak:
Haha yeah its/it's/its' can be awful!

PR #610, opened 2022-09-28
Return a better error when from_lua loader fails

robertgzr:
Prior to this patch a erroneous snippet file would lead to an error
along the lines of: `attempting to call a nil value`
With this the error clearly points to the faulty file.

Signed-off-by: Robert Günzler <r@gnzler.io>


L3MON4D3:
Makes sense to provide this info, thank you :+1:

PR #612, opened 2022-10-01
[feature] implement condition objects

atticus-sullivan:
Refer to #600 for most details. As it turned out that memoization is pretty hard to achive (moreover the invalidation) and we're not sure if it is worth it, we decided to do the basic condition objects (and -> `*` , or -> `+`) in a separate PR.

TODOs:
- [x] Write some tests for this
- [x] Write documentation
- [x] Move to `extras/conditions/` / split into `_expand`/`_show` and `_objects`

Anything I'm missing?

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1064,6 +1064,31 @@ ls.add_snippets("all", {
 })
 ```
 
+- `conditions.show`: Contains typical predicates/functions used as
+  `show`-condition. Currently this is just `line_end`
+- `conditions.expand`: Contains typical predicates/functions used as
+  `expand`-condition. Currently this is just `line_begin`
+  Contains everything from `conditions.show` as well.
+- `conditions`: Provides a function `make_condition(foo)` which takes a function
+  as argument and returns a *condition object* for which several operators are
+  defined:
+  - `c1 + c2 -> c1 or c2`
+  - `c1 * c2 -> c1 and c2`
+  - `-c1 -> not c1`
+  - `c1 ^ c2 -> c1 xor/!= c2`
+  - `c1 % c2 -> c1 xnor/== c2`: This decision may look weird but as we weren't
+	able to use `==`, we decided to take something that makes one scratch ones
+	head (and thus avoid making false assumptions).
+	For more details look at [this comment](https://github.com/L3MON4D3/LuaSnip/pull/612#issuecomment-1264487743).
+
+  `conditions.show`s and `conditions.expand`s members all are also condition
+  objects so you can work with those too.
+
+  Thus you can easily combine existing predicates. Like in
+  `conditions.expand.line_end + conditions.expand.line_begin` instead of doing
+  something like
+  `function(...) return conditions.expand.line_end(...) or conditions.expand.line_begin(...) end`.
+
 <!-- panvimdoc-ignore-start -->
 
 extras1: ![extras1](https://user-images.githubusercontent.com/25300418/184359431-50f90599-3db0-4df0-a3a9-27013e663649.gif)
--- a/lua/luasnip/extras/conditions/expand.lua
+++ b/lua/luasnip/extras/conditions/expand.lua
@@ -1,0 +1,14 @@
+local cond_obj = require("luasnip.extras.conditions")
+
+-- use the functions from show as basis and extend/overwrite functions specific for expand here
+local M = vim.deepcopy(require("luasnip.extras.conditions.show"))
+-----------------------
+-- PRESET CONDITIONS --
+-----------------------
+local function line_begin(line_to_cursor, matched_trigger)
+	-- +1 because `string.sub("abcd", 1, -2)` -> abc
+	return line_to_cursor:sub(1, -(#matched_trigger + 1)):match("^%s*$")
+end
+M.line_begin = cond_obj.make_condition(line_begin)
+
+return M
--- a/lua/luasnip/extras/expand_conditions.lua
+++ b/lua/luasnip/extras/expand_conditions.lua
@@ -1,13 +1,1 @@
-local M = {}
-
-function M.line_begin(line_to_cursor, matched_trigger)
-	-- +1 because `string.sub("abcd", 1, -2)` -> abc
-	return line_to_cursor:sub(1, -(#matched_trigger + 1)):match("^%s*$")
-end
-
-function M.line_end(line_to_cursor)
-	local line = vim.api.nvim_get_current_line()
-	return #line_to_cursor == #line
-end
-
-return M
+return require("luasnip.extras.conditions.expand")
--- a/Examples/snippets.lua
+++ b/Examples/snippets.lua
@@ -17,7 +17,8 @@ local dl = require("luasnip.extras").dynamic_lambda
 local fmt = require("luasnip.extras.fmt").fmt
 local fmta = require("luasnip.extras.fmt").fmta
 local types = require("luasnip.util.types")
-local conds = require("luasnip.extras.expand_conditions")
+local conds = require("luasnip.extras.conditions")
+local conds_expand = require("luasnip.extras.conditions.expand")
 
 -- If you're reading this file for the first time, best skip to around line 190
 -- where the actual snippet-definitions start.
@@ -320,16 +321,27 @@ ls.add_snippets("all", {
 			return line_to_cursor:match("%s*//")
 		end,
 	}),
-	-- there's some built-in conditions in "luasnip.extras.expand_conditions".
+	-- there's some built-in conditions in "luasnip.extras.conditions.expand" and "luasnip.extras.conditions.show".
 	s("cond2", {
 		t("will only expand at the beginning of the line"),
 	}, {
-		condition = conds.line_begin,
+		condition = conds_expand.line_begin,
 	}),
 	s("cond3", {
 		t("will only expand at the end of the line"),
 	}, {
-		condition = conds.line_end,
+		condition = conds_expand.line_end,
+	}),
+	-- on conditions some logic operators are defined
+	s("cond4", {
+		t("will only expand at the end and the start of the line"),
+	}, {
+		-- last function is just an example how to make own function objects and apply operators on them
+		condition = conds_expand.line_end
+			+ conds_expand.line_begin
+				* conds.make_condition(function()
+					return true
+				end),
 	}),
 	-- The last entry of args passed to the user-function is the surrounding snippet.
 	s(
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1050,9 +1050,9 @@ is only a short outline, their usage is shown more expansively in
 
 
 
-- `conditions.show`: Contains typical predicats/functions used as
+- `conditions.show`: Contains typical predicates/functions used as
     `show`-condition. Currently this is just `line_end`
-- `conditions.expand`: Contains typical predicats/functions used as
+- `conditions.expand`: Contains typical predicates/functions used as
     `expand`-condition. Currently this is just `line_begin` Contains everything
     from `conditions.show` as well.
 - `conditions`: Provides a function `make_condition(foo)` which takes a function
@@ -1068,7 +1068,7 @@ is only a short outline, their usage is shown more expansively in
         For more details look at this comment <https://github.com/L3MON4D3/LuaSnip/pull/612#issuecomment-1264487743>.
     `conditions.show`s and `conditions.expand`s members all are also condition
     objects so you can work with those too.
-    Thus you can easily combine existing predicats. Like in
+    Thus you can easily combine existing predicates. Like in
     `conditions.expand.line_end + conditions.expand.line_begin` instead of doing
     something like `function(...) return conditions.expand.line_end(...) or
     conditions.expand.line_begin(...) end`.
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*          For NVIM v0.5.0          Last change: 2022 September 28
+*luasnip.txt*           For NVIM v0.5.0           Last change: 2022 October 08
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
@@ -1047,6 +1047,31 @@ is only a short outline, their usage is shown more expansively in
       s("extras7", { i(1), t { "", "" }, extras.dynamic_lambda(2, l._1 .. l._1, 1) }),
     })
 <
+
+
+
+- `conditions.show`: Contains typical predicats/functions used as
+    `show`-condition. Currently this is just `line_end`
+- `conditions.expand`: Contains typical predicats/functions used as
+    `expand`-condition. Currently this is just `line_begin` Contains everything
+    from `conditions.show` as well.
+- `conditions`: Provides a function `make_condition(foo)` which takes a function
+    as argument and returns a _condition object_ for which several operators are
+    defined:
+    - `c1 + c2 -> c1 or c2`
+    - `c1 * c2 -> c1 and c2`
+    - `-c1 -> not c1`
+    - `c1 ^ c2 -> c1 xor/!= c2`
+    - `c1 % c2 -> c1 xnor/== c2`: This decision may look weird but as we weren’t
+        able to use `==`, we decided to take something that makes one scratch ones
+        head (and thus avoid making false assumptions).
+        For more details look at this comment <https://github.com/L3MON4D3/LuaSnip/pull/612#issuecomment-1264487743>.
+    `conditions.show`s and `conditions.expand`s members all are also condition
+    objects so you can work with those too.
+    Thus you can easily combine existing predicats. Like in
+    `conditions.expand.line_end + conditions.expand.line_begin` instead of doing
+    something like `function(...) return conditions.expand.line_end(...) or
+    conditions.expand.line_begin(...) end`.
 
 
 FMT                                                              *luasnip-fmt*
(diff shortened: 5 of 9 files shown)


L3MON4D3:
Nice, I think that's it, not missing anything :+1:

lua/luasnip/extras/expand_conditions.lua
```diff
@@ -1,13 +1,62 @@
...
+		return function(...)
+			return o1(...) == o2(...)
+		end
+	end,
+	-- use table like a function by overloading __call
+	__call = function(tab, line_to_cursor, matched_trigger, captures)
+		return tab.func(line_to_cursor, matched_trigger, captures)
+	end,
+}
+function M.condition_factory(func)
```
L3MON4D3, line 43: WDYT of `make_condition` instead? Feels a bit more descriptive

atticus-sullivan:
[Lua 5.4 reference](https://www.lua.org/manual/5.4/manual.html#2.4) states that for `__eq`
> The result of the call is always converted to a boolean.

The for luajit more relevant [Lua 5.1 reference](https://www.lua.org/manual/5.1/manual.html#2.8) contains no such statement (lua5.2 states `Note that the result is always a boolean.`).

In my tests this also holds. Lua seems to be doing something like `return self.__eq(other) ~= nil and self.__eq(other) ~= false`.
As we need to return a function from the operator, this is a problem for us.

In effect we cannot use `==`, `<` and `>`. Operators which are left are
- `__sub` (`-` bad Idea as we're using the `__unm` for `not`)
- `__div` (`/`)
- `__mod` (`%`)
- `__concat` (`..`)
- `__len` (`#`)

Nothing seems to be really fitting for the `xnor` aka `==` operator. Any ideas?

atticus-sullivan:
Should we add a composite test of multiple operators (including parentheses)? And if so any suggestions what expression might be interesting?

lua/luasnip/extras/expand_conditions.lua
atticus-sullivan: Yes, `factory` is comprehensive if you've been working with something like this before, but `make_condition` should be better in general.

lua/luasnip/extras/expand_conditions.lua
```diff
@@ -1,13 +1,62 @@
...
+-----------------------
+local condition_mt = {
+	-- logic operators
+	-- not '-'
+	__unm = function(o1)
+		return function(...)
+			return not o1(...)
+		end
+	end,
+	-- or '+'
```
bew, line 14: `+` not meaning `and` feels weird to me.

For `or` I think `/` (`__div`) feels more natural: `foo / bar` (foo or bar)
Following this, what about `//` (`__idiv`) for xor?

bew (commented):
Also in the spec, when you require stuff from `luasnip.extras.expand_conditions`, the local var is currently always called `cond`, I think it'd be more readable to have:
- `mkcond` when importing `make_condition`
- `conds` for the module `luasnip.extras.expand_conditions`

L3MON4D3:
> [Lua 5.4 reference](https://www.lua.org/manual/5.4/manual.html#2.4) states that for `__eq`
> 
> > The result of the call is always converted to a boolean.
> 
> The for luajit more relevant [Lua 5.1 reference](https://www.lua.org/manual/5.1/manual.html#2.8) contains no such statement (lua5.2 states `Note that the result is always a boolean.`).
> 
> In my tests this also holds. Lua seems to be doing something like `return self.__eq(other) ~= nil and self.__eq(other) ~= false`. As we need to return a function from the operator, this is a problem for us.
> 
> In effect we cannot use `==`, `<` and `>`. Operators which are left are
> 
> * `__sub` (`-` bad Idea as we're using the `__unm` for `not`)
> * `__div` (`/`)
> * `__mod` (`%`)
> * `__concat` (`..`)
> * `__len` (`#`)
> 
> Nothing seems to be really fitting for the `xnor` aka `==` operator. Any ideas?

Oh wow, that's unfortunate :/
Tough, I'm not sure whether a confusing symbol is better than not having one.. == is also quickly achieved via `-(a^b)`, but having a direct operator for it would be better.
Of the above, I think `%` would be best. It feels the most unusual and hopefully triggers a "Okay no idea what this is supposed to mean, let's have a look at the docs", instead of an "Ahh, this probably means ___, alright"

> Also in the spec, when you require stuff from `luasnip.extras.expand_conditions`, the local var is currently always called `cond`, I think it'd be more readable to have:
> 
> * `mkcond` when importing `make_condition`
> * `conds` for the module `luasnip.extras.expand_conditions`

Ah, `expand_conditions` used to be named `conditions`, but when `show_conditions` was introduced, this lead to some ambiguity, and was changed.
Putting the condition-objects into their own module sounds good though👍

One important fact: `show_condition`s can be passed to `expand_condition`, so we could have
* `extras/show_conditions` (which contains the conditions  we currently have (they only rely on `line_to_cursor`))
* `extras/expand_condition`, which re-exports all `show_conditions` (passing a `show_condition` to `(expand_)condition` seems fishy, and the separation could prevent a breaking change if `show_condition`s have to be incompatible with `expand_condition` for whatever reason)
* `extras/condition` (not sure about that name tbh, `condition_objects` is an alternative), which is relied on by both modules, and implements the boolean expression-functionality

Would be a bit nicer if all were in one common directory, so `extras/conditions/{init,show,expand}`, but that would be a breaking change (although, I'm not completely averse to making that one, just have to release 1.1.0 before going straight to 2.0.0 xD)

lua/luasnip/extras/expand_conditions.lua
L3MON4D3: Mhmm `+` for `or` and `*` for `and` are pretty common on the mathematical side of CS, I'd stick to them, fewer surprises. (they do feel weird in the beginning, but one gets used to them :D)
`^` is commonly used for bitwise xor, should keep that one too.

leiserfg:
For XOR I think `__pow` is the best option (it even matches the mathematical notation)

L3MON4D3:
> Should we add a composite test of multiple operators (including parentheses)? And if so any suggestions what expression might be interesting?

Ah tests, right: One fun way to test them would be running the same expression for all true/false combinations of their variables, and comparing that to the expected results (if it gets too complex, don't do it though :D)
But I don't have any specific expressions in mind, just testing every operator (and maybe composability (negation is quickly implement if you end up using the truth-tables ;) )) once should do it.

atticus-sullivan:
> Ah tests, right: One fun way to test them would be running the same expression for all true/false combinations of their variables, and comparing that to the expected results (if it gets too complex, don't do it though :D)

That's what I'm currently doing for the single operators.

> But I don't have any specific expressions in mind, just testing every operator (and maybe composability (negation is quickly implement if you end up using the truth-tables ;) )) once should do it.

Hm then I'll have to think of an expression.

Well just thought of composite ones and there is work to do (currently these opterators simply return functions -> e.g, `(f1+f2)*(f3+f4)` shouldn't work (`*` not defined for functions) so I'll have to wrap those in tables as well)
[always good to have (or at least think about) tests xD]

atticus-sullivan:
> Tough, I'm not sure whether a confusing symbol is better than not having one.. == is also quickly achieved via -(a^b), but having a direct operator for it would be better.
> Of the above, I think `%` would be best. It feels the most unusual and hopefully triggers a "Okay no idea what this is supposed to mean, let's have a look at the docs", instead of an "Ahh, this probably means ___, alright"

I agree. But I'm not sure as well whether not implement `==` or use `%` (well no one foces one to use `%` though :thinking:)

atticus-sullivan:
> Would be a bit nicer if all were in one common directory, so extras/conditions/{init,show,expand}, but that would be a breaking change

I'd be in favor of this. Whether we do this or not is your call though.

L3MON4D3:
> That's what I'm currently doing for the single operators.

Hah, Perfect :D

> Well just thought of composite ones and there is work to do (currently these opterators simply return functions -> e.g, `(f1+f2)*(f3+f4)` shouldn't work (`*` not defined for functions) so I'll have to wrap those in tables as well) [always good to have (or at least think about) tests xD]

Ohh wow, yeah completely missed that as well :facepalm:

> I agree. But I'm not sure as well whether not implement `==` or use `%` (well no one foces one to use `%` though thinking)

Only downside I can see is confusion as to what `%` does, but that's better than not offering it at all, sooo let's include it

> I'd be in favor of this. Whether we do this or not is your call though.

Yeah let's do it backwards-compatibly, re-export `extras/conditions/expand` (not 100% happy with that name btw, any ideas?) in `extras/expand_conditions` for now.

tests/unit/expand_conditions_spec.lua
```diff
@@ -0,0 +1,252 @@
...
+				vim = {api = {nvim_get_current_line = function() return "hello world ending" end}}
+				local cond = require("luasnip.extras.expand_conditions")
+				local c = cond.line_end
+				return not c("hello world") ~= false -- allow nil/object
+				]])
+			end
+			assert.has_no.errors(foo)
+			assert.is_true(foo())
+		end)
+	end)
```
L3MON4D3, line 251: I think these would be nicer as integration-tests, easier to maintain that way.

also:
faking `nvim_get_current_line` is a cool idea, the `vim`-table leaks to the following tests though (`extend_decorator` is missing `helpers.clear()` `before_each` :grimacing:)

atticus-sullivan: Well, I just tried to get around integration tests as it is tedious to always write these buffers for reference. Tbh I think it is easier to read this way where the relevant information (the content of the corresponding line) is right there.
Fixed `before_each` (hopefully, I'm still a bit unsure about the testing framework)

L3MON4D3: Ohhhh that's my bad, you can use `screen:snapshot_util()` to capture the screen. Wow srry, I completely forgot about that, pita to write them by hand😅😬
Mhmmm proceed as you like, I'm fine with unit tests as well

L3MON4D3:
> What name exactly are you not happy with, `extras/`**`condition`** or `extras/condition/`**`expand`**?

Oh, just the `expand` (one more idea: `for_expand`. It's more descriptive, but still not perfect)

tests/unit/expand_conditions_spec.lua
atticus-sullivan: ^^
Still I'd prefer the unit test tbh

atticus-sullivan:
Hm, I've got no really good idea as well.

tests/unit/expand_conditions_spec.lua
```diff
@@ -1,8 +1,13 @@
...
-
-local works = function(snippets, opts) end
+local ls_helpers = require("helpers")
 
 describe("expand_conditions", function()
+	before_each(function()
+		helpers.clear()
+		ls_helpers.session_setup_luasnip()
+		helpers.exec_lua("noop = function() end")
+	end)
```
L3MON4D3: Ah, I meant `before_each` for the `extend_decorator`-tests, the `vim`-global from the last test here isn't cleared in there

lua/luasnip/extras/conditions/expand.lua
```diff
@@ -0,0 +1,14 @@
...
+-- use the functions from show as basis and extend/overwrite functions specific for expand here
+local M = require("luasnip.extras.conditions.show")
+-----------------------
+-- PRESET CONDITIONS --
+-----------------------
+local function line_begin(line_to_cursor, matched_trigger)
+	-- +1 because `string.sub("abcd", 1, -2)` -> abc
+	return line_to_cursor:sub(1, -(#matched_trigger + 1)):match("^%s*$")
+end
+M.line_begin = cond_obj.make_condition(line_begin)
```
L3MON4D3, line 12: I think this also modifies the `conditions.show`-table, `vim.tbl_extend` should do it

L3MON4D3: no, `deepcopy` :facepalm:

tests/unit/expand_conditions_spec.lua
atticus-sullivan: Ah now I see, maybe we should simply restore `vim` after the tests in `conditions_spec.lua` as well. But I've added `before_each` in `extend_decorator` as well (hopefully like you meant)

lua/luasnip/extras/conditions/expand.lua
atticus-sullivan: Should make no difference in this case (up to now the module only contains functions) I think (not entirely sure). But deepcopy definitely is more descriptive

L3MON4D3: Yeah, `vim.tbl_extend` isn't optimal because the `copy` is a bit more hidden than with copy/deepcopy

tests/unit/expand_conditions_spec.lua
L3MON4D3: Yup, that's alright :+1:

L3MON4D3:
Okay, I think this is actually it, as far as functionality is concerned :D

atticus-sullivan:
So just to summarize what is to be done until the merge:
- [ ] write documentation
- [ ] rename `expand.lua`/`show.lua` modules if we find a better name (in my opinion `conditions.expand` looks better than `conditions.for_expand`, but maybe we find something better (though no reason to delay the merge in the end if we don't come up with something better, I think))
- [ ] update `line_end` (compare strings instead of lengths)
- [ ] rebase -> merge

(just summarizing as I can't do it right now (will be tomorrow), and we don't forget something)

DOC.md
```diff
@@ -1064,6 +1064,20 @@ ls.add_snippets("all", {
...
 ```
 
+- `conditions.show`: Contains typical predicats/functions used as
+  `show`-condition. Currently this is just `line_end`
+- `conditions.expand`: Contains typical predicats/functions used as
+  `expand`-condition. Currently this is just `line_begin`
+  Contains everything from `conditions.show` as well.
+- `conditions`: Provides a function `make_condition(foo)` which takes a function
+  as argument and returns a *condition object* for which several operators are
+  defined: `c1 + c2 -> c1 or c2`, `c1 * c2 -> c1 and c2`, `-c1 -> not c1`,
```
L3MON4D3, line 1074: I think this would be easier to take in as a list:
* `a or b`: `a + b`
* `a and b`: `a * b`
* `not a`: `-a`
* ...

The reasoning for `%` as `==` might also be interesting (i.e. `%` is very unexpected when one only knows that `==` can be overloaded, and not its limitations), maybe link your comment about that?

atticus-sullivan: No other reasons from my side. Trying to shortly (not sure if its short) explain why and for more details provide the link (so that you don't have to open/follow the link to understand but if you're interested you know where to look) sounds nice :+1:

L3MON4D3: Yeah, that's alright :+1:
Throw a few newlines in there :P (alternatively, two spaces for a new paragraph)

Oh! Another thing I don't think I ever mentioned: you can use [grip](https://github.com/joeyespo/grip) to take a look at the markdown as rendered by github, locally! Makes writing doc muuuch easier, without it those issues can only be spotted after push.

Apart from that, I've no more issues, go ahead and squash,rebase

atticus-sullivan: Oh thanks, that's a really nice hint. From a first view this looks way more flexible/suited for this case then [gollum](https://github.com/gollum/gollum) (usecase is more for the wiki stuff not single `.md` files)

DOC.md
```diff
@@ -1064,6 +1064,31 @@ ls.add_snippets("all", {
 })
 ```
 
+- `conditions.show`: Contains typical predicats/functions used as
```
L3MON4D3, line 1067: typo: predi:cat::cat: -> predicates :P (few lines lower too)

L3MON4D3:
Ah, also: would you include a deprecation-notice for `expand_conditions` in the first commit? (Or, optionally, extract the split of `expand_conditions` into a separate commit (before/after condition-objects wouldn't make much of a difference afaict))

atticus-sullivan:
Phu I know no way how to cleanly split up the commit afterwards (would be interested though if you know one). So I'll just add the note in the commit message.
By `deprecation-notice` you mean a comment in the code or a note in the commit message?

DOC.md
atticus-sullivan: Oh, right. Thanks

L3MON4D3:
> Phu I know no way how to cleanly split up the commit afterwards (would be interested though if you know one). So I'll just add the note in the commit message. By `deprecation-notice` you mean a comment in the code or a note in the commit message?

Just the commit message :+1:
AFAIK packer highlights commits that contain an exclamation mark, so best format it as `feat!` or `refactor!`, something like that.

Concerning the split yeeeah, that would be a bit involved, I think
* soft-reset all
* commit `conditions/init.lua` (+tests??)
* commit line-ending-change
* commit changes `expand_conditions`, `conditions/expand` and `conditions/show` (+tests)
* commit doc

But that's honestly too much work for the tiny benefit of having the commits techincally more "pure" (really a technicality, they're all merged at once anyway)

atticus-sullivan:
> AFAIK packer highlights commits that contain an exclamation mark, so best format it as feat! or refactor!, something like that.

Is this what you had in mind?

L3MON4D3:
Yup, that's it :+1::+1:
Thank you for another clean addition!!

mathjiajia:
It seems you haven't updated the default config file https://github.com/L3MON4D3/LuaSnip/blob/aa7acef98392f35288590f2fcf984b19e0d5bc29/lua/luasnip/config.lua#L139

atticus-sullivan:
Oh yes. Didn't thought of this file. Just noticed that the tests also still use the now deprecated module

L3MON4D3:
> Just noticed that the tests also still use the now deprecated module

I think that's okay, I don't dislike having those as security that the functions can still be accessed via the deprecated way. About config.. we should change that some time, but there's no hurry

PR #620, opened 2022-10-11
Give a valid exmple for loaders.from_lua

YuanYuYuan:
As a first-time user trying to use the Lua loaders, I found that there's no `parse` function defined through the doc. Maybe we can write down a runnable example for better guidance. Thanks!

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -2010,15 +2010,15 @@ possible to customize them by setting `snip_env` in `setup`.
 `~/snippets/all.lua`:
 ```lua
 return {
-	parse("trig", "loaded!!")
+	s("trig", t("loaded!!"))
 }
 ```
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Oh, `parse` is defined in the default `snip_env`, which means it is accessible inside files loaded by the lua-loader, even if not defined explicitly. Do the examples not run for you?

But regardless of that, it definitely makes sense to remove this as a source of confusion :+1:

YuanYuYuan:
Hi, @L3MON4D3, thanks for your clarification! Yes, it actually works great.  I'm just aware of the fact that it's simply a re-export of `parse_snippet`. But as we both agreed, let's explicitly write it down in case of any confusion. May thanks!

PR #628, opened 2022-10-14
fix: replaced spaces with tabs

Cassin01:
Those spaces cause an error (I couldn't get the error out again after fixed) when I startup my neovim.

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*           For NVIM v0.5.0           Last change: 2022 October 13
+*luasnip.txt*           For NVIM v0.5.0           Last change: 2022 October 14
 
 ==============================================================================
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Oh that seems very unexpected.
Could you show the error you got?

Cassin01:
I am unable to generate the error again . Is _setup (@local c @config.lua) something that is run once without error and never run again?

L3MON4D3:
Mhmm, that's unfortunate.. let's close this (I don't think it makes a difference), and if you encounter the error again, open an issue (or look into what's causing it and do another PR :) )

leiserfg:
The doc itself can't cause errors. It could be that whatever you are using to generate doc tags is breaking your nvim instance :thinking:

L3MON4D3:
As far as I understood the doc did for sure not cause this issue, spaces vs tabs in `vim.cmd` are the current suspect

PR #631, opened 2022-10-20
Add path to luasnips loader

rohanorton:
In order to keep my snippets in better order, I wanted to split my snippets into multiple files (one snippet per file), as here:

```
~/.config/nvim/snippets
├── all.lua
├── ...
└── lua
    ├── init.lua
    ├── snippet_1.lua
    ├── snippet_2.lua
    └── snippet_3.lua
```

This means having a `snippets/lua/init.lua` file that imports the other files in the directory, e.g.

```lua
local snippets = {
    require('lua.snippet_1'),
    require('lua.snippet_2'),
    require('lua.snippet_3'),
}

return snippets, {}
```

The problem that I ran into was that the require function didn't work as it didn't have an appropriate path property, so could not resolve the imports.

In order to resolve this issue, I have implemented a new property on the config, `snip_path`, which takes a lua path string.

If this `snip_path` property exists it is then injected into the snippet (at the same time as the env is injected), changing the `package.path` property which is used in Neovim to resolve imports.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,5 +163,6 @@ Here are some LuaSnip videos and tutorials on the Web:
 Setting `snip_env` to `{ some_global = "a value" }` will add the global variable `some_global` while evaluating these files.
 If you mind the (probably) large number of generated warnings, consider adding the keys set here to the globals
 recognized by lua-language-server or add `---@diagnostic disable: undefined-global` somewhere in the affected files.
+- `snip_path`: Define the require path of lua environment of files loaded by the [lua-loader](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#lua-snippets-loader).
 
 Inspired by [vsnip.vim](https://github.com/hrsh7th/vim-vsnip/)
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*           For NVIM v0.5.0           Last change: 2022 October 14
+*luasnip.txt*           For NVIM v0.5.0           Last change: 2022 October 20
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -604,6 +604,12 @@ local function refresh_notify(ft)
 	end
 end
 
+local function setup_snip_path()
+	if session.config.snip_path then
+		package.path = session.config.snip_path
+	end
+end
+
 local function setup_snip_env()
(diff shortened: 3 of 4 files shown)


L3MON4D3:
Mhmm, interesting idea, but luasnip should be able to load the file-structure you showed above just fine, have you tried that?

rohanorton:
Hi @L3MON4D3, 

Thanks for taking a look!

Yes, I did try as you suggest. It works when I open my editor inside the nvim config directory (provided I change the require statements a little: `require("lua.snippet_.lua") => require("snippets.lua.snippet_.lua")`). But when I open the editor from any other directory, it fails. 

Logging out `package.path` from my snippet file, I get the following path entries:

```
./?.lua
/usr/local/share/luajit-2.1.0-beta3/?.lua
/usr/local/share/lua/5.1/?.lua
/usr/local/share/lua/5.1/?/init.lua
/Users/rohan/.cache/nvim/packer_hererocks/2.1.0-beta3/share/lua/5.1/?.lua
/Users/rohan/.cache/nvim/packer_hererocks/2.1.0-beta3/share/lua/5.1/?/init.lua
/Users/rohan/.cache/nvim/packer_hererocks/2.1.0-beta3/lib/luarocks/rocks-5.1/?.lua
/Users/rohan/.cache/nvim/packer_hererocks/2.1.0-beta3/lib/luarocks/rocks-5.1/?/init.lua
```

As you can see, the first entry is `./?.lua`, which is what allows it to work when editing inside the nvim config directory. But, because it is a relative path, when the working directory changes, it ceases to function.

rohanorton:
Derp. Having written all that up, it occurred to me that the path is being inherited from vim itself, so I can just change the path in vim and not require any of the changes made here.

L3MON4D3:
Ah, srry, I wasn't very precise:
What I meant was that luasnip can load
```
snippets/
  json.lua
  markdown.lua
  lua/
    snippet1.lua
    snippet2.lua
```
natively

L3MON4D3:
> Derp. Having written all that up, it occurred to me that the path is being inherited from vim itself, so I can just change the path in vim and not require any of the changes made here.

Hah, yeah that will also work :+1:

I'd recommend `dofile` instead of `require`, it can prevent issues due to aliasing (if the snippet-file has the same name as some other module, there might arise some confusion)

rohanorton:
> Ah, srry, I wasn't very precise: What I meant was that luasnip can load
> 
> ```
> snippets/
>   json.lua
>   markdown.lua
>   lua/
>     snippet1.lua
>     snippet2.lua
> ```
> 
> natively

Thank you, this works perfectly!

PR #654, opened 2022-11-10
lua-loader: add file-global table through which snippets may be added

L3MON4D3:
We can make the lua loader more flexible by not only loading snippets from the tables returned by each file, but by also loading snippets from some global (file-global via `setfenv`, because it shouldn't be visible from the outside, and every file needs its own) table.
Doing so allows us to, for example, add a new function to `snip_env`, which automatically adds snippets defined through it:
```lua
snip_env = {
	...,
	s = function(...)
		local snip = ls.s(...)
		-- getfenv is necessary here, just `ls.file_snippets` would immediately upvalue to `nil`.
		-- 2 to get the environment of the function calling this.
		table.insert(getfenv(2).ls_file_snippets, snip)
	end,
}
```
which I think is a pretty great improvement over the current "return it in a table"-approach.

I'm not sure yet how to best introduce this, afaict just replacing `s` in the default snip_env might break stuff.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -2707,14 +2707,21 @@ the lazy_load.
 - `refresh_notify(ft:string)`: Triggers an autocmd that other plugins can hook
   into to perform various cleanup for the refreshed filetype.
   Useful for signaling that new snippets were added for the filetype `ft`.
+
 - `set_choice(indx:number)`: Changes to the `indx`th choice.
   If no `choiceNode` is active, an error is thrown.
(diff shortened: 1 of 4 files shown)


L3MON4D3:
Okay, this is ready for merge as far as I'm concerned. I've added an example for the modified `snip_env` to the doc, should be visible enough there.
I'll merge this sometime tomorrow, or monday, if anyone has some comments, get them in now :P

PR #658, opened 2022-11-15
support wordTrig in vscode snippet

tomtomjhj:


L3MON4D3:
Nice👍
Could you add, in the commitmessage, that this does not change existing setups since `wordTrig=nil` will default it to true?

L3MON4D3:
Very nice, thank you!
Would you add this to `DOC.md`? There's a [short section](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#vscode) on luasnip-specific options in vscode-snippets (sorry for another back-forth for this, I didn't think of it earlier :sweat_smile:)

Issue #465, opened 2022-06-17
fmt and multiple key node


uyha:
When the snippet below is expanded, only the last node is being replaced with actual value when typing, the first `{answer}` stays 42, is it possible that `fmt` automatically recognizes that the key is being reused and replaces it with `rp`?
```fmt("{answer} {answer}", {answer= i(1, "42)}```

L3MON4D3:
I think we're inserting a [copy](https://github.com/L3MON4D3/LuaSnip/blob/6e10a30178240182781955ce062618c8d793a37b/lua/luasnip/extras/fmt.lua#L94-L102).

uyha:
I'm not sure why a copy is used instead of a repeat, wouldn't a repeat be better in this case?

L3MON4D3:
In this case yes, we could probably implement that as well, but I'm not sure it's the right thing to do: `fmt` is very simple and easy to reason about (although inserting a copy already doesn't 100% fit with that, reprospectively I would've preferred an error for duplicated keys), tacking more functionality on it does not seem right.

uyha:
I see, my thinking was it is just kind of expected that it would allow reusing keys since (I think) that is the main use case for that, otherwise it provides very little benefit compare to just using indices. In case you don't think it should be implemented, then maybe it's reasonable to mention in DOC.md that this is not supported.

L3MON4D3:
Reusing keys is allowed (okay they're copied which doesn't do much, I agree), what `fmt` does not do is generate new nodes.
But that being your assumption is good to know, I'll mention it in the doc.
Thank you :)

uyha:
My wording was bad there, I mean they are technically allowed, but they do something wildly unexpected. I think saying that `fmt` doesn't generate new nodes in this case is not true, since it does generate new nodes, it just doesn't generate unmentioned types of nodes (I am being nitpicky here). I still think it's better to insert repeat nodes, since it helps alot in practice.

L3MON4D3:
Okay, also bad wording on my part, "new nodes" -> "different nodes".

We could probably introduce duplicate->repeat-node as a setting for `fmt` (though it'll default to off since I'm not confident that there aren't configs which rely on the current behaviour).

uyha:
The current behaviour is a bit nonsensical to rely upon in my opion (at least I don't see how you can do something useful with it but you'll never know), but it is awesome that we can opt in into the new behaviour.

danielo515:
[...]

PR #665, opened 2022-11-19
feat(extras): Insert repeat node when duplicated keys occur

uyha:
Experimental implementation for #465

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -124,6 +124,11 @@ only apply to some nodes (`user_args` for both function and dynamicNode).
 These `opts` are only mentioned if they accept options that are not common to
 all nodes.
 
+## Node-Api:
+
+- `get_jump_index()`: this method returns the jump-index of a node. If a node 
+  doesn't have a jump-index, this method returns `nil` instead.
+
 # SNIPPETS
 
 The most direct way to define snippets is `s`:
@@ -237,7 +242,7 @@ which is passed to the function.
 (in most cases `parent == parent.snippet`, but the `parent` of the dynamicNode
 is not always the surrounding snippet, it could be a `snippetNode`).
 
-## Api:
+## Snippet-Api:
 
 - `invalidate()`: call this method to effectively remove the snippet. The
   snippet will no longer be able to expand via `expand` or `expand_auto`. It
@@ -1204,6 +1209,13 @@ ls.add_snippets("all", {
   }, {
     delimiters = "<>"
   })),
+  s("example4", fmt([[
+  repeat {a} with the same key {a}
+  ]], {
+    a = i(1, "this will be repeat")
+  }, {
+    repeat_duplicates = true
+  }))
 })
 ```
 
@@ -1238,6 +1250,9 @@ any way, correspond to the jump-index of the nodes!
   	when passing multiline strings via `[[]]` (default true).
   * `dedent`: remove indent common to all lines in `format`. Again, makes
   	passing multiline-strings a bit nicer (default true).
+  * `repeat_duplicates`: repeat nodes when a key is reused instead of copying
+        the node if it has a jump-index, refer to [jump-index](#jump-index) to
+        know which nodes have a jump-index (default false).
 
 There is also `require("luasnip.extras.fmt").fmta`. This only differs from `fmt`
 by using angle-brackets (`<>`) as the default-delimiter.
--- a/lua/luasnip/extras/fmt.lua
+++ b/lua/luasnip/extras/fmt.lua
@@ -2,6 +2,7 @@ local text_node = require("luasnip.nodes.textNode").T
 local wrap_nodes = require("luasnip.util.util").wrap_nodes
 local extend_decorator = require("luasnip.util.extend_decorator")
 local Str = require("luasnip.util.str")
+local rp = require("luasnip.extras").rep
 
 -- https://gist.github.com/tylerneylon/81333721109155b2d244
 local function copy3(obj, seen)
@@ -39,11 +40,13 @@ end
 --   opts:
 --     delimiters: string, 2 distinct characters (left, right), default "{}"
 --     strict: boolean, set to false to allow for unused `args`, default true
+--     repeat_duplicates: boolean, repeat nodes which have jump_index instead of copying them, default false
 -- Returns: a list of strings and elements of `args` inserted into placeholders
 local function interpolate(fmt, args, opts)
 	local defaults = {
 		delimiters = "{}",
 		strict = true,
+		repeat_duplicates = false,
 	}
 	opts = vim.tbl_extend("force", defaults, opts or {})
 
@@ -97,7 +100,12 @@ local function interpolate(fmt, args, opts)
 		-- The nodes are modified in-place as part of constructing the snippet,
 		-- modifying one node twice will lead to UB.
 		if used_keys[key] then
-			table.insert(elements, copy3(args[key]))
+			local jump_index = args[key]:get_jump_index() -- For nodes that don't have a jump index, copy it instead
+			if not opts.repeat_duplicates or jump_index == nil then
+				table.insert(elements, copy3(args[key]))
+			else
+				table.insert(elements, rp(jump_index))
+			end
 		else
 			table.insert(elements, args[key])
 			used_keys[key] = true
@@ -211,6 +219,7 @@ local function format_nodes(str, nodes, opts)
 		end
 	end, parts)
 end
+
 extend_decorator.register(format_nodes, { arg_indx = 3 })
 
 return {
--- a/tests/unit/fmt_spec.lua
+++ b/tests/unit/fmt_spec.lua
@@ -7,9 +7,8 @@ local works = function(params)
 		-- fails if number- and string-keys are mixed ({1,2} works fine, {1, b=2} fails).
 		-- So we limit ourselves to just passing strings, which are then turned into tables
 		-- while load()ing the function.
-		local result = helpers.exec_lua(
-			string.format(
-				[[
+		local result = helpers.exec_lua(string.format(
+			[[
 						local args = %s
 						local mock_args = {}
 
@@ -35,11 +34,10 @@ local works = function(params)
 						end
 						return str_result
 						]],
-				params.args,
-				params.fmt,
-				params.opts
-			)
-		)
+			params.args,
+			params.fmt,
+			params.opts
+		))
 		assert.are.same(params.expected, result)
 	end)
 end
--- a/tests/unit/fmt_spec.lua
+++ b/tests/unit/fmt_spec.lua
@@ -1,20 +1,46 @@
 local helpers = require("test.functional.helpers")(after_each)
 
-local works = function(msg, fmt, args, expected, opts)
-	it(msg, function()
+---@param params { msg: string, fmt: string, args: string, expected: string, opts: string, prolouge: string}
+local works = function(params)
+	it(params.msg, function()
 		-- normally `exec_lua` accepts a table which is passed to the code as `...`, but it
 		-- fails if number- and string-keys are mixed ({1,2} works fine, {1, b=2} fails).
 		-- So we limit ourselves to just passing strings, which are then turned into tables
 		-- while load()ing the function.
 		local result = helpers.exec_lua(
 			string.format(
-				'return require("luasnip.extras.fmt").interpolate("%s", %s, %s)',
-				fmt,
-				args,
-				opts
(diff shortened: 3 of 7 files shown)


L3MON4D3:
Nice so far :+1:
I'd like to see the access to `node.pos` wrapped in a function, like `node.get_jump_indx`. (I think this should make clear that the jump-index may not be changed)

lua/luasnip/extras/fmt.lua
```diff
@@ -97,7 +99,11 @@ local function interpolate(fmt, args, opts)
 		-- The nodes are modified in-place as part of constructing the snippet,
 		-- modifying one node twice will lead to UB.
 		if used_keys[key] then
-			table.insert(elements, copy3(args[key]))
+			if opts.dup_copy then
+				table.insert(elements, copy3(args[key]))
+			else
+				table.insert(elements, rp(args[key].pos))
```
L3MON4D3, line 105: Oh, some nodes (think `functionNode`) don't have a jump-indx (/`pos`), they should still be copied (it's fine in their case)

uyha:
Hmmm, isn't it better to stick with `pos` for now and check if `pos` is `nil` or not to decide between `copy3` and `rp`? or do you have in mind that `get_jump_index` can be used in other use cases?

L3MON4D3:
> Hmmm, isn't it better to stick with `pos` for now and check if `pos` is `nil` or not to decide between `copy3` and `rp`? 

Oh, yeah that's what I also had in mind, I only think we should wrap the access to `pos` in a function (so the nodes implement a new function `node:get_jump_indx()` which returns `node.pos`)
Both because `pos` isn't very informative (could be the jump-index, or the position in the buffer, or the position of the node in its parent), and because it feels like a function is the nicer interface.

> do you have in mind that `get_jump_index` can be used in other use cases?

Not really, but I'd like the extras to not rely on any internals of luasnip, only public API.

uyha:
I added the function, can you have another look?

L3MON4D3:
Yup :+1:
There's one issue with the [tests](https://github.com/L3MON4D3/LuaSnip/blob/73865a6f8e0058ddc5b299f8db47fac14cdcb0df/tests/unit/fmt_spec.lua#L36-L170):  they expect numbers (for simplicity, and because it used to be pretty generic), but now we need to make it work with nodes (or objects that pretend to be nodes, eg. implement exactly the functionality we need in `interpolate`).
I think the best way to fix that is just mocking the nodes, and adding a `__tostring`-metamethod to still get the expected strings from `table.concat`.

Aside from that, just docs are left, and some tests for the new functionality.
Are you familliar with neovim's test-framework?

lua/luasnip/extras/fmt.lua
```diff
@@ -44,6 +45,7 @@ local function interpolate(fmt, args, opts)
 	local defaults = {
 		delimiters = "{}",
 		strict = true,
+		dup_copy = true,
```
L3MON4D3, line 48: I feel like `repeat_duplicates` is a bit more descriptive, wdyt?

uyha:
I am not familiar with it, but I will take a look.

lua/luasnip/extras/fmt.lua
uyha: yes, I thought `dup_copy` is good enough but `repeate_duplicates` works also

uyha:
so, i am getting to the mocking the nodes, but i am not sure how to actually do it since it requires the node definition to be global and I am not sure how to proceed with that.

L3MON4D3:
Oh, I think it's enought to only mock the functions we actually need (so just `get_jump_indx`, and I think that can just always return nil).
Since those tests are mainly about making sure the interpolation is correct, we only need to simulate nodes so far that it actually works, but no further.

For actual tests of the functionality, I'd say we can test it in a real setting, I feel like that's easiest to write tests for too (use `screen:snapshot_util()` to generate a capture of the tested neovim-instance, and test it via `screen:expect` (bunch of examples for that in `tests/integration`))

uyha:
but the problem i have here is not knowing how to pass a mock object to the `works` call, I tried putting

```lua
local FakeNode = {}
function fake_node() end
```

at the beginning of `fmt_spec.lua` and have a call like

```lua
works(“”, “{a}{a}”, “{a = fake_node()}”, “”, “{repeat_duplicates = true}”)
```

I know that it shouldn’t work but for the reason of the test case being wrong, here the test case fails due to `fake_node` is not visible to the caller.

L3MON4D3:
Ahhh, you can run code in the tested neovim-instance via `exec_lua`, defining `fakeNode` with that should work

uyha:
I added a test, it is messy but I want to see if the direction is correct first. Also, it is a bit weird since tests like this won't tell if there `repeat` is being used or not.

L3MON4D3:
That test looks good👍👍
Can you make it generic, so the existing tests can be re-used? (ie put part of the new test into the works-function)

> Also, it is a bit weird since tests like this won't tell if there repeat is being used or not

Yeah, it's unfortunate that the current tests for just the interpolation rely on this function handling objects other than nodes :/
The actual new tests for this feature should probably be integration-tests (so create a snippet with repeated nodes, expand it, and check that everything works as expexted), at least that's how I like to test most features.

uyha:
I generalize the `works` function, but there is a problem. Since `interpolate`'s result will have a `repeat` node which doesn't have a `__tostring`, so it can't be stringified. Any idea how i should proceed?

uyha:
The previous code works because I have a bug in the mock table btw

L3MON4D3:
> I generalize the `works` function, but there is a problem. Since `interpolate`'s result will have a `repeat` node which doesn't have a `__tostring`, so it can't be stringified which leads to the test failing. Any idea how i should proceed?

Turn of repeat-insertion for those tests? That should be fine as long as we check it elsewhere

uyha:
I am not sure what you mean by turning off repeat-insert here

L3MON4D3:
Ah, `repeat_duplicates` to `false`

uyha:
oh, ok

L3MON4D3:
I took the liberty of implementing my idea of handling the existing tests (figured explaining would take longer than quickly handling it myself :sweat_smile:) WDYT?

uyha:
I was trying to make it as general as possible since I am not sure how you want it, but this is simpler in my opinion. So, I should move on to create the integration test, right?

L3MON4D3:
Ah, I might've been a bit misleading then, I only wanted to re-use those tests (but they're also more readable now, with the explicit argument-names, so that's nice :D)

Umm yeah, integration tests sound good👍👍

L3MON4D3:
Oh great idea to forward `TEST_FILE` :+1:

uyha:
yeah, it was annoying having to run the whole test set everytime

uyha:
I added 2 tests, 1 one for `fmt` and 1 for `rep`, both of them are failing but I'm not sure why since testing on real nvim instance works fine

L3MON4D3:
Ah, `screen:expect` is very particular about the text it receives :/
Try formatting the text there like in the passing tests (especially whitespace!)

uyha:
that’s not really the problem, the problem is the repeating text doesn’t appear at all in the actual portion of the screen 😅

uyha:
```
Expected:
  |*asdf^ repeat asdf                                  |
Actual:
  |*asdf^ repeat                                       |
```

L3MON4D3:
Ahhh xD
Okay that's probably due to `updateevents` being different, try jumping after feeding the text (jump always causes an update)

uyha:
eh, how do you do a jump in this case 😅?

L3MON4D3:
Oh, manually, binding some keys to jump is not necessary: `exec_lua(ls.jump())`

uyha:
I think it is done now, could you have another look?

DOC.md
```diff
@@ -1238,10 +1245,15 @@ any way, correspond to the jump-index of the nodes!
   	when passing multiline strings via `[[]]` (default true).
   * `dedent`: remove indent common to all lines in `format`. Again, makes
   	passing multiline-strings a bit nicer (default true).
+  * `repeat_duplicates`: repeat nodes when a key is reused instead of copying
```
L3MON4D3, line 1248: Oh, best add a "when possible", for example functionNode can, and will, not be repeated

lua/luasnip/extras/fmt.lua
```diff
@@ -219,4 +228,5 @@ return {
 	-- alias
 	fmt = format_nodes,
 	fmta = extend_decorator.apply(format_nodes, { delimiters = "<>" }),
+	fmtrp = extend_decorator.apply(format_nodes, { repeat_duplicates = true }),
```
L3MON4D3, line 231: Ah, I'm a bit on the fence with this, normally `extend_decorator` should be used locally to generate such "overloads".
(fmta is an exception because it existed before extend_decorator, and there's not much reason to remove it)

Do you have some arguments in favor of providing this overload here?

uyha: Not really, I just think it’s nice to have a shortcut for this, I guess I can add this in my local config also.

DOC.md
uyha: ok, i’ll add it when i’m back from work

lua/luasnip/extras/fmt.lua
L3MON4D3: I think I'd prefer that, it's easy to add them, and if we add more here it'll be hard to draw a line in the future

uyha: yup

L3MON4D3: Ty😊

uyha:
sure

uyha:
can you have a look again?

uyha:
welp, looks like either github action's internet is down or neovim-ppa is down

L3MON4D3:
off and on again did it :+1:

uyha:
hmmm, it seems like i messed up the commits, i will fix it

L3MON4D3:
Ah my bad, I just noticed that `get_jump_index` should be officially exported -.-
Best add a subsection `Api` (like in `Snippets` below) [here](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#node)

uyha:
everything should be good now, could you have another look?

L3MON4D3:
Ohh right, just `Api` doesn't work, then that tag is duplicated.. Could you change the titles to `Snippet-Api` and `Node-Api`?

lua/luasnip/extras/fmt.lua
```diff
@@ -97,7 +100,18 @@ local function interpolate(fmt, args, opts)
...
 		-- modifying one node twice will lead to UB.
 		if used_keys[key] then
-			table.insert(elements, copy3(args[key]))
+			local jump_index = nil -- For nodes that don't have a jump index, copy it instead
+			if
+				type(args[key]) == "table"
+				and args[key].get_jump_index ~= nil
+			then
+				jump_index = args[key]:get_jump_index()
+			end
```
L3MON4D3, line 109: Wait, this isn't supposed to be here..

uyha: yeah, i messed up when i rearranged the commits, sorry

uyha: ok, I hope there's no more problems

L3MON4D3: Okay :D

tests/unit/fmt_spec.lua
```diff
@@ -1,20 +1,31 @@
...
+                        %s
+
+                        local result = require("luasnip.extras.fmt").interpolate("%s", %s, %s)
+                        local str_result = ""
+                        for _, value in pairs(result) do
+                          str_result = str_result .. tostring(value)
+                        end
+                        return str_result
+                        ]],
+				params.prolouge or "",
```
L3MON4D3, line 22: Oh, this change is missing as well.

uyha: could you add back the change here?

L3MON4D3: Yeah, ofc :)

uyha:
looks like everything is in order

L3MON4D3:
I agree :+1::+1:
Nice work, thank you very much for sticking through all of my demands :sweat_smile:

uyha:
not a problem, thank you very much for the plugin

danielo515:
This is awesome. Thanks

PR #666, opened 2022-11-20
fix(util): Path.expond current working directory expansion expanding any '.' prefixed path

mrzzy:
## Issue
`Path.expand()` incorrectly expands all `.` in paths as `MYCONFIG_ROOT`.
- eg. `Path.expand(".local/snippets")` will be expanded as `MYCONFIG_ROOT .. "local/snippets"` which is unintended.
- expanding just `.` is ambiguous : are we using `.` as part of a hidden path or alias for the current working directory?

## Contents
Patch `Path.expand()` to only substitute `.` + OS file separator   for `MYCONFIG_ROOT`  instead of just `.`

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*           For NVIM v0.5.0          Last change: 2022 November 17
+*luasnip.txt*           For NVIM v0.5.0          Last change: 2022 November 20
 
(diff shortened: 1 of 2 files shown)


leiserfg (approved):

L3MON4D3:
Ahh, ofc!
Thank you!

PR #668, opened 2022-11-22
fix: <nomodeline> when triggering snippet events

tomtomjhj:
See also: https://github.com/L3MON4D3/LuaSnip/commit/67e203f46149437f04a1fe82d33067ffe71d3b1b


L3MON4D3:
Aha, I missed those then.
Thank you :+1:

PR #680, opened 2022-12-01
Fixed syntax in example under "Environment Namespaces" section

Andrew15-5:
Fixed code syntax for it to work with Lua (5.2.4)


Andrew15-5:
Either this is because of version difference, or you shouldn't forget to run your example code to make sure that at least its syntax is correct. :)

L3MON4D3:
Whoops :sweat_smile: 
Thank you!

PR #691, opened 2022-12-11
log: don't print a message on each startup

L3MON4D3:
Pretty straightforward, printing a message on each startup is not a good way to provide some quick way to verify that the log is working.
Instead, provide a `ping`-method which just appends something to the log.

I'll leave this up for a bit so whoever wants to, can comment on it.


L3MON4D3:
OTOH, that seems like it's complicating the whole thing a bit, this should do it

Issue #706, opened 2022-12-29
<C-r> triggers autosnippets when the trigger is the last part of the pasted text


carbon-steel:
I've noticed that autosnippets sometimes get triggered when using `<C-r>` in insert mode. This causes yanked text to come out differently when pasted via `<C-r>`.

Here is a minimal example:
```
-- init.lua
local lazypath = "/tmp/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

require("lazy").setup({'L3MON4D3/LuaSnip' }, { root = "/tmp/lazy", lockfile = "/tmp/config/lazy-lock.json" }
)


local ls = require("luasnip")

ls.config.set_config {
  enable_autosnippets = true,
}

local s = ls.snippet
local t = ls.text_node

local snippets = {
s({trig = ".c", wordTrig=false}, t('hi')),
}

ls.add_snippets("all", snippets, {type="autosnippets"})

```

Steps to reproduce:
1. start `nvim -u init.lua` with the init.lua above.
2. Set paste mode with `:set paste`. We do this to avoid triggering the snippet at this point.
3. Write: `hello.c`
4. Turn paste mode off with `:set nopaste`
5. Delete the word with `diW`. It is very important to use this specific command as this will not work if you also yank the newline at the end of the word.
6. Go into insert mode.
7. Press `<C-r>"`. You should see that the pasted text actually is not the same as the original due to the autosnippet being triggered. It will say `hellohi`.

Without the specific condition that the trigger appears at the very end of the pasted text, pasted text do not trigger autosnippets. This can be quite annoying since people don't expect pasted text to be different from what they yanked. I believe text pasted with `<C-r>` should never trigger autosnippets. Can this fix be made?

L3MON4D3:
Oh, that's very valid.
Unfortunately, I've no idea how to detect this state, and prevent auto-expansion. If you want to play around with it, the autocommands for autosnippets are defined [here](https://github.com/L3MON4D3/LuaSnip/blob/5570fd797eae0790affb54ea669a150cad76db5d/lua/luasnip/config.lua#L192-L193), maybe CharInsertPre was the wrong choice, or there's some way to detect pasting at that point

PR #709, opened 2022-12-30
Fix autosnippets

carbon-steel:
Hey @L3MON4D3, this PR is in regards to https://github.com/L3MON4D3/LuaSnip/issues/706.

You'll have to forgive my code and inexperience. I've only ever contributed to C++ codebases. I believe the tests were meant for an Ubuntu environment. I am running arch so I'll have to figure out how to run them tomorrow (unless you know what I should do?).

I'd be happy to hear any notes you may have :)

### DIFF
--- a/lua/luasnip/config.lua
+++ b/lua/luasnip/config.lua
@@ -188,9 +188,12 @@ c = {
 			"Remove buffers' nodes on deletion+wipeout.
 			autocmd BufDelete,BufWipeout * lua current_nodes = require("luasnip").session.current_nodes if current_nodes then current_nodes[tonumber(vim.fn.expand("<abuf>"))] = nil end
 		]]
-					.. (session.config.enable_autosnippets and [[
+					.. (session.config.enable_autosnippets and
+        -- Luasnip_just_inserted makes sure we only trigger autosnippets after we inserted characters in insert mode.
+        -- vim.b.Luasnip_last_changedtick allows us to compute how many changes occurred in the file between two TextChanged* events. We use this information to distinguish typing characters by hand vs using <C-r> to paste multiple characters in insert mode. We want to allow expanding autosnippets in the first case and not in the second. After some investigation, this is the only way I found to distinguish the two cases.
+        [[
 			autocmd InsertCharPre * lua Luasnip_just_inserted = true
-			autocmd TextChangedI,TextChangedP * lua if Luasnip_just_inserted then require("luasnip").expand_auto() Luasnip_just_inserted=nil end
+			autocmd TextChangedI,TextChangedP * lua if Luasnip_just_inserted then Luasnip_just_inserted=nil if vim.b.Luasnip_last_changedtick == nil or vim.b.changedtick==vim.b.Luasnip_last_changedtick + 1 then require("luasnip").expand_auto() end vim.b.Luasnip_last_changedtick = vim.b.changedtick end
 		]] or "")
 					.. [[
 		augroup END
--- a/lua/luasnip/config.lua
+++ b/lua/luasnip/config.lua
@@ -188,13 +188,16 @@ c = {
 			"Remove buffers' nodes on deletion+wipeout.
 			autocmd BufDelete,BufWipeout * lua current_nodes = require("luasnip").session.current_nodes if current_nodes then current_nodes[tonumber(vim.fn.expand("<abuf>"))] = nil end
 		]]
-					.. (session.config.enable_autosnippets and
-        -- Luasnip_just_inserted makes sure we only trigger autosnippets after we inserted characters in insert mode.
-        -- vim.b.Luasnip_last_changedtick allows us to compute how many changes occurred in the file between two TextChanged* events. We use this information to distinguish typing characters by hand vs using <C-r> to paste multiple characters in insert mode. We want to allow expanding autosnippets in the first case and not in the second. After some investigation, this is the only way I found to distinguish the two cases.
-        [[
+					.. (
+						session.config.enable_autosnippets
+							-- Luasnip_just_inserted makes sure we only trigger autosnippets after we inserted characters in insert mode.
+							-- vim.b.Luasnip_last_changedtick allows us to compute how many changes occurred in the file between two TextChanged* events. We use this information to distinguish typing characters by hand vs using <C-r> to paste multiple characters in insert mode. We want to allow expanding autosnippets in the first case and not in the second. After some investigation, this is the only way I found to distinguish the two cases.
+							and [[
 			autocmd InsertCharPre * lua Luasnip_just_inserted = true
 			autocmd TextChangedI,TextChangedP * lua if Luasnip_just_inserted then Luasnip_just_inserted=nil if vim.b.Luasnip_last_changedtick == nil or vim.b.changedtick==vim.b.Luasnip_last_changedtick + 1 then require("luasnip").expand_auto() end vim.b.Luasnip_last_changedtick = vim.b.changedtick end
-		]] or "")
+		]]
+						or ""
+					)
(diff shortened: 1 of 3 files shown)


L3MON4D3:
Oh, I'm also running them on arch, they should work as-is.
Could you outline why this works?
Does it work if only a single character is inserted?

L3MON4D3:
Oh and a big thank you already for taking a look, awesome! :)

carbon-steel:
> Oh, I'm also running them on arch, they should work as-is.

Ah, I got confused because of the use of apt-get in the github workflows.

> Could you outline why this works?

[b:changedtick](https://neovim.io/doc/user/eval.html#b%3Achangedtick) is a monotonically increasing per-buffer count of the number of changes made to the buffer.

You can observe its behavior by using:
`:autocmd TextChangedI * echom b:changedtick`

Then type some stuff on the buffer you called the command from. You can then see how `b:changedtick` changes with each character you insert by looking at the printed results:
`:messages`

You can see how `b:changedtick` gets incremented by one with each character you type in. If you yank some string with multiple characters and you paste it (using <C-r> in insert mode), you'll notice that `b:changedtick` increases by the number of characters that were pasted. However, the TextChangedI event only occurs once after the paste is complete. By this, we can distinguish when a TextChangedI event was triggered by single character from a normal keypress vs multiple characters from <C-r>. I use this to prevent autosnippets from expanding when text was pasted using <C-r>.

> Does it work if only a single character is inserted?

No, this will not work if you paste a single character with <C-r> (I checked) since b:changedtick's behavior will be the same as if you had typed that single character in yourself.

carbon-steel:
I believe remapping `<C-r>` to just use the paste command from the vim.api could fix this problem for all cases:
```
local function insertModePaste()
  local register = vim.cmd([[
  let s:register = getchar()
  echo  s:register
  ]], true)
  vim.api.nvim_paste(vim.fn.getreg(register), true, -1)
end

map('i', '<C-r>', insertModePaste)
```

I don't know if you think this should go in the plugin, but users could certainly use this code to completely fix this issue for themselves for all cases.

L3MON4D3:
Thanks for the clarification👍
Seems like an okay solution, it's a bit involved, and doesn't work for all cases (if the first action is such a paste, luasnip_changedtick would be nil and we would auto-expand (not 100% sure this is possible but it might be)), but it is better that the status quo.

Changing C-R seems dangerous 😅😅

AFAICT ultisnips could have had this same issue, would you be interested in figuring out how they handled it?

carbon-steel:
I've fixed the test that I broke and I added a test to document and preserve the new behavior. I've been using Luasnip with my changes and it seems to work well. I think this pull request is ready to go.

What do you think about adding the code from [my above post](https://github.com/L3MON4D3/LuaSnip/pull/709#issuecomment-1368021442) to the wiki to let people know they can fully fix this issue for themselves if the add it to their configs?

carbon-steel:
I just tested the pasting triggers in insert mode with ultisnips and their autosnippets just expand in every case. It looks like they did not solve this problem.

L3MON4D3:
Thank you for taking a look at ultisnips, in that case this seems to be the best way to do it

> I've fixed the test that I broke and I added a test to document and preserve the new behavior. I've been using Luasnip with my changes and it seems to work well. I think this pull request is ready to go.
> 
> What do you think about adding the code from [my above post](https://github.com/L3MON4D3/LuaSnip/pull/709#issuecomment-1368021442) to the wiki to let people know they can fully fix this issue for themselves if the add it to their configs?

Yes! Would you mention that in the commit-message as well? Or put it into the doc (since it fixes an issue, and is not just a new feature), though I'm not sure which section would be appropriate

lua/luasnip/config.lua
```diff
@@ -188,10 +188,16 @@ c = {
 			"Remove buffers' nodes on deletion+wipeout.
 			autocmd BufDelete,BufWipeout * lua current_nodes = require("luasnip").session.current_nodes if current_nodes then current_nodes[tonumber(vim.fn.expand("<abuf>"))] = nil end
 		]]
-					.. (session.config.enable_autosnippets and [[
+					.. (
+						session.config.enable_autosnippets
+							-- Luasnip_just_inserted makes sure we only trigger autosnippets after we inserted characters in insert mode.
+							-- vim.b.Luasnip_last_changedtick allows us to compute how many changes occurred in the file between two TextChanged* events. We use this information to distinguish typing characters by hand vs using <C-r> to paste multiple characters in insert mode. We want to allow expanding autosnippets in the first case and not in the second. After some investigation, this is the only way I found to distinguish the two cases.
```
L3MON4D3, line 194: I'm not 100% happy with this comment, I think it could be more descriptive, ie:
1. what is the issue
2. how can typing characters and pasting multiple characters be distinguished
3. the solution

Also, could you make the comment 80-100 characters wide? Normally stylua should take care of this, but here it doesn't :shrug:

lua/luasnip/config.lua
```diff
@@ -188,10 +188,16 @@ c = {
...
+							and [[
 			autocmd InsertCharPre * lua Luasnip_just_inserted = true
-			autocmd TextChangedI,TextChangedP * lua if Luasnip_just_inserted then require("luasnip").expand_auto() Luasnip_just_inserted=nil end
-		]] or "")
+			autocmd TextChangedI,TextChangedP * lua if Luasnip_just_inserted then Luasnip_just_inserted=nil if vim.b.Luasnip_last_changedtick == nil or vim.b.changedtick==vim.b.Luasnip_last_changedtick + 1 then require("luasnip").expand_auto() end vim.b.Luasnip_last_changedtick = vim.b.changedtick end
```
L3MON4D3, line 197: Similar here, could you spread this over multiple lines to increase readability?
Alternative would be toswitch to lua-autocommands, but I don't think this alone warrants it.

lua/luasnip/config.lua
L3MON4D3: (Actually, maybe stylua isn't supposed to..)

L3MON4D3:
> Or put it into the doc (since it fixes an issue, and is not just a new feature), though I'm not sure which section would be appropriate

Actually, this behaviour might be unexpected, would you add a new section on it in `DOC.md`?

tests/integration/add_snippets_spec.lua
```diff
@@ -291,6 +303,22 @@ describe("add_snippets", function()
 			{2:-- INSERT --}                                      |]],
 		})
 
+		-- Test to make sure that autosnippets do not get triggered while pasting in insert mode.
+		feed("<ESC>dd") -- clear line
+		helpers.feed_command("set paste") -- disable autosnippets
```
L3MON4D3, line 308: Oh, why does `set paste` disable autosnippets? Had a cursory read of the help-section on it, but nothing in there caught my eye

lua/luasnip/config.lua
L3MON4D3: Oh, and in a similar vein, could you shorten the comments in the tests?

L3MON4D3:
Hey, are you still up for this? I could also take over, but I don't understand the `set paste`-part, I'd appreciate if you can explain that to me :sweat_smile:

leiserfg:
I don't get it either, also `paste` is deprecated it may stop working at any point.

L3MON4D3:
Oh good to know, I guess we can ignore that then.
Still, important to know what's going on there, maybe it does not only happen when paste is set, but under other circumstances as well

carbon-steel:
Hi, sorry for dropping out from this PR. It slipped through the cracks when things in life started happening. I'll close this PR for now so as to not leave clutter behind and to signal that anyone should feel free to pick this work up for themselves.

PR #719, opened 2023-01-08
add meta data for each `add_snippets` call

molleweide:
- implements basis for adding meta data to snippets w/new config param.
- improves `available()` to be a bit more versatile.
- adds `jump_to_snip()` under extras

### DIFF
--- a/lua/luasnip/config.lua
+++ b/lua/luasnip/config.lua
@@ -24,6 +24,7 @@ local defaults = {
 	region_check_events = "User None",
 	delete_check_events = "User None",
 	store_selection_keys = nil, -- Supossed to be the same as the expand shortcut
+	store_meta_data = false, -- For each snippet, eg. source path.
 	ext_opts = {
 		[types.textNode] = {
 			active = { hl_group = "LuasnipTextNodeActive" },
--- a/lua/luasnip/config.lua
+++ b/lua/luasnip/config.lua
@@ -25,15 +25,15 @@ local defaults = {
 	delete_check_events = "User None",
 	store_selection_keys = nil, -- Supossed to be the same as the expand shortcut
 
-  -- TODO: generates a function which, given a filename produces the metadata
-  -- that shall badded to snippets loaded by one of the loaders.
-  --
-  -- @params table:
-  --        may contain: filename
-  --        ...
+	-- TODO: generates a function which, given a filename produces the metadata
+	-- that shall badded to snippets loaded by one of the loaders.
+	--
+	-- @params table:
+	--        may contain: filename
+	--        ...
 	store_meta_data = false, -- For each snippet, eg. source path.
 	store_meta_data2 = function(opts)
-	  return {}
+		return {}
 	end,
 
 	ext_opts = {
--- a/lua/luasnip/config.lua
+++ b/lua/luasnip/config.lua
@@ -24,7 +24,18 @@ local defaults = {
 	region_check_events = "User None",
 	delete_check_events = "User None",
 	store_selection_keys = nil, -- Supossed to be the same as the expand shortcut
+
+  -- TODO: generates a function which, given a filename produces the metadata
+  -- that shall badded to snippets loaded by one of the loaders.
+  --
+  -- @params table:
+  --        may contain: filename
+  --        ...
 	store_meta_data = false, -- For each snippet, eg. source path.
+	store_meta_data2 = function(opts)
+	  return {}
+	end,
+
 	ext_opts = {
 		[types.textNode] = {
 			active = { hl_group = "LuasnipTextNodeActive" },
--- a/lua/luasnip/extras/init.lua
+++ b/lua/luasnip/extras/init.lua
@@ -115,10 +115,13 @@ end
 
 -- TODO: MOVE TO OWN MODULE
 local function jump_to_snip(snip)
-	local meta_data = require("luasnip.session.snippet_collection").get_meta_data_by_snip_id(snip.id)
+	local meta_data =
+		require("luasnip.session.snippet_collection").get_meta_data_by_snip_id(
+			snip.id
+		)
 	local source = meta_data.source
-	if  source then
-	  vim.cmd("edit " .. source)
+	if source then
+		vim.cmd("edit " .. source)
 		if snip.name then
 			local feed_str = vim.api.nvim_replace_termcodes(
 				"/" .. snip.name .. "<CR>",
--- a/lua/luasnip/extras/init.lua
+++ b/lua/luasnip/extras/init.lua
@@ -113,6 +113,27 @@ local function match(index, _match, _then, _else)
 	return F(func, index)
 end
 
+local function jump_to_snip(snip)
+	local meta_data = require("luasnip.session.snippet_collection").get_meta_data_by_snip_id(snip.id)
+	local source = meta_data.source
+	if  source then
+	  vim.cmd("edit " .. source)
+		if snip.name then
+			local feed_str = vim.api.nvim_replace_termcodes(
+				"/" .. snip.name .. "<CR>",
+				true,
+				true,
+				true
+			)
+			vim.fn.feedkeys(feed_str, "n")
+		end
+	else
+		print(
+			"No source found for id: " .. snip.id .. " with name: " .. snip.name
+		)
+	end
+end
+
 return {
 	lambda = lambda,
 	match = match,
@@ -161,4 +182,6 @@ return {
 	--alias
 	l = lambda,
 	m = match,
+
+	jump_to_snip = jump_to_snip,
 }
--- a/lua/luasnip/extras/init.lua
+++ b/lua/luasnip/extras/init.lua
@@ -113,6 +113,7 @@ local function match(index, _match, _then, _else)
 	return F(func, index)
 end
 
+-- TODO: MOVE TO OWN MODULE
 local function jump_to_snip(snip)
 	local meta_data = require("luasnip.session.snippet_collection").get_meta_data_by_snip_id(snip.id)
 	local source = meta_data.source
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -63,24 +63,66 @@ local function default_snip_info(snip)
 	}
 end
 
-local function available(snip_info)
+---Retrieve available snippets
+---@param snip_info function: takes a snip and returns whatever props you want
+---         from the snippet
+---
+---@param opts: May contain:
+---         scope <nil|string("buf","everything")|table(filetypes)
+---           buf: returns snippets for current buffers filetypes
+---           everything: returns all loaded snippets
+---           table: allows you to filter snippet subsets by filetype(s)
+---         include_autosnippets bool
+---
+---@return table: the resulting table of ft snip mappings after fitlers have
+--          been applied
+local function available(snip_info, opts)
+	opts = opts or {}
+	opts.scope = opts.scope or "buf"
+	opts.include_autosnippets = opts.include_autosnippets or true
 	snip_info = snip_info or default_snip_info
-
-	local fts = util.get_snippet_filetypes()
 	local res = {}
-	for _, ft in ipairs(fts) do
-		res[ft] = {}
-		for _, snip in ipairs(get_snippets(ft)) do
-			if not snip.invalidated then
-				table.insert(res[ft], snip_info(snip))
-			end
-		end
-		for _, snip in ipairs(get_snippets(ft, { type = "autosnippets" })) do
+
+	local function validate_and_append_to_res(ft, t_snips)
+		for _, snip in ipairs(t_snips) do
 			if not snip.invalidated then
 				table.insert(res[ft], snip_info(snip))
 			end
 		end
 	end
+
+	local function filter_all_snippets(t_snips)
+		for ft, snips in pairs(t_snips) do
+			if
+				opts.scope == "everything" or vim.tbl_contains(opts.scope, ft)
+			then
+				if not res[ft] then
+					res[ft] = {}
+				end
+				validate_and_append_to_res(ft, snips)
+			end
+		end
+	end
+
+	if opts.scope == "buf" then
+		local fts = util.get_snippet_filetypes()
+		for _, ft in ipairs(fts) do
+			res[ft] = {}
+			validate_and_append_to_res(ft, get_snippets(ft))
+			if opts.include_autosnippets then
+				validate_and_append_to_res(
+					ft,
+					get_snippets(ft, { type = "autosnippets" })
+				)
+			end
+		end
+	elseif opts.scope == "everything" or type(opts.scope) == "table" then
+		filter_all_snippets(get_snippets())
+		if opts.include_autosnippets then
+			filter_all_snippets(get_snippets(nil, { type = "autosnippets" }))
+		end
+	end
+
 	return res
 end
 
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -56,6 +56,7 @@ end
 local function get_context(snip)
 	return {
 	  id = snip.id,
+	  meta_data = snip.meta_data,
 		name = snip.name,
 		trigger = snip.trigger,
 		description = snip.dscr,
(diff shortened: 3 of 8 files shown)


molleweide:
should I have squashed all commits?

molleweide:
It seems that the CI removed my additions to the `doc/luasnip.txt` file. [47a669d](https://github.com/L3MON4D3/LuaSnip/pull/719/commits/47a669df92ed92e31bb25598cb36c3e98c227361)

L3MON4D3:
Ah, yeah add doc to `DOC.md`, doc/luasnip.txt is auto-generated from it :sweat_smile:

molleweide:
> Ah, yeah add doc to `DOC.md`, doc/luasnip.txt is auto-generated from it 😅

fixed it

L3MON4D3:
Nice :+1:

lua/luasnip/config.lua
```diff
@@ -24,6 +24,7 @@ local defaults = {
 	region_check_events = "User None",
 	delete_check_events = "User None",
 	store_selection_keys = nil, -- Supossed to be the same as the expand shortcut
+	store_meta_data = false, -- For each snippet, eg. source path.
```
L3MON4D3, line 27: I think I'd like this to be a table, which can accept keys to determine what is stored. Most capable would be just a function which receives all available data and creates the metadata for the snippet, but I think this is overdoing it, since we probably don't need much flexibility there.

Still, making this a table would somewhat future-proof it, like

```lua
metadata = {sourcefile=true, line=true, custom = somefn}
```

lua/luasnip/init.lua
```diff
@@ -55,6 +55,8 @@ end
 
 local function get_context(snip)
 	return {
+		id = snip.id,
+		meta_data = snip.meta_data,
```
L3MON4D3, line 59: Oh, could you rebase on master? this is now a bit different, the `get_context` used by `available` is configurable now, I don't think this is necessary anymore.

lua/luasnip/loaders/init.lua
```diff
@@ -37,6 +57,11 @@ function M.edit_snippet_files(opts)
 		return {}
 	end
 
+	if opts.target_snippet then
+		edit_specific_snippet(opts.target_snippet, edit)
+		return
+	end
+
```
L3MON4D3, line 64: I think this edit specific snippet-funcitonality does not really fit in here: `edit_snippet` has this choice, which doesn't really make sense if all wants to do is edit a specific snippet. How about putting something like this in `extras`?

lua/luasnip/session/snippet_collection.lua
```diff
@@ -270,6 +284,10 @@ function M.add_snippets(snippets, opts)
 			table.insert(by_prio[snip.snippetType][snip.priority][ft], snip)
 			table.insert(by_ft[snip.snippetType][ft], snip)
 			by_id[snip.id] = snip
+
+			if opts.source and opts.store_meta_data then
+				snip["meta_data"] = snippets_meta_data[#snippets_meta_data]
```
L3MON4D3, line 289: One downside of storing a direct reference to that table in the snippet is that it would be copied recursively, which I don't think we want. I feel like an id per add-snippets-call and associating the snippets with that id would be the right choice, but I can't really rationalize

molleweide: Aaah so we just add a table in the middle which maps all ids to the respective meta table and then get meta data by `id_to_meta[snip.id]`

lua/luasnip/loaders/init.lua
molleweide: Sure!

lua/luasnip/init.lua
molleweide: I can rebase ofc but I don't understand the issue - could you explain further. I am on latest master locally.

lua/luasnip/config.lua
molleweide: I think using a subtable is good idea. I don't understand your reasoning about the function.

lua/luasnip/session/snippet_collection.lua
L3MON4D3: I think using `snip.id` might be the easiest way to implement it for now, but yes, essentially that's what I'm getting at :D

lua/luasnip/config.lua
L3MON4D3: Nice, thanks!
My idea with the function was to allow complete customization of what is put into the metadata, essentially by calling a function like

```lua
function loader_metadata(filename, line)
	return {source=filename, source_line = line}
end
```
in the loaders.

Since there isn't this much customization possible (I think), letting the user set some keys for what exactly should be stored in the metadata-table seems like a more straightforward, and easier to use simplification.

lua/luasnip/init.lua
L3MON4D3: Are you sure you're on latest master? I merged a PR this morning, which touches `available`, and I don't see its changes here

L3MON4D3: Or I'm completely missing something xD

molleweide: Ima look into this tomorrow. Im in sweden time zone

lua/luasnip/config.lua
molleweide: Ill try to making it into a func and see how it feels

lua/luasnip/session/snippet_collection.lua
molleweide: I think this is what i did in my first few commits with a getter fn that gets the meta data by snip id

lua/luasnip/init.lua
molleweide: > I merged a PR this morning, which touches `available`, and I don't see its changes here

You were right ofc. I must have pulled right before you pushed...

molleweide: I think that available should be modified again, so that one can pass it 
1. "everything" -> return every single snippet
2. table list of filetypes -> return all requested snippets
3. current version -> return only available to current buffer.

L3MON4D3: Ahh unfortunate

lua/luasnip/session/snippet_collection.lua
L3MON4D3: Nice, I think that that is the best choice for us.
Giving a snippet another id is also a bit superfluous since we can just map id -> other_id

molleweide:
I modified `ls.availalable()` to be a bit more versatile.

Now it works as before if you don't supply the `scope` param, but with the scope param you can choose if you want all possible snips to be returned and trimmed by the snip info func OR you can pass a table list of filetypes which retrieves only the filetype subset.

molleweide:
1. I undid my changes to the `default_snip_info` func now that I pulled your latest changes.
2. I added an `opts` param to the `ls.available()` func so that one can retrieve snippets with a little bit more flexibility.

lua/luasnip/config.lua
molleweide: Now, this is the only thing left to do.

molleweide: I believe that we can have default as a simple `false` and then check `if func` and then run the user meta func, otherwise just default to nothing.

molleweide: and ig also if user sets store meta == true, then store everything?

molleweide: How would one get the source line of the snippet in this function? Is there something builtin to luasnip already?

molleweide: i pushed a basic version

molleweide:
i created a base for using a user function when loading meta data. check it out and tell me what you think.

lua/luasnip/config.lua
L3MON4D3: > How would one get the source line of the snippet in this function? Is there something builtin to luasnip already?

Oh, that's very hard to do for lua and vscode, I used it because I wanted an example for why more flexibility might be good.
I think we should not determine all lines up-front (except if it's trivial, like in the snipmate-loader), since only a tiny subset of this information will be useful, and it should, I think, be calculated on-demand

L3MON4D3: Maybe not very hard, non-trivial

molleweide: > Maybe not very hard, non-trivial

Aight, for now my feed keys thingy works good nuff

molleweide:
Also feel free to try my picker that I linked to under the issue discussion. it shows you how I use all of the features in the `available()` func.

lua/luasnip/config.lua
L3MON4D3: Right! heuristics like that should do it for 99%

L3MON4D3:
> 1. I undid my changes to the `default_snip_info` func now that I pulled your latest changes.
> 2. I added an `opts` param to the `ls.available()` func so that one can retrieve snippets with a little bit more flexibility.
> 
> Now you can either:
> 
> 1. retrieve snips for buf
> 2. .... all snips
> 3. .... only a subset of filtypes if you pass `opts.scope = { "ft1", "ft2" }`

I think we can use `get_snippets` for this usecase? (Don't really want to tack more onto `available`, we kind of messed up its design by having it take an optional argument directly, and not in an `opts`-table🤦)

L3MON4D3:
I'll take a look at the other changes later😅

molleweide:
> I'll take a look at the other changes later😅

Aight no stress!

molleweide:
> I think we can use `get_snippets` for this usecase? (Don't really want to tack more onto `available`, we kind of messed up its design by having it take an optional argument directly, and not in an `opts`-table🤦)

Cool yeah im down for whatever. I'll wait untill you've had a closer look and then I modify it accordingly.

lua/luasnip/extras/init.lua
```diff
@@ -113,6 +113,30 @@ local function match(index, _match, _then, _else)
 	return F(func, index)
 end
 
+local function jump_to_snip(snip)
```
L3MON4D3, line 116: I think this deserves its own module, then we can cleanly provide different heuristics for finding the correct snippet (after making that configurable ofc :D)
For example, finding in snipmate/lua/vscode could be made more precise by writing it specifically for that type of snippet

lua/luasnip/init.lua
```diff
@@ -63,24 +63,66 @@ local function default_snip_info(snip)
 	}
 end
 
-local function available(snip_info)
```
L3MON4D3, line 66: Okay afaict this can also be achieved with the old available (scope=buf) and get_snippets (everything else). Even though this design is nice, separate functions are backward compatible, so I think we should stick with that

lua/luasnip/session/snippet_collection.lua
```diff
@@ -128,6 +130,33 @@ local by_id = setmetatable({}, {
...
 })
 
+-- If `add_snippets` opts.source and config.store_meta_data then each call to
+-- add snippets stores the source file and potentially other meta data in this
+-- table which then each snippet points to so that one can `jump-to-snip`
+-- by calling `loaders.edit_snippet_files({ target_snippet = snip_context_x })`
+local snippets_meta_data = {}
+local snip_id_to_meta_map = {}
+
+local function initialize_snippet_meta_data(opts)
```
L3MON4D3, line 140: Okay, I think this mechanism is good, but we need to split it up a bit:
* `config` generates a function which, given a filename produces the metadata that shall be added to snippets loaded by one of the loaders. We should not pass the filename directly, but wrap it in a table like `{filename=<the filename>}`, since we can then add more arguments and the function will stay comfortable to use (no need to write a handler like `function(_, _, _, theOneArgINeed) ... end`)
* this function is used in loaders to generate the metadata for the snippets (no ifs or something like that, we can just call the function. This is flexible enough to probably not require any changes, except if new parameters are added)
* `add_snippets` receives `metadata`-table, which is associated with the snippets added in this call, unconditionally

lua/luasnip/extras/init.lua
molleweide: ok. i'll move this into its own module.

lua/luasnip/init.lua
molleweide: quick note: my addition should be backwards compatible.
I'll get back with another reply later.

lua/luasnip/session/snippet_collection.lua
molleweide: Ok, I have to ponder your reply and come back later with a revised version.

lua/luasnip/init.lua
molleweide: Allright, so I think I understand what you mean but could you tell me a bit more specifically what you'd like me to do. Is my modification worth keeping in a modified state or should we scrap it completely?

L3MON4D3: True, of course it is backward compatible :facepalm:
What I don't like about this is that there's no good reason for `snip_info` to not be a part of `opts`, except staying backward compatible.

I think we'll have to scrap it :/
An alternative would be to expose it in addition to `available` (and implement `available` using it), but all of its functionality can be covered better by `get_snippets` (since get_snippets can access the implementation of `snippet_collection`)

lua/luasnip/session/snippet_collection.lua
L3MON4D3: Looking forward to it :D
If you have any questions or other ideas, go ahead, I'd love to find something better (especially the association of snippet with its metadata is a bit annoying, since we have to do that id-redirection. Can't think of a better way though :/)

lua/luasnip/init.lua
molleweide: Allright yeah I agree. I'll fix this.

molleweide: > An alternative would be to expose it in addition to available (and implement available using it), but all of its functionality can be covered better by get_snippets (since get_snippets can access the implementation of snippet_collection)

Yeah this makes sensr. I'll fix this.

lua/luasnip/session/snippet_collection.lua
molleweide: Cool, yeah i cant really come up with something better either atm.

L3MON4D3:
Hey!
I just found out about `debug.getinfo`, we could use this to accurately get the lines a snippet is defined on (for lua!!!).
With this, and snipmate working nicely (ie. we can get the lines the snippet is defined on), I have to admit that snippet-specific metadata is probably the way to go, and wouldn't mind just setting `snippet.metadata` (though I'd still like an API through `snippet_collection`, that would allow us to replace this later on)

Are you still interested in working on this? No worries if not, I'd be up for getting this done

L3MON4D3:
Was feeling a bit too motivated the last few days, couldn't resist doing some work on this :sweat_smile:
Check #826, I'd love to hear your thoughts :D

molleweide:
Hey mate!!
Wow, I gotta checkout what you've done. I haven't spent a lot of time on this since we last spoke - i just haven't had time + plus too many macos issues with vim so there has been some other fires ive had to try to put out 🙄 (i have to migrate over to linux asap)
I'll look at what you did and see if I can come with some valid feedback. I'm really excited to hear that you got something good working. :)

L3MON4D3:
> I haven't spent a lot of time on this since we last spoke - i just haven't had time + plus too many macos issues with vim so there has been some other fires ive had to try to put out 🙄 (i have to migrate over to linux asap)

Phuh, that's rough😅

> I'll look at what you did and see if I can come with some valid feedback. I'm really excited to hear that you got something good working. :)

👍👍

> Dude, you added the feature where one can jump to the source from within an opened snippet - that is pretty damn cool.

Yeah! Really happy that worked out so well :D

molleweide:
> Yeah! Really happy that worked out so well :D

Has it already improved your workflow?

L3MON4D3:
Nah xD
But it was fun to write!
And it feels like such a nice addition, I'm sure it'll come in handy sometimes

leiserfg:
> But it was fun to write!

Reminds me that after all the work in LuaSnip, I have just like 5 custom snippets :smile:

L3MON4D3:
> Reminds me that after all the work in LuaSnip, I have just like 5 custom snippets :smile:

:laughing::laughing:
You're just using friendly-snippets, right? I guess that +lsp covers almost anything one would practically need, most snippets beyond that are just eye(/finger?/brain? :D)-candy

molleweide:
> But it was fun to write!

lol

L3MON4D3:
I've finally merged #826, closing this as well.
Sorry we couldn't use your work :/

PR #727, opened 2023-01-17
vscode snippet loader ignores paths after invalid path 

bootleq:
Hi, the loader with specific `paths` might behave unexpectedly, where first **invalid** path make further items ignored.
For example:

```lua
require('luasnip.loaders.from_vscode').load({
  paths = {
    '/NOT/EXIST/PATH',  -- this directory doesn't exist
    '/path/to/friendly-snippets',
  },
})
```
should still find `friendly-snippets` path, while it doesn't, all paths were ignored and not loaded.

Can also reproduce in my forked test, thanks.


L3MON4D3:
Thank you for opening issue+solution, very nice!!
I think this is an okay solution, but I'd rather have the list passed to deduplicate not have missing items (such that it is continuous and can be iterated using `ipairs`).
Would you look into making that change instead?

bootleq:
Thanks for suggestion, I've changed the implementation.
Also found other loaders (lua, snipmate) have the same problem, so handled them, too.

If the changes look ok, I can to do a rebase if needed.

L3MON4D3:
Looks perfect :+1:
Squashing into one commit seems appropriate, would be great if you could do that

L3MON4D3:
No more changes, this is great :D
Thank you very much for the fix!!

PR #745, opened 2023-01-24
Fix dot expansion in path

leiserfg:


L3MON4D3:
Nice, yeah that looks good :+1:

PR #777, opened 2023-02-17
refactor with some newer apis

max397574:


max397574:
@L3MON4D3 are those test failures from my changes?

lua/luasnip/config.lua
```diff
@@ -214,29 +215,54 @@ c = {
...
-		)
+		local augroup = vim.api.nvim_create_augroup("luasnip", {})
+		vim.api.nvim_create_autocmd(session.config.delete_check_events, {
+			callback = function()
+				require("luasnip").unlink_current_if_deleted()
+			end,
+			group = augroup,
+		})
+		vim.api.nvim_create_autocmd(session.config.update_events, {
+			callback = function()
```
leiserfg, line 226: Here you can pass `require("luasnip").active_update_dependents` directly without a function wrapper, the same in the other ones.

max397574: I prefer it like this because here are two cases where I can and 4 where I can't and I prefer to keep it the same. @L3MON4D3 what should I do?

L3MON4D3:
> @L3MON4D3 are those test failures from my changes?

I think so.. they can (should be able to, at least) also run locally, via `make test`. Could you try if that works?

lua/luasnip/config.lua
L3MON4D3: It doesn't matter much, if you like it keep it in there (though without seems a bit better, in a "nothing left to take away"-sense)

max397574:
I can't run the tests because I get an error when compiling the dependencies (I have no clue what this is xD)

L3MON4D3:
Okay, I can take a closer look later.
What's the error, though? Maybe it's a quick fix😅

lua/luasnip/config.lua
```diff
@@ -214,29 +215,54 @@ c = {
...
+				require("luasnip").exit_out_of_region(
+					require("luasnip").session.current_nodes[vim.api.nvim_get_current_buf()]
+				)
+			end,
+			group = augroup,
+		})
+		-- Remove buffers' nodes on deletion+wipeout.
+		vim.api.nvim_create_autocmd({ "BufDelete", "BufWipeout" }, {
+			callback = function()
+				current_nodes = require("luasnip").session.current_nodes
```
leiserfg, line 242: @L3MON4D3  I think this should be `local` (which is wrong in the original as well)

max397574: tru I also thought that
but I wasn't sure if there is a global created somewhere else as well

L3MON4D3:
Taking a full look at the code, I think it would be easier to read if we abstract the `vim.api.nvim_create_autocmd(event, {callback=..., group = augroup})` to a function `ls_autocmd(event, callback)` (locally inside `_setup`)

max397574:
done

max397574:
anything else that should be done @L3MON4D3

L3MON4D3:
I'm taking a look at the tests right now, they're acting up and I can't quite figure out why :/
But I don't think there's anything left, I did a few small changes.. I'll push them, then you can take a look at those, but then, apart from fixing the tests, nothing (at least code-wise, we'll need to add a breaking-change notice in the commit-messages, but let's wait with that until I've got my changes in :D)

max397574:
I'd give the function a shorter name

L3MON4D3:
Oh, right, that's an artifact from when I though `remap`/`noremap` was causing issues.
I'll change that

L3MON4D3:
Okay soo `"User None"` as the default-event obviously no longer works, seems like a race-condition caused the tests to fail (autocommand and lua-call modify snippet simultaneously => bad) (not completely sure what the autocommands are registered for if `"User None"` is passed as the event, but nothing good, apparently xD)

I'm on the fence on how to handle this problem: if someone had `"User <something>"` in their `setup`, this now acts differently from before, BUT I don't really think anyone has that in their config. Leaning towards only putting a warning into the breaking change notice.

Oh, and I missed that the doc should be updated now that `*_events` also accepts a table :/
Would you do that, and make proper commits (including the breaking change notice)? Your first commits are nice, but my last commits have to be split up a bit :|
If you don't want to deal with that, I can do that later, too, but I'll take a break now :D

max397574:
I probably won't have time to do it in the next two days (perhaps also more)
so if you want to get this merged sooner you should do it

L3MON4D3:
Okay :+1:

L3MON4D3:
@leiserfg wanna take a look?

leiserfg:
On it

doc/luasnip.txt
```diff
@@ -2761,52 +2761,56 @@ print a short message to the log.
...
 - `ext_opts`: Additional options passed to extmarks. Can be used to add
     passive/active highlight on a per-node-basis (more info in DOC.md)
 - `parser_nested_assembler`: Override the default behaviour of inserting a
-    `choiceNode` containing the nested snippet and an empty `insertNode` for
-    nested placeholders (`"${1: ${2: this is nested}}"`). For an example
-    (behaviour more similar to vscode), check
-    here <https://github.com/L3MON4D3/LuaSnip/wiki/Nice-Configs#imitate-vscodes-behaviour-for-nested-placeholders>
+    `choiceNode` containing the nested snippet and an empty `insertNode` for nested
+    placeholders (`"${1: ${2: this is nested}}"`). For an example (behaviour more
+    similar to vscode), check here
```
leiserfg, line 2793: There is an extra `)` after vscode

tests/helpers.lua
```diff
@@ -29,12 +29,13 @@ function M.session_setup_luasnip(opts)
...
 	helpers.exec("set rtp+=" .. os.getenv("LUASNIP_SOURCE"))
 	helpers.exec(
-		string.format(
-			"source %s/plugin/luasnip.vim",
-			os.getenv("LUASNIP_SOURCE")
-		)
+		("source %s/plugin/luasnip.vim"):format(os.getenv("LUASNIP_SOURCE"))
+	)
+	helpers.exec(
+		("luafile %s/plugin/luasnip.lua"):format(os.getenv("LUASNIP_SOURCE"))
```
leiserfg, line 38: You don't need to change it, but it's also valid to use `source` here.

leiserfg: The rest seems "good to go".

L3MON4D3: Ah, right!
Yeah, I'll leave it like this :+1:

L3MON4D3:
Okay perfect, thank you :+1:

L3MON4D3:
Thank you for initiating this, and making the majority of the changes so quickly @max397574 :)

PR #778, opened 2023-02-18
Remove vim.fn.json fallbacks

leiserfg:


leiserfg:
Now that we require 0.7 is safe to remove them

L3MON4D3:
Right! Good catch👌

PR #811, opened 2023-03-08
Fix a bug where the loading from lua with an invalid content is ignored

### LINKED ISSUE #810: Bug when from_lua snippets are loaded with incorrectly formatted content
When loading a snippet that is incorrectly formatted, the error catching is bypassed ass it is checking the wrong value.

To reproduce:

Add this to your luasnip config:

```
require("luasnip.loaders.from_lua").load({paths = "~/snippets"})
```

Then add this to all.lua in that directory:

```
{
  "Standard Starter Template": {
    "prefix": "sst",
    "body": [
      "#include <stdlib.h>",
      "#include <stdio.h>",
      "",
      "int main(int argc, char *argv[])",
      "{$0",
      "",
      "\treturn EXIT_SUCCESS;",
      "}"
    ],
    "description": "Standard starter template for a tiny C program"
}
}
```

Open NVIM and see the error:

```
Error detected while processing /Users/dwain/.config/nvim/after/plugin/luasnip.lua:                                                                                                   
E5113: Error while calling lua chunk: ...ck/packer/start/LuaSnip/lua/luasnip/loaders/from_lua.lua:77: bad argument #1 to 'setfenv' (number expected, got nil)  
```

dwainm:
Fixes #810

### DIFF
--- a/lua/luasnip/util/parser/ast_utils.lua
+++ b/lua/luasnip/util/parser/ast_utils.lua
@@ -173,45 +173,6 @@ local function replace_position(ast, p1, p2)
 	end)
 end
 
-local is_interactive
-local has_interactive_children = function(node, root)
-	-- make sure all children are not interactive
-	for _, child in ipairs(node.children) do
-		if is_interactive(child, root) then
-			return false
-		end
-	end
-	return true
-end
-local type_is_interactive = {
-	[types.SNIPPET] = has_interactive_children,
-	[types.TEXT] = util.no,
-	[types.TABSTOP] = function(node, root)
-		local tabstop_is_copy = false
-		predicate_ltr_nodes(root, function(pred_node)
-			-- stop at this tabstop
-			if pred_node == node then
-				return true
-			end
-			-- stop if match found
-			if pred_node.tabstop == node.tabstop then
-				tabstop_is_copy = true
-				return true
-			end
-			-- otherwise, continue.
-			return false
-		end)
-		-- this tabstop is interactive if it is not a copy.
-		return not tabstop_is_copy
-	end,
-	[types.PLACEHOLDER] = has_interactive_children,
-	[types.VARIABLE] = util.no,
-	[types.CHOICE] = util.yes,
-}
-local function is_interactive(node, snippet)
-	return type_is_interactive[node.type](node, snippet)
-end
-
 function M.fix_zero(ast)
 	local zn, ast_child_with_0_indx, is_copied = real_zero_node(ast)
 	-- if zn exists, is a tabstop, an immediate child of `ast`, and does not
--- a/lua/luasnip/util/util.lua
+++ b/lua/luasnip/util/util.lua
@@ -547,7 +547,7 @@ local function no()
 end
 
 local function yes()
-	return false
+	return true
 end
 
 local function reverse_lookup(t)


lua/luasnip/loaders/from_lua.lua
```diff
@@ -34,9 +34,9 @@ local function load_files(ft, files, add_opts)
 		local func_string = path_mod.read_file(file)
 
 		local load_ok, func = pcall(loadstring, func_string)
-		if not load_ok then
+		if not load_ok or func == nil then
 			log.error("Failed to load %s\n: %s", file, func)
-			error("Failed to load " .. file .. "\n: " .. func)
+			error(string.format("Failed to load %s\n: %s", file, func))
```
L3MON4D3, line 39: You can omit the `string.format` and just pass `file, func` to `error`, [it passes its arguments to `string.format`](https://github.com/L3MON4D3/LuaSnip/blob/45670282dc861b51c4e028e6930c75371fe0b88f/lua/luasnip/util/log.lua#L102-L106)

L3MON4D3: Yes :+1:
I'm not saying you should not use `string.format`, I'm saying that `string.format` is called with the arguments passed to `error`, and there's no need to call that manually before passing the string to `error`.

lua/luasnip/loaders/from_lua.lua
```diff
@@ -34,9 +34,9 @@ local function load_files(ft, files, add_opts)
 		local func_string = path_mod.read_file(file)
 
 		local load_ok, func = pcall(loadstring, func_string)
```
L3MON4D3, line 36: But actually, now it would be best to just do `load_ok, func = loadfile(file)`, afaict this should work without any other changes.

L3MON4D3: (Didn't do this before since we used to modify `func_string`, and I missed this when removing that)

dwainm: That works. Updating.

dwainm: I've updated the code to use loadfile in https://github.com/L3MON4D3/LuaSnip/commit/c6de1243b875cf8927c51843f13370d549fdb156

dwainm: Somehow it regressed. Load file load ok keeps being null. I'm going to revert to the old method.

dwainm:
@L3MON4D3 let me know if you need any other changes here.

lua/luasnip/loaders/from_lua.lua
L3MON4D3: Aaah sorry, my bad, I thought that was a `log.error` 🤦🤦🤦

lua/luasnip/loaders/from_lua.lua
L3MON4D3: Right, ofc🤦
It would have to be `func, err_msg = loadfile...`, it does not match pcall exactly :/

dwainm: Right 🙈 , forgive me, it was late, and I assumed matching API's. It's working fine with loadfile: https://github.com/L3MON4D3/LuaSnip/pull/811/commits/c8bc64b9225b4a8253cb8595f5842388fb6b95b9

L3MON4D3: same xD

L3MON4D3:
Could you squash all of this into one commit?

leiserfg:
> Could you squash all of this into one commit?

You can do that while merging via github UI.

L3MON4D3:
Technically yes, but I dislike that there's no way to confirm it actually worked how I wanted it to :/

dwainm:
@L3MON4D3 please see https://github.com/L3MON4D3/LuaSnip/pull/820. I had some issues with the rebase push. Seem like GH likes fewer changes :)

L3MON4D3:
github :roll_eyes: :P

PR #820, opened 2023-03-11
Fix a loader bug: from_lua

### LINKED ISSUE #810: Bug when from_lua snippets are loaded with incorrectly formatted content

dwainm:
This fixes #810 

Fix an issue where loading from lua with invalid content provides an incorrect error message. This change also uses loadfile instead of loadstring.

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*            For NVIM v0.8.0            Last change: 2023 March 08
+*luasnip.txt*            For NVIM v0.8.0            Last change: 2023 March 11
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Yes, looks perfect :+1:
Thank you!

PR #826, opened 2023-03-16
Generate source for loaded snippets.

L3MON4D3:
This is essentially another approach at storing a snippets source, less generic than #719, and only focused on source, not general metadata.

Major differences:
* store data per-snippet (map id -> data). I think this is the right approach because we can get the line a snippet is defined on for lua and snipmate, and I'd say that makes the additional required storage worth it
* less generic. Can only store source for snippets, but not any metadata. I think this is fine, since I'm not sure what other metadata might be interesting. If I'm wrong, and there is more useful metadata to attach, we could implement source using a more generic interface, later.

I've also written a pretty cool consumer for the metadata:

https://user-images.githubusercontent.com/41961280/225597134-8a2c01a4-ac1f-4ccb-9197-32ef753de969.mp4

It finds the range of the snippet via treesitter for lua and vscode, and fully via attached source-data (we know start and end of the snippet-definition since we do the parsing fully ourselves, and can thus provide this info) for snipmate

Missing:
- [x] Documentation
- [x] Tests


hinell:
I think we need better explanation what's going on tha video. Do you generate new snippet from selection or jump to the snippet definition? I'm puzzled. :roll_eyes:

L3MON4D3:
😅😅
I jump to the source of the current snippet

PR #838, opened 2023-03-22
Keyed nodes

L3MON4D3:
This adds a way to address some node in the expanded snippet, both at runtime (for example when evaluating a dynamic/functionNode-function) and when passing the nodes they should depend on.

For example:
```lua
ls.setup_snip_env()

local k = require("luasnip.nodes.key_indexer").new_key

ls.snip_expand(s("a", {
	-- set key here
	i(1, "preset", {key="outer_i1"}),
	c(2, {
		t"some text",
		{
			i(1), t"repeated: ", f(function(args)
				return args[1]
			-- reference node via key!!
			end, k("outer_i1"))
		}
	})
}), {
	key = "adsf"
})
```

or, a bit more complicated:
```lua
ls.setup_snip_env()

local k = require("luasnip.nodes.key_indexer").new_key

ls.snip_expand(s("a", {
	i(1, "preset", {key="outer_i1"}),
	c(2, {
		t"some text",
		{
			i(1), t"repeated: ", f(function(_, snip)
				-- map nodes inside add_sNode to their text.
				local node_contents = vim.tbl_map(function(node)
					return node:get_text()[1]
				end, snip:get_keyed_node("add_sNode").insert_nodes)

				-- add numbers together.
				return tostring(node_contents[1] + node_contents[2] + node_contents[3])
			end, k("outer_i1"))
		}
	}),
	sn(3, {
		t" ", i(1, "1"), t" ", i(2, "2"), t" ", i(3, "3")
	}, {key="add_sNode"})
}), {
	key = "adsf"
})
```

I believe `key_indexer` is an improvement upon `absolute_indexer`, since that required specifying the "address" to some node as a sequence of jump-indices, whereas `key_indexer` allows much clearer addresses.

Needs:
- [x] Doc
- [x] Tests


L3MON4D3:
Okay this somehow blindsided me completely: with this, it's now also possible to depend on a `functionNode`!
(as usual, not completely sure how it can be useful, but want to give it a tad more visibility, thus this comment :D)

bew:
Thank you for this! 👌

L3MON4D3:
Glad it helps you :)

PR #842, opened 2023-03-26
Add nix support to Makefile

chuwy:
Sorry, I never wrote Makefiles and maybe what I did is non-sense (or I broke non-nix installs), but it now works on my non-brew macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,10 +27,14 @@ nvim: | ${NVIM_PATH}
 	git -C ${NVIM_MASTER_PATH} fetch --depth 1
 
 OS:=$(shell uname)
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Looks good :+1: (I think at least, but I've no experience with nix, so can't comment on its various invocations.. @leiserfg would you take a look?)
I'm not sure why the tests are failing, they work just fine on my machine... Maybe some kind of race condition. I'll investigate (aka re-run the action and hope for the best)

L3MON4D3:
I guess non-nix oses should handle `-lluajitxx` just fine withouth passing `-L` explicitly, so it should be fine to pass this only in the nix-case.
Thank you for the addition @chuwy!

PR #852, opened 2023-03-29
Improve rgrav-adjustments

L3MON4D3:
The previous implementation was a bit messy, this is much more focused and faster.

Needs some more tests, to make sure everything is actually working consistently.


L3MON4D3:
(tests fail due to https://github.com/neovim/neovim/issues/23734, will just wait for the upstream-fix, and not modify tests for this)

PR #860, opened 2023-04-09
fix: `expand_or_locally_jumpable()` checks the next node

loichyan:
`expand_or_locally_jumpable()` should always check whether the next node is jumpable.


L3MON4D3:
Oh!
Didn't spot that the default is -1, thank you :+1:

PR #866, opened 2023-04-13
doc: mention MacOS Makefile variable for jsregexp

fejzuli:
The LUAJIT_OSX_PATH variable which allows MacOS users to specify the location of their lib directory was added in a835e3d680c5940b61780c6af07885db95382478
and needed to be documented.


L3MON4D3:
Right, of course
Thank you!

PR #868, opened 2023-04-16
feat (ffmt): introduce `ffmt` to read format string from separate file

pysan3:
Hi, thanks a lot for the amazing plugin!

I would like to add a functionality to define the format string in a different file and use the content to expand a snippet.
I believe this would be useful to define a template file.

I have implemented a function `file_format_nodes`, aliased to `ffmt` that is kind of a wrapper around `fmt` node.

So the usage is very simple.

### Old way: Using `fmt` node

```lua
snippets[#snippets + 1] = s(
  e("journal", "template for journal page"),
  fmt(
    [[
* {title}
  {{:{yesterday}:}}[Yesterday] - {{:{tomorrow}:}}[Tomorrow]

** Daily Review
   - {}

** Today's Checklist
   - ( ) Write my daily review

** Achievements
   -
    ]],
    {
      title = f(function(_, _)
        return parse_date(0, file_title())
      end),
      yesterday = f(function(_, _)
        return parse_date(-1, file_title())
      end),
      tomorrow = f(function(_, _)
        return parse_date(1, file_title())
      end),
      i(0),
    }
  )
)
```

### Instead, using `ffmt`

- `~/norg-template/journal.norg`
```txt
* {title}
  {{:{yesterday}:}}[Yesterday] - {{:{tomorrow}:}}[Tomorrow]

** Daily Review
   - {}

** Today's Checklist
   - ( ) Write my daily review

** Achievements
   -
```

```lua
snippets[#snippets + 1] = s(
  e("journal", "template for journal page"),
  ffmt("~/norg-template/journal.norg", {
    title = f(function(_, _)
      return parse_date(0, file_title())
    end),
    yesterday = f(function(_, _)
      return parse_date(-1, file_title())
    end),
    tomorrow = f(function(_, _)
      return parse_date(1, file_title())
    end),
    i(0),
  }
)
```

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*            For NVIM v0.8.0            Last change: 2023 April 14
+*luasnip.txt*            For NVIM v0.8.0            Last change: 2023 April 16
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
--- a/lua/luasnip/extras/fmt.lua
+++ b/lua/luasnip/extras/fmt.lua
@@ -222,10 +222,44 @@ end
 
 extend_decorator.register(format_nodes, { arg_indx = 3 })
 
+-- Use file that contains format string with placeholders to interpolate nodes.
+--
+-- See `interpolate` documentation for details on the format.
+--
+-- Arguments:
+--   path: path to file that contains format string
+--   nodes: snippet node or list of nodes
+--   opts: optional table
+--     trim_empty: boolean, remove whitespace-only first/last lines, default true
+--     dedent: boolean, remove all common indent in `str`, default true
+--     ... the rest is passed to `interpolate`
+-- Returns: list of snippet nodes
+local function file_format_nodes(path, nodes, opts)
+	local uv = vim.loop
+	local abs_path = vim.fn.fnamemodify(path, ":p")
+	local st = uv.fs_stat(abs_path)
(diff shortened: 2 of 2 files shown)


leiserfg:
I see how this can be useful for you, but if we add this then it will make sense to do the same for the other formats (which is not very clean IMHO). Maybe a helper function for reading a file that you can use as `fmt(read_file(path), ....)` will be more flexible.

pysan3:
At least my code has nothing to do with formats and simply treats the content of the file as a string. And I don't think we need to care anything about what file types we are dealing and just let people put `{key_name}` inside the files so that the snippet engine can handle them.

The reason why I would like to avoid the helper function route is that since we define snippets for different file types, I would then have to copy that function to all files I want this which would make my code a bit harder to maintain.

I am not thinking to use this only for `norg` files but with other filetypes as well.

L3MON4D3:
Making it easier to define filetype-templates with luasnip seems cool, but I'm not convinced `ffmt` is the right approach, since it requires that the placeholders in the fmt-file and the nodes provided in the call are in sync.

Have you considered parsing lsp-style snippets from a file? Parsed snippets are of course less capable than their lua counterparts, but some of that could be alleviated with [custom variables](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#environment-namespaces)

pysan3:
Thanks for your opinions.

Alright, I'll simply add code on my end.

Closing ;)

PR #876, opened 2023-04-28
refactor: remove deprecated function

huylg:
hi @L3MON4D3,

vim.pretty_print is deprecated so I have changed it to vim.print

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*            For NVIM v0.8.0            Last change: 2023 April 23
+*luasnip.txt*            For NVIM v0.8.0            Last change: 2023 April 28
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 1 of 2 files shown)


huylg:
note: tested at nvim nightly 0.10

leiserfg:
This is deprecated since 0.9 (current stable) so it's fine to change it in the master branch.

L3MON4D3:
Ah, thank you :+1:
Actually, right now we have compatibility down to nvim0.7 (and I don't think we need to break it for this), could you add something like [this](https://github.com/L3MON4D3/LuaSnip/blob/f3bdf8cf535b1df6f7c0122d963c6546199aa0a3/lua/luasnip/extras/snip_location.lua#L4-L9), such that 0.7 keeps working?

L3MON4D3:
(I'll have to add a test later to catch this, they should be failing on this PR :grimacing:)

L3MON4D3:
Superseded by #880😅

PR #889, opened 2023-05-20
fix(docs): type annotation can by nil

kunzaatko:


L3MON4D3:
Right, thank you :+1:

PR #894, opened 2023-05-25
Support `.code-snippets`.

L3MON4D3:
Both in `package.json`-style snippet-packages and as standalone files, where snippet-filetypes have to be set via `scope`.

### DIFF
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -132,7 +132,8 @@ local function load_snippet_files(lang, files, add_opts)
 				-- only load snippets matching the language set in `package.json`.
 				file_lang_snippets,
 				vim.tbl_extend("keep", {
-					-- again, include filetype, same reasoning as with augroup.
+					-- provide key s.t. snippets will be replaced if the file
+					-- is reloaded.
 					key = string.format("__%s_package_snippets_%s", lang, file),
 					refresh_notify = false,
 				}, add_opts)
@@ -336,7 +337,11 @@ function M.edit_snippet_files()
 end
 
 local function standalone_add(path, add_opts)
+	-- by default, put snippets into global scope.
 	local file_snippets = get_file_snippets(path, "all")
+
+	-- store add_opts for this file: file might be edited, and then we want to
+	-- reload it with these same add_opts.
 	standalone_cache.path_snippets[path] = {
 		add_opts = add_opts
 	}
@@ -346,6 +351,8 @@ local function standalone_add(path, add_opts)
 		nil,
 		file_snippets,
 		vim.tbl_extend("keep", {
+			-- provide key s.t. snippets will be replaced if the file is
+			-- reloaded.
 			key = string.format("__standalone_snippets_%s", path),
 		}, add_opts)
 	)
@@ -375,6 +382,7 @@ function M._reload_file(filename)
 		end
 		ls.clean_invalidated({ inv_limit = 100 })
 	end
+
 	local standalone_cached_data = standalone_cache.path_snippets[filename]
 	if standalone_cached_data then
 		log.info("Re-loading snippets contributed by %s", filename)
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -48,7 +48,7 @@ local function get_file_snippets(file, default_filetype, opts)
 	local ignore_scope = util.ternary(opts.ignore_scope, opts.ignore_scope, false)
 
 	-- since most snippets we load don't have a scope-field, we just insert this here by default.
-	local snippets_by_ft = {[default_filetype] = {snippets = {}, autosnippets = {}}}
+	local snippets_by_ft = {[default_filetype] = {}}
 
 	local snippet_set_data = read_json(file)
 	if snippet_set_data == nil then
@@ -74,6 +74,7 @@ local function get_file_snippets(file, default_filetype, opts)
 					dscr = parts.description or name,
 					wordTrig = ls_conf.wordTrig,
 					priority = ls_conf.priority,
+					snippetType = ls_conf.autotrigger and "autosnippet" or "snippet"
 				}, body)
 
 				if session.config.loaders_store_source then
@@ -88,16 +89,15 @@ local function get_file_snippets(file, default_filetype, opts)
 				-- scope can have multiple components.
 				if parts.scope and not ignore_scope then
 					for _, scope in ipairs(vim.split(parts.scope, ",", { plain = true })) do
-						snippets_by_ft[scope] = snippets_by_ft[scope] or {snippets = {}, autosnippets = {}}
+						snippets_by_ft[scope] = snippets_by_ft[scope] or {}
 						table.insert(ft_tables, snippets_by_ft[scope])
 					end
 				else
 					table.insert(ft_tables, snippets_by_ft[default_filetype])
 				end
 
-				local snippet_type = ls_conf.autotrigger and "autosnippets" or "snippets"
 				for _, t in ipairs(ft_tables) do
-					table.insert(t[snippet_type], snip)
+					table.insert(t, snip)
 				end
 			end
 		end)
@@ -129,27 +129,16 @@ local function load_snippet_files(lang, files, add_opts)
 			ls.add_snippets(
 				lang,
 				-- only load snippets matching the language set in `package.json`.
-				file_lang_snippets.snippets,
+				file_lang_snippets,
 				vim.tbl_extend("keep", {
-					type = "snippets",
 					-- again, include filetype, same reasoning as with augroup.
 					key = string.format("__%s_snippets_%s", lang, file),
 					refresh_notify = false,
 				}, add_opts)
 			)
-			ls.add_snippets(
-				lang,
-				file_lang_snippets.autosnippets,
-				vim.tbl_extend("keep", {
-					type = "autosnippets",
-					key = string.format("__%s_autosnippets_%s", lang, file),
-					refresh_notify = false,
-				}, add_opts)
-			)
 			log.info(
-				"Adding %s snippets and %s autosnippets for filetype `%s` from %s",
-				#file_lang_snippets.snippets,
-				#file_lang_snippets.autosnippets,
+				"Adding %s snippets for filetype `%s` from %s",
+				#file_lang_snippets,
 				lang,
 				file
 			)
--- a/lua/luasnip/loaders/from_vscode.lua
+++ b/lua/luasnip/loaders/from_vscode.lua
@@ -338,7 +338,6 @@ end
 local function standalone_add(path, add_opts)
 	local file_snippets = get_file_snippets(path, "all")
 	standalone_cache.path_snippets[path] = {
-		snippets = vim.deepcopy(file_snippets),
 		add_opts = add_opts
 	}
 
@@ -379,8 +378,9 @@ function M._reload_file(filename)
 	local standalone_cached_data = standalone_cache.path_snippets[filename]
 	if standalone_cached_data then
 		log.info("Re-loading snippets contributed by %s", filename)
-		package_cache.path_snippets[filename] = nil
-		local add_opts = package_cached_data.add_opts
+
+		standalone_cache.path_snippets[filename] = nil
+		local add_opts = standalone_cached_data.add_opts
 
 		standalone_add(filename, add_opts)
 		ls.clean_invalidated({ inv_limit = 100 })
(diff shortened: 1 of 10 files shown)


hinell:
@L3MON4D3 I would say this is rather a rough solution...  What's the purpose of the cache in short?

Lack of comments for that code actually prevents any sensible contributions.

L3MON4D3:
Oh, why would you say this is rough? And where do you more comments would help?
Good point about the cache, I forgot to put in the part where it actually re-uses the snippets it loaded previously (although, maybe that's not actually necessary for `.code-snippets`, I guess loading the same file twice shouldn't occur, and is most likely an error)

lua/luasnip/loaders/from_vscode.lua
```diff
@@ -67,69 +76,76 @@ local function get_file_snippets(file)
...
+						snippets_by_ft[scope] = snippets_by_ft[scope] or {}
+						table.insert(ft_tables, snippets_by_ft[scope])
+					end
 				else
-					table.insert(lang_snips, snip)
+					table.insert(ft_tables, snippets_by_ft[default_filetype])
+				end
+
+				for _, t in ipairs(ft_tables) do
+					table.insert(t, snip)
```
hinell, line 105: Too puzzling. Can be rewritten to a much shorter form:

```lua
-- https://code.visualstudio.com/docs/editor/userdefinedsnippets#_language-snippet-scope
-- scope can have multiple components.
if parts.scope and not ignore_scope then
	for _, scope in
	ipairs(vim.split(parts.scope, ",", { plain = true }))
	do
		snippets_by_ft[scope] = snippets_by_ft[scope] or {}
		table.insert(snippets_by_ft[scope], snip)
	end
else
	table.insert(snippets_by_ft[default_filetype], snip)
end
```

hinell:
Why not every `s(....)`  takes a separate option `filetype`? I think it should be an option.

hinell:
Well, I still think we should be able to have `.code-snippets`  listed in the `package.json` to be picked up by the `.lazy_load(...)` / `load()`.

I wouldn't want to config snips paths in different places (some lua file, etc). It's logically reasonable to have paths in `package.json`.

lua/luasnip/loaders/from_vscode.lua
L3MON4D3: Ah, good point, yeah that's a bit more straightforward :+1:

L3MON4D3:
> Why not every `s(....)` takes a separate option `filetype`? I think it should be an option.

`.code-snippets` is the only source which provides snippets for multiple filetypes from one file :/
I don't think it's worth it to move that complexity into `add_snippets`

L3MON4D3:
> Well, I still think we should be able to have `.code-snippets` listed in the `package.json` to be picked up by the `.lazy_load(...)` / `load()`.
> 
> I wouldn't want to config snips paths in different places (some lua file, etc). It's logically reasonable to have paths in `package.json`.

Oh, that's (or should be, at least) implemented via 6a46f76

L3MON4D3:
> > Why not every `s(....)` takes a separate option `filetype`? I think it should be an option.
> 
> `.code-snippets` is the only source which provides snippets for multiple filetypes from one file :/ I don't think it's worth it to move that complexity into `add_snippets`

Although, on second thought, it wouldn't lead to much more complexity, and would be pretty cool for `multisnippet`..... Hmmmmmmm
The issue is that reloading would become waaaaay more complicated, since now any file may contribute snippets for any filetype

hinell:
>The issue is that reloading lazy-loading would become waaaaay a bit more complicated

It ain't. It's logical for a snippet to have an associated file type. This way the function that applies snippets may sort them out in advance and update them on demand.

L3MON4D3:
It's also logical to group ones own snippets by filetype :P
I'm not sure what you mean by "sort them out and update them on demand".
The issue im talking about is: we want to lazy-load snippets, which means that snippets are only loaded when they are actually required, for example when a buffer with their filetype is edited. With `package.json`-snippets, this works pretty good because that file describes which filetype any file contributes snippets to, and we can thus delay reading that file until it's filetype is actually needed.
If now any snippet in that file declares that it belongs to a different filetype, that snippet would not be loaded in time, and would be missing.
I'm sure there are ways around that, but imo the vscode-loader covers most common scenarios, and for anything more complicated, lua should be able to handle it.

hinell:
> _I'm sure there are ways around that,_

@L3MON4D3  This is why I'm advocating the `package.json`-centric approach. With `package.json` specified paths, it's easy to do  async/lazy/postponed/ondemand snippets loading. This optimization though is not a performance issue IMO. Dunno why you are caring so much about it.

E.g.:

`snippets/package.json`
```json
{
	"name": "snippets",
	"contributes": {
		"snippets": [
		...
		, { "language": [ "markdown" ], "path": "./markdown.json" }
		...
		]
	}
}
````

**Comment by L3MON4D3** [comment]
I don't understand what that example is supposed to show me, could you explain in more detail?

**Comment by L3MON4D3** [comment]
But, maybe do so in another issue if it's actually about another feature, this is getting off-topic I think

**Comment by hinell** [comment]
>I don't understand what that example is supposed to show me

I made a small mistake. It's supposed to show you how to load `.code-snippets` via `package.json`, not `load_standalone` function.

**Comment by L3MON4D3** [comment]
Ah, okay
Loading just files with the extension `.code-snippets` is possible, what (fundamentally) won't work is that the file advertises filetypes A, B, C in `package.json`, and has snippets that have scope set to filetype D (because then `lazy_load` won't load that file for filetype D, and the snippet will be missing).
If you're advocating for `scope`-support in `package.json`-loaded `.code-snippets`, what is the advantage you see?
One could just as well split the `.code-snippets`-file into multiple `.json(c)`, without `scope` (if that seems like too much work, have you considered using the snipmate-format? It has much less boilerplate, and vscode-snippets can be converted to it with `snippet-converter.nvim`)

**Comment by hinell** [comment]
>_won't load that file for filetype D, and the snippet will be missing_

Not a problem. The `package.json` can easily give a hint to the snippet engine on what kind of file-type scoped snippets it should import from, e.g:
```
{ "language": [ "A", "B", "C", "D" ], "path": "./tool-specific.code-snippets" }
````

>_, what is the advantage you see?_

I told you [elsewhere](https://github.com/L3MON4D3/LuaSnip/pull/846#issuecomment-1487196090) alredy that I have a collection of snippets that I've separated by specific tools and purpose. It would be outright stupid to keep all these in one single `*.json`.  Because of that I even plan to convert all my json files to `yml` for easier editing.

>_snipmate_

I prefer to keep my snips in a cross-platform format actually. Except for LuaSnip, which is kinda different. Would be best if it would b 90% convertible to vscode/textmate format in the future.

**Comment by hinell** [comment]
>_won't load that file for filetype D, and the snippet will be missing_

Not a problem. The `package.json` can easily give a hint to the snippet engine on what kind of file-type scoped snippets it should import from, e.g:
```
{ "language": [ "A", "B", "C", "D" ], "path": "./tool-specific.code-snippets" }
````

>_, what is the advantage you see?_

I told you [elsewhere](https://github.com/L3MON4D3/LuaSnip/pull/846#issuecomment-1487196090) alredy that I have a collection of snippets that I've separated by specific tools and purpose. It would be outright stupid to keep all these in one single per-language `*.json`.  Because of that I even plan to convert all my json files to `yml` for easier editing.

>_snipmate_

I prefer to keep my snips in a cross-platform format actually. Except for LuaSnip, which is kinda different. Would be best if it would b 90% convertible to vscode/textmate format in the future.

**Comment by L3MON4D3** [comment]
Ah okay, okay, now I'm convinced😅
Hadn't kept the reason for all that in mind since the last time we discussed this.
I guess the way to explain this feature would just be that `scope` further filters the filetypes given in `package.json`. How does vscode behave if the scope-filetype is not one of those?

**Comment by hinell** [comment]
>  How does vscode behave if the scope-filetype is not one of those?

[Replied elsewhere already.](https://github.com/L3MON4D3/LuaSnip/pull/846#issuecomment-1490908044) (scope is ignored afaik)

**Comment by L3MON4D3** [comment]
Oh, I kinda though I must've misunderstood something there, because you were talking about compatibility/a standard format.
IIUC a collection which relies on this scope-mechanism to filter snippets can't be used with vscode then, right?

**Comment by hinell** [comment]
BTW, if you want 100% compatibility the VS Code it says that the `package.json#contributions.language` field  should be a of `string` type instead of an array. 

It seems like there is embeded schema for that say it's error (I made an extension to test it):
image

And also, there is no `all` language identifier for that field. You should use empty string instead: 

```json
{
  "contributes": {
    "snippets": [
      {
        "language": "",
        "path": "./snippets/foo.code-snippets"
      }
    ]
  }
}


```

L3MON4D3:
Oh, `language` as an array is used pretty extensively by friendly-snippets, that alone makes it worth supporting (have you tested if vscode accepts it?)

hinell:
BTW, if you want 100% compatibility the VS Code it says that the `package.json#contributions.language` field should be a of `string` type, NOT an array.  It turns out there is embeded schema for that. VS Code says it's error to use array (I made an extension to test it):
image

You should use empty string instead as there is no `all` [language identifier](https://code.visualstudio.com/docs/languages/identifiers) for that field to get snippets filtered based on `scope`s:
```json
{
  "contributes": {
    "snippets": [
      {
        "language": "",
        "path": "./snippets/foo.code-snippets"
      }
    ]
  }
}
```

This way snippets are imported and filtered based on `scope` field. 

>* (have you tested if vscode accepts it?)*

It's likely won't accept it. See the first screenshot (spot the error).

hinell:
I think making LuaSnip 100% compatible with VS Code extension isn't important. Though, the recent findings make it hard to hint engine on what kind of snippets `.code-snippets` contains. Anyway, I don't think it's worth even discuss it as all these files are very small and there is no evidence there is a performance impact on that. Unless of course you can prove me wrong.

L3MON4D3:
Wait, so if `language = ""`, vscode does look at `scope`? Ewww :(

L3MON4D3:
> Anyway, I don't think it's worth even discuss this optimization (i.e. language-aware lazy loading) as all these files are very small and there is no evidence there is a performance impact on that. Unless of course you can prove me wrong

Oh you can easily test that, load something like `friendly-snippets` via `load` and then compare that with `lazy_load`. There's enough of a difference that I wouldn't want the former to run during startup

hinell:
We can use the following hinting approach (using the same file twice):

```json
{
  "contributes": {
    "snippets": [
      {
        "language": "go",
        "path": "./snippets/collection.code-snippets"
      },
      {
        "language": "rust",
        "path": "./snippets/collection.code-snippets"
      },

    ]
  }
}
```

This way engine may figure to use cached file to avoid loading it twice and filter snippets based on `scope`.

>*Oh you can easily test that...*

It's slow indeed. Is there a compilation step for snippets that LuaSnip performs?  Just wondering. :thinking:

L3MON4D3:
> load()ing friendly-snippets is slow indeed. Is there a compilation step for snippets that LuaSnip performs? Just wondering. :thinking:

Hah, it used to be way worse. We used to parse the snippets outright, this is just reading+parsing json and adding them a kind of proxy, which parses them on expand

> This way engine may figure to use cached file to avoid loading it twice and filter snippets based on scope.

Filtering with `scope` should be fine, doesn't seem to introduce much more complexity.
I'll look into that after getting this here done.

L3MON4D3:
Decoupling the filetype(s) which trigger a `lazy_load` from the actual filetype of the snippet is a really nice idea.

One cool thing that is possible with that is setting the `language` in `package.json` to  some framework, and then `scope` each snippets to css, html, js, whatever.
One could then modify [`load_ft_func`](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#config-options) to load the filetype `framework` as soon as a file using that framework is loaded, which will cause our loaders to actually load all snippets associated with that framework.
This `load_ft_func`-modification was already possible before, but requires that snippets are split up into filetypes like `framework1-css`, `framework1-html`, ..., which may bring some cognitive overhead as opposed to having everything in one file (but might be a bit faster, since everything can be loaded even more lazily)
I was a reluctant to implement this because I thought it would require lots of recursive 
copying of snippets, but that is actually not the case

I'm going to implement this here, since it simplifies the implementation of `load_standalone`, seems weird to merge this and then change it immediately.

hinell:
@L3MON4D3 What's the bottleneck of loading snippets?

L3MON4D3:
I suspect that right now, it might be copying around of snippets, but I'm working on that right now
If you want to figure it out, try generating a flamegraph of `load` or `lazy_load`

L3MON4D3:
Continued in #906

PR #896, opened 2023-05-25
fix: merge opts with context

MunifTanjim:
```lua
local increment = postfix("++", {
  f(function(_, parent)
    return string.format("%s = %s + 1", parent.snippet.env.POSTFIX_MATCH, parent.snippet.env.POSTFIX_MATCH)
  end, {}),
}, {
  hidden = true,
  condition = has_word_before,
})
```

^ `hidden = true` in this example was not working as expected.

### DIFF
--- a/lua/luasnip/nodes/snippet.lua
+++ b/lua/luasnip/nodes/snippet.lua
@@ -159,6 +159,8 @@ end
 
 -- context, opts non-nil tables.
 local function init_snippet_context(context, opts)
+	context = opts and vim.tbl_extend("error", context, opts) or context
+
 	local effective_context = {}
 
 	-- trig is set by user, trigger is used internally.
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Oh, those are separate by design: `opts` has options that relate to the snippet itself, while `context` has those that don't (options for expansion (trigger etc), or for how the snippet shows up in `cmp` (hidden)) (at least that's in a perfect world, here condition and show_condition are allowed in the opts-table, where they don't belong).
The example should work with `{trig = "++", hidden = true}`.

MunifTanjim:
Ah, okay. I thought it was supposed to be just another way of defining snippets... as in, the trigger as frist param, snippet as second and additional info in third param.

L3MON4D3:
Ahh okay. But nice initiative :)

If you need hidden often, you can do something like

```lua
ls.setup_snip_env()

local postfix = require("luasnip.extras.postfix").postfix

local hidden_postfix = ls.extend_decorator.apply(postfix, {hidden = true})

ls.add_snippets("all", {hidden_postfix("++", {t"asdf"})}, {key = "asdf"})
```
(doc for extend_decorator [here](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#extend-decorator))

PR #898, opened 2023-05-26
feat: add default condition for postfix

MunifTanjim:
Discussion: https://github.com/L3MON4D3/LuaSnip/discussions/897#discussioncomment-6005691

### DIFF
--- a/lua/luasnip/extras/postfix.lua
+++ b/lua/luasnip/extras/postfix.lua
@@ -55,6 +55,18 @@ local function generate_opts(match_pattern, user_callback)
 	}
 end
 
+local function wrap_condition(user_condition, match_pattern)
+	if not user_condition then
+		user_condition = require("luasnip.util.util").yes
+	end
+
+	return function(line_to_cursor, matched_trigger, captures)
+		return line_to_cursor:sub(1, -1 - #matched_trigger):match(match_pattern)
+				~= nil
+			and user_condition(line_to_cursor, matched_trigger, captures)
+	end
+end
+
 local function postfix(context, nodes, opts)
 	opts = opts or {}
 	local user_callback = vim.tbl_get(opts, "callbacks", -1, events.pre_expand)
@@ -65,17 +77,11 @@ local function postfix(context, nodes, opts)
 		user_callback = { user_callback, { "nil", "function" } },
 	})
 
-	local match_pattern
-	if type(context) == "string" then
-		context = {
-			trig = context,
-		}
-		match_pattern = matches.default
-	else
-		match_pattern = context.match_pattern or matches.default
-	end
+	context = node_util.wrap_context(context)
+	context.wordTrig = false
+	local match_pattern = context.match_pattern or matches.default
+	context.condition = wrap_condition(context.condition, match_pattern)
 
-	context = vim.tbl_deep_extend("keep", context, { wordTrig = false })
 	opts = vim.tbl_deep_extend(
 		"force",
 		opts,
--- a/DOC.md
+++ b/DOC.md
@@ -1652,8 +1652,7 @@ The optional third argument is the same as the third (`opts`) argument to the
 
 The postfix snippet works using a callback on the pre_expand event of the
 snippet. If you pass a callback on the pre_expand event (structure example
-below) it will get run after the builtin callback. This means that your
-callback will have access to the `POSTFIX_MATCH` field as well.
+below) it will get run after the builtin callback.
 
 ```lua
 {
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*             For NVIM v0.8.0             Last change: 2023 May 25
+*luasnip.txt*             For NVIM v0.8.0             Last change: 2023 May 26
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
@@ -1562,8 +1562,7 @@ The optional third argument is the same as the third (`opts`) argument to the
 
 The postfix snippet works using a callback on the pre_expand event of the
 snippet. If you pass a callback on the pre_expand event (structure example
-below) it will get run after the builtin callback. This means that your
-callback will have access to the `POSTFIX_MATCH` field as well.
+below) it will get run after the builtin callback.
 
 >lua
     {


L3MON4D3:
Nice, thank you for doing the PR :+1:

lua/luasnip/extras/postfix.lua
```diff
@@ -65,17 +72,16 @@ local function postfix(context, nodes, opts)
...
-		match_pattern = context.match_pattern or matches.default
-	end
+	local match_pattern = context.match_pattern or matches.default
+
+	context = node_util.wrap_context(context)
+	context = vim.tbl_extend(
+		"keep",
+		context,
+		get_default_context(context.trig, match_pattern)
+	)
```
L3MON4D3, line 82: How about we combine the condition-functions if some are passed by the user?

lua/luasnip/extras/postfix.lua
```diff
@@ -65,17 +72,16 @@ local function postfix(context, nodes, opts)
...
-	local match_pattern
-	if type(context) == "string" then
-		context = {
-			trig = context,
-		}
-		match_pattern = matches.default
-	else
-		match_pattern = context.match_pattern or matches.default
-	end
+	local match_pattern = context.match_pattern or matches.default
```
L3MON4D3, line 75: WDYT about doing this check after `wrap_context` (btw, good that you replaced that :+1:)?
Feels a bit weird to index a potential string, I know it doesn't error, but doesn't sit quite right :sweat_smile:

lua/luasnip/extras/postfix.lua
MunifTanjim: How does it look now? The default condition will only run if user doesn't provide a function or if the user provided function returns `nil`.

L3MON4D3: Mhmm almost, I was thinking about something like
```lua
if user_condition then
	context.condition = function(...)
		return user_condition(...) and enable_postfix_condition(...)
	end
else
	context.condition = enable_postfix_condition
end
```
so both are evaluated (snippet is disabled entirely if there is no text to postfix, and is enabled if there is text to postfix, and the user-condition is true)

MunifTanjim: Updated.

lua/luasnip/extras/postfix.lua
```diff
@@ -55,6 +55,13 @@ local function generate_opts(match_pattern, user_callback)
 	}
 end
 
+local function wrap_condition(user_condition, trigger, match_pattern)
+	return function(line_to_cursor, ...)
+		return (not user_condition or user_condition(line_to_cursor, ...))
+			and line_to_cursor:gsub(trigger, ""):match(match_pattern) ~= nil
```
MunifTanjim, line 61: This `:gsub` might now work as expected. It'll treat `trigger` as lua pattern.

I think replacing it with `:sub` makes more sense.

lua/luasnip/extras/postfix.lua
```diff
@@ -55,6 +55,14 @@ local function generate_opts(match_pattern, user_callback)
 	}
 end
 
+local function wrap_condition(user_condition, trigger, match_pattern)
+	return function(line_to_cursor, ...)
+		return (not user_condition or user_condition(line_to_cursor, ...))
+			and line_to_cursor:sub(1, -1 - #trigger):match(match_pattern)
```
L3MON4D3, line 61: Ah, one thing which is a bit annoying:
`context.trig` might also be a pattern, so just removing its length could be an issue.
Also, if `show_condition` is invoked, the trigger is not yet in the file, so it would not have to be removed from `line_to_cursor`.

Creating separate functions for `condition` and `show_condition` should fix both issues, since `condition` provides the matched trigger, and that can then be removed correctly (and show_condition can just check against the entire `line_to_cursor`)

MunifTanjim: I tried to separate them:

```lua
local function wrap_show_condition(user_condition, match_pattern)
	return function(line_to_cursor)
		local value = (not user_condition or user_condition(line_to_cursor))
			and line_to_cursor:match(match_pattern) ~= nil
		print(vim.inspect({line_to_cursor, value}))
		return value
	end
end
```

```lua
	context.show_condition = wrap_show_condition(context.show_condition, match_pattern)
```

But for some reason it's never showing up in the completion menu after this change. Even though I'm seeing the `value` is `true` in the print statement. 🤔

MunifTanjim: I think it's because completion engine will filter out entries that doesn't match with the input. If you typed `count` then trigger completion, `++` won't match with `count`, so it's hidden from completion menu even tho `show_condition` returned `true`.

L3MON4D3: Ohh, that's a bummer :/
Makes sense though.. there should be some ways to customize the fuzzy matching of cmp, maybe we can hijack that somehow.

MunifTanjim: Is it okay to not do anything with `show_condition` right now and just fix `condition`? 🤔

L3MON4D3: Yeah, feel free to :+1:
I think that needs some fix in `cmp_luasnip` anyway

MunifTanjim: Done.

lua/luasnip/extras/postfix.lua
```diff
@@ -55,6 +55,19 @@ local function generate_opts(match_pattern, user_callback)
 	}
 end
 
+local function wrap_condition(user_condition, match_pattern)
+	return function(line_to_cursor, matched_trigger, captures)
+		return (
+			not user_condition
+			or user_condition(line_to_cursor, matched_trigger, captures)
+		)
```
L3MON4D3, line 63: Yup :+1:
I think it would be better to do the check for `user_condition` once while constructing the function, and not every time it is called, wdyt?

L3MON4D3:
Could you also work in an option to disable our condition? Should be fine to pass it via `context`, no need for yet another `opts`-table xD

lua/luasnip/extras/postfix.lua
MunifTanjim: > Could you also work in an option to disable our condition? Should be fine to pass it via context, no need for yet another opts-table xD

I didn't want to add another option 😂

Can't we do one of the following?
1. If user provides `condition` just use that instead of the internal one
2. Let use return `nil` from the provided `condition` to fallback to the internal one

1 is very simple and clean. 2 gives the user to handle specialized cases and fallback to the internal logic when they want (by returning `nil` instead of `boolean`).

---

If you definitely want to add another option, just give me a name for it (naming is hard 😂) and I can add it in.

MunifTanjim: > I think it would be better to do the check for user_condition once while constructing the function, and not every time it is called, wdyt?

If we check `user_condition` outside the actual callback, we either need to:
- assign a fallback function that always returns `true` to `user_condition`
- or return another callback with only `return line_to_cursor:sub(1, -1 - #matched_trigger):match(match_pattern) ~= nil`.

Which one do you prefer?

L3MON4D3: > Which one do you prefer?

I guess the first one will look a bit better in the code.. There's a `yes`-function in util, `require("luasnip.util.util").yes`


> I didn't want to add another option :joy:

I do get that xD
But I think it's the best way to optionally turn off our condition I think

> Can't we do one of the following?
> 
> 1. If user provides `condition` just use that instead of the internal one
> 2. Let use return `nil` from the provided `condition` to fallback to the internal one
> 
> 1 is very simple and clean. 2 gives the user to handle specialized cases and fallback to the internal logic when they want (by returning `nil` instead of `boolean`).
> 
> If you definitely want to add another option, just give me a name for it (naming is hard joy) and I can add it in.

Ahh what I don't really like there is that "I want to expand postfix regardless of what is in front of it" is unrelated to wanting a `condition`, so it's a bit weird to require a condition to get that desired behaviour.

MunifTanjim: > I want to expand postfix regardless of what is in front of it

That can be controlled by tweaking `match_pattern` right? Using `match_pattern = [[[%w%.%_%-%"%']*$]]` would enable user to expand postfix even if there is nothing in front of it.

L3MON4D3: Ouuuuhh yeah that's cool :+1:

L3MON4D3:
Aaaalrigth, this is done :ok_hand:
Thank you!!

Issue #911, opened 2023-06-06
cannot clone luasnip with submodules from mirror site without GitHub access


gongfarmer:
I am running neovim at a large company within a secure development environment.
There is no access to https://github.com/, however there is an internal mirror site with a different URL that provides access to GitHub projects.

The lazy.nvim package manager gets plugins from our local mirror with this neovim configuration:

```lua
local opts = {
  git = {
    url_format = 'https://repomirror.corporate.com/github/%s.git',
  }
}
require('lazy').setup('plugins', opts)
```

This does not work for the luasnip plugin.  This plugin defines a submodule with a GitHub URL that is not reachable:

```
$ cat .gitmodules
[submodule "deps/jsregexp"]
	path = deps/jsregexp
	url = https://github.com/kmarius/jsregexp/
```

This has an [easy solution](https://confluence.atlassian.com/bitbucketserverkb/cloning-submodule-fails-from-mirror-946044514.html): Use a relative path for the url in .gitmodules.

Here is what the problem looks like (using plain git for clarity, but lazy.nvim uses this exact command internally for its clones:)

```
$ git clone --filter=blob:none --recurse-submodules --origin=origin  --progress  https://repomirror.corporate.com/github/L3MON4D3/LuaSnip.git
Cloning into 'LuaSnip'...
Fetching objects: 11498, done.
Submodule 'deps/jsregexp' (https://github.com/kmarius/jsregexp/) registered for path 'deps/jsregexp'
Cloning into '/home/bob/LuaSnip/deps/jsregexp'...
fatal: unable to access 'https://github.com/kmarius/jsregexp/': gnutls_handshake() failed: Error in the pull function.
fatal: clone of 'https://github.com/kmarius/jsregexp/' into submodule path '/home/bob/LuaSnip/deps/jsregexp' failed
Failed to clone 'deps/jsregexp'. Retry scheduled
Cloning into '/usr/local/ldr/LuaSnip/deps/jsregexp'...
fatal: unable to access 'https://github.com/kmarius/jsregexp/': gnutls_handshake() failed: Error in the pull function.
fatal: clone of 'https://github.com/kmarius/jsregexp/' into submodule path '/home/bob/LuaSnip/deps/jsregexp' failed
Failed to clone 'deps/jsregexp' a second time, aborting
```

L3MON4D3:
(Closed with #912)

PR #912, opened 2023-06-06
fix: use relative url for submodule definition (#911)

gongfarmer:
Tested within secure environment without GitHub access.
Tested in normal situation with GitHub access.


L3MON4D3:
Ah, nice👍
I'm a tiny bit hesitant because I have no idea if luasnip and jsregexp will always be developed on the same service, but I guess until then this is pretty practical.

Thank you!

PR #923, opened 2023-06-14
Implement more trigger-types

L3MON4D3:
This implements a new option, `trigMatcher`, which determines how the trigger is matched to the current line.

I've veered from calling it `trigType`, since it sounds like it would only accepts a string, whereas I'd like to make it possible to completely customize the trigger-matching by passing a function.
So, currently, `trigMatcher` accepts either a function (then uses that), a string (any from "plain", "pattern", "vim", "ecma" for plain strings, lua-patterns, vim-regex and ecma-regex respectively (the last two are not yet implemented)). If `trigMatcher` is not set at all, we look if `regTrig` is set and use "pattern" if it is, and if that is not set as well, we fall back to "plain".
I think this will also deprecate `regTrig` since that name was misleading from the start (but a very soft deprecation, I don't think we will actually remove it)

I'm not completely convinced by the name `trigMatcher`, if somebody has a different idea, mention it :D

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -170,7 +170,11 @@ s({trig="trigger"}, {})
   ```
 
   The following keys are valid:
-  - `trig`: string, plain text by default. The only entry that must be given.
+  - `trig`: string, the trigger of the snippet. If the text in front of (to the
+    left of) the cursor when `ls.expand()` is called matches it, the snippet
+    will be expanded.  
+    By default, "matches" means the text in front of the cursor matches the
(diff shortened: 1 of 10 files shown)


aikow:
maybe just use 'trigger' instead of `trigMatcher`. Or you could follow regex terminology and use something like `trigEngine`...

L3MON4D3:
Oh I like trigEngine, ty👍👍

L3MON4D3:
Alright, this should be it.
Any comments, now's the time :D

PR #930, opened 2023-06-19
fix: remove debug print

stasjok:
I guess it shouldn't be there. It prints something like

```
true table: 0x7fffea36bbc0
true table: 0x7fffea36bab0
true table: 0x7fffea36bbc0
true table: 0x7fffea36bab0
```


L3MON4D3:
Oh no, I missed that😬😬
TY!

PR #945, opened 2023-07-04
Subtraction for condition objects

bhataktaBhai:
This adds `c1 - c2` to be defined as `c1 and not c2` for condition objects. This seems clear and intuitive to me, and prettier than `c1 * (-c2)`.

My use case looked like `condition = tex.in_mathzone - tex.in_command`.

EDIT: On deeper thoughts, perhaps this could be confused with `c1 + (-c2)`.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1505,9 +1505,14 @@ Additionally, `expand` contains all conditions provided by `show`.
 like functions, be passed to `condition` or `show_condition`, but can also be
 combined with each other into logical expressions:
 
-- `c1 + c2 -> c1 or c2`
-- `c1 * c2 -> c1 and c2`
 - `-c1 -> not c1`
+- `c1 * c2 -> c1 and c2`
+- `c1 + c2 -> c1 or c2`
+- `c1 - c2 -> c1 and not c2`: This is similar to set differences:
+  `A \ B = {a in A | a not in B}`. This makes `-(a + b) = -a - b` an identity
+  representing de Morgan's law: `not (a or b) = not a and not b`. However,
+  since boolean algebra lacks an additive inverse, `a + (-b) = a - b` does not
+  hold. Thus, this is NOT the same as `c1 + (-c2)`.
 - `c1 ^ c2 -> c1 xor(!=) c2`
 - `c1 % c2 -> c1 xnor(==) c2`: This decision may seem weird, considering how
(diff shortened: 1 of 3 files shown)


L3MON4D3:
Hey, thank you for the PR :)

I'd agree that this could lead to some confusion, both with +- (as you said), but it might also be confusing to have the unary minus and minus, for pretty different things.
OTOH, weirdly, I do agree that `cond1 - cond2` intuitively means `cond1 given cond2 does not hold`, but I'm not sure where this intuition is coming from, and so am hesitant to add `-` just based on that :sweat_smile:

Can you think of a reason for this intuition? Maybe a good mnemonic device to put into `DOC.md`, so it's easy to understand for those who don't think it's intuitive?

bhataktaBhai:
The difference of two sets is defined very similarly. `A - B` (or `A \ B`) `= {a in A | a not in B}`.

Moreover, if you want `-(a + b) = - a - b`, where the unary minus and binary plus are defined as 'not' and 'or', binary minus would have to be defined as above for de Morgan's law to hold (`not (a or b) = not a and not b`). Also, `a - a` should obviously be false (0), not _always true_ as `a + -a` would be.

I don't know if any of these make a good mnemonic.

L3MON4D3:
Ahhh it's probably the set-minus, okay.
Would you address the potential for confusion in the documentation, as well as that mnemonic (and mention that there is some internal inconsistency in the boolean algebra if - is set to either definition (either DeMorgan holds, or A + (-B) = A-B))
Would you also squash the commits into just one?

bhataktaBhai:
Does this look fine?

Sorry about all the force-pushing messages. I am learning my way around git.

atticus-sullivan:
Force-pushing is necessary if you need to squash the commits. (And it's absolutely fine if you don't expect someone else working on the same branch, which holds for a PR here)

I like the reference to sets :+1:
Implementation also looks good to me :+1:

L3MON4D3:
Thanks for taking a look @atticus-sullivan!
I think the test-failure is due to some weird race-condition I've yet to see during normal usage, but I'll merge once i can verify it myself (am away rn), or CI reports success (or until anyone else can confirm that no test fails, wink wink nudge nudge)

L3MON4D3:
Ah there it is :D
Thank you @bhataktaBhai, a nice addition :)

PR #960, opened 2023-07-18
fix: install jsregexp

henryhchchc:
`make install_jsregexp` will fail when there is a space in the path to luasnip


L3MON4D3:
Ofc!
Very clean, thank you!

PR #966, opened 2023-07-26
fix: catch undocumented edge case in scandir()

troiganto:
While [libuv/vim.loop docs](https://github.com/luvit/luv/blob/master/docs.md#uvfs_scandir_nextfs) say that `fs_scandir_next()` either returns `nil` or `string, string` or fails, there is an edge case on networked filesystems. In those cases, `uv__fs_scandir()` returns the type `UV_DIRENT_UNKNOWN` and `luv.fs_scandir_next()` [converts][1] this into returning a single string.

This means in those cases, `name` is a string and `type` is `nil`. See the [upstream bug report][2].

The situation can be remedied by explicitly calling `fs_stat()` on those files. This always fetches the correct type, as far as I can see.

PS: Thanks for this awesome plugin! :slightly_smiling_face: 

[1]: https://github.com/luvit/luv/blob/master/src/fs.c#L116
[2]: https://github.com/luvit/luv/issues/660

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*             For NVIM v0.8.0            Last change: 2023 July 18
+*luasnip.txt*             For NVIM v0.8.0            Last change: 2023 July 26
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Damn, that's a clean PR👌👌👌👌
Thank you very much!!

PR #974, opened 2023-08-03
support multiline-trigger

TwIStOy:


L3MON4D3:
Hi!
Could you explain how this helps you? afaict we still only consider the current line when matching the trigger, so I can't tell how this makes multi-line triggers work

TwIStOy:
I'm trying to use `treesitter` to trigger my snippets. Here's my custom `trig_engine`, https://github.com/TwIStOy/dotvim/blob/master/lua/ht/snippets/trig_engines.lua

Some of matched TSNodes are multi-line.

Here's the example:

I have a postfix-like snippet `.mv` to expand `{  "call_expression", "identifier", "template_function", "subscript_expression", "field_expression", "user_defined_literal",
}` expressions to `std::move($expr)`.

In the following example, the `$expr` is multi-line.

```
Test()
  .test()
  .test()
  .test().mv
```

L3MON4D3:
First of all, very cool!!! I'm very happy to see some snippets making use of the more advanced capabilities of luasnip :D

What I dislike a bit about this approach is "misusing" `trigEngine` to effectively implement a condition (I'm assuming you chose to do that because there is no obvious way to modify the trigger from `condition`)
Instead of working around this limitation, I think it would be better to directly tackle it, and provide some way to pass data from `condition` into the snippet/to `snip_expand` (in this case, the range of the treesitter-match).

Should be possible, `condition` is evaluated in `Snippet:matches`, and its return-value is `expand_params` in `luasnip.expand`. If we specify that `expand_params` has to provide the `clear_region` (instead of computing it in expand using the trigger), and the return of `condition` may overwrite the default which is computed from the trigger, that would be handled.
Getting the region into the snippet will probably be a bit messier, we could again specify that some return value from `condition` is attached to the expanded snippet-object, so it can be used by functionNode/dynamicNode, but I'm not as sure about this.

TwIStOy:
make sense! but the only problem, is field `condition` the best name to do this? should we introduce a new option to overload `expand_params` and `captures`?

TwIStOy:
Update it! Add a new option `context_matcher` returns `expand_params` with more fields.

L3MON4D3:
Oh awesome, taking a look

lua/luasnip/nodes/snippet.lua
```diff
@@ -20,6 +20,23 @@ local trig_engines = require("luasnip.nodes.util.trig_engines")
...
+		if not self.condition(line_to_cursor, match, captures) then
+			return nil
+		end
+	end
+
+	return {
+		trigger = match,
+		captures = captures,
+		trig_region = nil,
+		env_override = nil,
```
L3MON4D3, line 36: WDYT about `trig_region` -> `clear_region`? To me, `trig_region` implies some interdependence with trigger, where `clear_region` is a bit more precise in what this is used for.

lua/luasnip/nodes/snippet.lua
```diff
@@ -228,9 +245,10 @@ local function init_snippet_context(context, opts)
 	end
 	effective_context.trig_matcher = engine(effective_context.trigger)
 
-	effective_context.condition = context.condition
-		or opts.condition
-		or true_func
+	effective_context.context_matcher = context.context_matcher
+		or opts.context_matcher
```
L3MON4D3, line 249: Ah, the `or opts.condition` above was for backwards compatibility, `context_matcher` should only be set from `context`.
(options from context relate to stuff happening before the expansion of the snippet, where as fields from opts influence behaviour after expansion)

lua/luasnip/nodes/snippet.lua
```diff
@@ -228,9 +245,10 @@ local function init_snippet_context(context, opts)
 	end
 	effective_context.trig_matcher = engine(effective_context.trigger)
 
-	effective_context.condition = context.condition
-		or opts.condition
-		or true_func
+	effective_context.context_matcher = context.context_matcher
```
L3MON4D3, line 248: What is your intuition for the name `context_matcher`? It doesn't really mean anything to me :sweat_smile:

lua/luasnip/nodes/snippet.lua
TwIStOy: i think `trig_region` is a property of this snippet expanding. it indicates the actual region where the snippets was triggered. The region is passed to `clear_region` because it needs to be cleared first in `Snippet:expand()`.

lua/luasnip/nodes/snippet.lua
TwIStOy: fix it.

lua/luasnip/nodes/snippet.lua
TwIStOy: I think the name is not good enough either. The function returns some properties that are used in snippet expanding. Actually, the `context` refers the context in snippet expanding. And the `matcher` refers if the snippet matches or not.
Do you have any suggestion?

lua/luasnip/nodes/snippet.lua
L3MON4D3: I'm not sure I understand your response :/
I was suggesting to change the name from `trig_region` to `clear_region` (or something else :D) because `trig_region` sounds like it might have to be the range `trigger` occupies in the buffer, which is not the case.

lua/luasnip/nodes/snippet.lua
L3MON4D3: I think we should go as generic as possible, because it's not impossible this will be used for even more stuff later on.
The only thing that would define it then is when it is called, so I'd suggest something like `post_trig_match` (although that is very verbose.. just `post_match` maybe? But that would imply that the snippet is matched, and we couldn't change that anymore, which we can..)

L3MON4D3:
Another thing: it would be nice to handle the case where a user passes a `condition` and a `context_matcher` gracefully by making the condition override whatever is returned by the context_matcher (since condition is more specific).
(Implementation-wise, generating a new context_matcher which calls the condition and the passed context_matcher seems reasonable)

lua/luasnip/init.lua
```diff
@@ -302,19 +303,15 @@ local function expand(opts)
...
 			expand_params = expand_params,
 			-- clear trigger-text.
-			clear_region = {
-				from = {
-					cursor[1],
-					cursor[2] - #expand_params.trigger,
-				},
-				to = cursor,
-			},
+			clear_region = expand_params.trig_region,
```
L3MON4D3, line 311: I think having the length of the trigger as a default here was good, otherwise every `context_matcher` which does not modify the clear_region has to include this boilerplate piece of code

lua/luasnip/nodes/snippet.lua
TwIStOy: I mean that maybe the function's return should be more generic. The name `clear_region` sounds like this field is only be used in snippet expanding to clear triggered region. The `trig_region` means the actual region where the snippets was triggered. Maybe it can be used for other function.

lua/luasnip/init.lua
TwIStOy: make sense! fix it!

lua/luasnip/nodes/snippet.lua
TwIStOy: Maybe `get_expanding_properties`? If this `get` fails, obviously the snippet can't be expanded.

TwIStOy:
> Another thing: it would be nice to handle the case where a user passes a `condition` and a `context_matcher` gracefully by making the condition override whatever is returned by the context_matcher (since condition is more specific). (Implementation-wise, generating a new context_matcher which calls the condition and the passed context_matcher seems reasonable)

Updated!

lua/luasnip/nodes/snippet.lua
L3MON4D3: Ah alright, now I understand :+1:
Hm, for me the trigger is what you typed to invoke the snippet, and intuitively, I wouldn't say that the text that is matched with some condition belongs to it.
I'd even say that using `clear_region` to remove this text is a bit of a hack, but manually removing it in `context_matcher` will shift the expand-position and make `TM_CURRENT_LINE`, an env-variable which contains the line the snippet was expanded on, contain the wrong text (because the text that was there originally was removed), which means that snip_expand has to handle the removal after it saved the position and stored the current line

lua/luasnip/nodes/snippet.lua
L3MON4D3: Works, but still not a clear winner for me :sweat_smile:
I think it's just that `get` sounds too much like there certainly is something.. Maybe `resolve`, `compute`..
`compute_expand_params` (`expand_params` because that's essentially where this stuff ends up?)

L3MON4D3: But all of those are too long :(((

lua/luasnip/nodes/snippet.lua
```diff
@@ -21,6 +21,19 @@ local true_func = function()
...
+	return function(self, line_to_cursor, match, captures)
+		if condition(line_to_cursor, match, captures) then
+			return {
+				trigger = match,
+				captures = captures,
+			}
+		else
+			return nil
+		end
+	end
```
L3MON4D3, line 34: Oh it should also process a user-provided context_matcher if the condition is true

TwIStOy: Ah, sorry. I misunderstand it. Fix it!

lua/luasnip/nodes/snippet.lua
TwIStOy: I update the name to `resolve_expand_params`~

lua/luasnip/nodes/snippet.lua
TwIStOy: emmm. Make sense! update it.

lua/luasnip/extras/postfix.lua
```diff
@@ -2,53 +2,61 @@ local snip = require("luasnip.nodes.snippet").S
...
-						},
-					}
-
-					return vim.tbl_deep_extend(
-						"keep",
-						user_env,
-						postfix_env_override
-					)
+					return user_env
 				end,
```
L3MON4D3, line 60: I think this can be removed entirely, all it does now is call the user-callback, and this also works by just passing `opts` to `snip(..)`

lua/luasnip/nodes/snippet.lua
```diff
@@ -228,9 +254,12 @@ local function init_snippet_context(context, opts)
...
 
-	effective_context.condition = context.condition
-		or opts.condition
-		or true_func
+	effective_context.resolve_expand_params =
+		generate_resolve_expand_params_func(
+			context.condition or opts.condition,
+			context.resolve_expand_params
+		)
+	effective_context.condition = context.condition or opts.condition
```
L3MON4D3, line 262: Technically, I don't think we even have to store `condition` anymore, it's only used in `resolve_expand_params`, and we can put it in there via `generate_resolve_expand_params_func`

L3MON4D3: (Oh btw, "gen_expand_params" would also be a good name, wdyt?)

lua/luasnip/extras/postfix.lua
```diff
@@ -2,53 +2,61 @@ local snip = require("luasnip.nodes.snippet").S
...
+		local res = {
+			trigger = match,
+			captures = captures,
+			clear_region = {
+				from = { pos[1], pos[2] - #postfix_match - #match },
+				to = pos,
+			},
+			env_override = {
+				POSTFIX_MATCH = postfix_match,
+			},
```
L3MON4D3, line 34: I've looked into the test-failures: Looks like `POSTFIX_MATCH` is not passed all the way into the snippet, does that work on your machine?

TwIStOy: i'm trying fix this. in postfix testcases, `snip_expand` is invoked directly, skip ` matches`, so the `POSTFIX_MATCH` would not be set. qaq

lua/luasnip/nodes/snippet.lua
TwIStOy: yeah, i will fix it later

lua/luasnip/extras/postfix.lua
TwIStOy: you are right. i will fix it later.

lua/luasnip/extras/postfix.lua
L3MON4D3: Ahh right, good catch.
I can handle the test-cases if you like :)

TwIStOy: fix it, by adding a new func `snip_try_expand`. take a look :)

lua/luasnip/init.lua
```diff
@@ -281,6 +281,37 @@ local function snip_expand(snippet, opts)
 	return snip
 end
 
+---Try to expand a snippet at the current cursor position.
+local function snip_try_expand(snippet, opts)
```
L3MON4D3, line 285: Cool idea!
Although I'd frame it not like "try to match the snippet here and expand if it works", but as "okay since the snippet may want to retrieve some additional data and put that into `expand_params`, here's this helper which retrieves all that".
Basically, I'm proposing to change the name to "snip_match_expand" xD

lua/luasnip/extras/postfix.lua
```diff
@@ -2,68 +2,56 @@ local snip = require("luasnip.nodes.snippet").S
...
-				end,
-			},
-		},
-	}
-end
+		if user_condition then
+			if not user_condition(line_to_cursor, match, captures) then
+				return nil
+			end
+		end
```
L3MON4D3, line 26: I think we can throw this out too now :D

TwIStOy: remove this `user_condition` support?

L3MON4D3: Yeah, I think that handles `snip.condition`, and that is entirely inside the `resolve_expand_params`-callback now (ie. no longer set at all)

TwIStOy: Ah, i got it. fix this.

lua/luasnip/nodes/snippet.lua
```diff
@@ -20,6 +20,32 @@ local trig_engines = require("luasnip.nodes.util.trig_engines")
...
+			trigger = match,
+			captures = captures,
+		}
+
+		if user_resolve then
+			local res = user_resolve(self, line_to_cursor, match, captures)
+			if res == nil then
+				return nil
+			end
+			return vim.tbl_extend("force", default_expand_params, res)
```
L3MON4D3, line 42: Didn't even notice this until now, very user-friendly behaviour :+1:

L3MON4D3:
Damn, okay code wise this looks pretty much perfect, amazing work :heart: (latest test-failure seems unrelated, and just the test-framework messing up)
Are you up for documenting the changes as well?

TwIStOy:
> Damn, okay code wise this looks pretty much perfect, amazing work ❤️ (latest test-failure seems unrelated, and just the test-framework messing up) Are you up for documenting the changes as well?

try to write a version, but i'm afraid my english is not good enough. feel free to modify it or ask me to fix. QAQ

L3MON4D3:
:sweat_smile:
Alright, I'll make corrections :D

TwIStOy:
Thanks! QAQ

L3MON4D3:
Alright, I wrote some documentation, and made small changes and added two more tetst.
Also took a closer look at `ls.snip_match_expand(snip)`, I think we can actually use `ls.expand()` instead.

L3MON4D3:
If you're fine with the changes, I think we can squash all of these commits, and then only have to wait for https://github.com/iurimateus/luasnip-latex-snippets.nvim/issues/14 to be fixed

L3MON4D3:
Also renamed `resolve_expand_params` to fit better with the camelCase of the other options in `context`

TwIStOy:
All looks good to me! Thank you!

L3MON4D3:
Alright, thank you very much for your work!!
If you ever have time to contribute again, I'd love to have your treesitter-postfix-snippets in `extras` as well :D

L3MON4D3:
Alright github doesn't seem to like what I did there, but this is successfully merged into master.

TwIStOy:
> Alright, thank you very much for your work!! If you ever have time to contribute again, I'd love to have your treesitter-postfix-snippets in `extras` as well :D

of source! i'm trying to make the code more generic. then i will create a new pr <(^_^)>

PR #975, opened 2023-08-03
docs: fix snip-env-src

b0ae989c:


L3MON4D3:
Thank you!
No idea how to automatically keep this from getting out of date, but this works too :P

Issue #979, opened 2023-08-12
Definition order for normal and postfix snippets with and without preceding "." in the trigger leads to inconsistent expansion


kylefhartzenberg:
As an example, consider the following snippets to be used to add the overline command inside LaTeX mathematics environments:
```lua
return{
    postfix( -- over-line (i.e. bar) post-fix
        {
            trig = ".ov",
            dscr = "Automatically post-fix text to be '\\overline{}'",
            snippetType = "autosnippet",
        },
        {
            f(function(_, parent)
                return "\\overline{" .. parent.snippet.env.POSTFIX_MATCH .. "}"
            end, {}),
        }
    ),
    s( -- over-line (i.e. bar)
        {
            trig = "ov",
            dscr="Automatically expand 'ov' to be '\\overline{}'",
            snippetType="autosnippet",
        },
        fmt(
            [[
            \overline{<>}
            ]],
            {  i(1),  },
            { delimiters = "<>" }
        )
    ),
}
```

If one types `x` followed by `.ov`, the postfix snippet expands correctly producing `\overline{x}`. However, if the `postfix` snippet above is defined *after* the normal snippet `s`, then the same key press sequence produces `x.\overline{}`.

I am not sure if this is an issue, or expected behaviour. If the latter, then it would be worth adding it to the [Postfix section of the documentation](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#postfix-snippet) as I couldn't readily find why my postfix snippets started behaving differently until I resorted to restoring my `tex.lua` snippets file. After restoration, I then noticed that it was the re-ordering that caused the inconsistent behaviour. Let me know if you'd like me to add a paragraph to the documentation and send through a pull request.

I am using the latest version of Neovim (0.9.1), LuaSnip and nvim-cmp at the time of writing.

L3MON4D3:
Ah, I'd categorise that as expected behaviour: if there are two snippets that could match the buffer-position, either one may expand (in reality it's deterministic (at least for now), because we check whether a snippet matches in the order they were added).

You can make the preference explicit by providing a priority.

I think we have some documentation on this (for example in the documentation for `add_snippets`, and `snippet`, IIRC), but if you have a good place to put it, I'd welcome the addition :)
(maybe something like `Troubleshooting`?)

kylefhartzenberg:
[...]

PR #981, opened 2023-08-15
Update DOC.md Troubleshooting section

kylefhartzenberg:
Pull request addresses https://github.com/L3MON4D3/LuaSnip/issues/979 by adding an explanation regarding the behaviour of snippet expansion when multiple snippets can potentially expand at the current buffer-position.

The explanation was appended to the Troubleshooting section of the documentation given that snippet priority is already discussed prior, and the intent is to make users aware of *potentially* "unexpected" behaviour when reordering snippets that share similar triggers.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -2294,22 +2294,23 @@ For easy editing of these files, LuaSnip provides a `vim.ui.select`-based dialog
 ([Loaders-edit_snippets](#edit_snippets)) where first the filetype, and then the
 file can be selected.
 
-### Snippet-specific filetypes
+## Snippet-specific filetypes
 Some loaders (vscode,lua) support giving snippets generated in some file their
 own filetype (vscode via `scope`, lua via the underlying `filetype`-option for
 snippets). These snippet-specific filetypes are not considered when determining
 which files to `lazy_load` for some filetype, this is exclusively determined by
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Ah, nice!
The existing troubleshooting-section is specific to the loaders, I don't think general advice like priority should be in there.

Could you instead create a toplevel-section on troubleshooting, move the content from loaders/troubleshooting to troubleshooting/adding snippets/loaders and put the new advice into its own subsection, troubleshooting/adding snippets/general? I'm not set on the structure, if you think something else makes more sense, go ahead :D

kylefhartzenberg:
That makes sense. I thought the same but I didn't want to interfere with the existing structure. The requested changes have been made. Let me know if you'd like anything else updated/edited.

L3MON4D3:
Good instinct :D
I'm happy with the content, could you move the section between `Config-Options` and `Api`?
And then squash all commits into just one?

kylefhartzenberg:
Done and done. Also, added a link at the bottom of the original [Snippet-specific filetypes](https://github.com/kylefhartzenberg/LuaSnip/blob/master/DOC.md#snippet-specific-filetypes) section to reference relevant sub-section that's now in Troubleshooting.

L3MON4D3:
Beautiful, thank you very much!

PR #1016, opened 2023-09-25
Add nil for type opt for parse_snippet

philolo1:
This commit fixes the typing issues that lua lsp would complain about the argument being required.

image


L3MON4D3:
Yup, TY!!

PR #1024, opened 2023-10-01
Fix lazy install tag in README.md

pockethook:


L3MON4D3:
Oh sorry, I glanced at this and then forgot to actually merge :sweat_smile:
Thank you, I hadn't spotted this mistake!

PR #1027, opened 2023-10-04
feat(snip_expand): fix improper call to `nvim_buf_set_text`

hinell:
https://github.com/L3MON4D3/LuaSnip/blob/ad089ed4580a65e0e4f89abb2876d7c132366713/lua/luasnip/init.lua#L326-L335

https://github.com/L3MON4D3/LuaSnip/blob/ad089ed4580a65e0e4f89abb2876d7c132366713/lua/luasnip/init.lua#L354-L361

**This PR ensures that text is never put below line 0.** 

`snip_expand()`  calls `nvim_buf_set_text()` which has zero-based indexing. The above chunks of code potentially introduce negative line rows. This PR should fix this.

This PR intended to save my patch. The issue was orignally discovered when I was using the [benfowler/telescope-luasnip.nvim](https://github.com/benfowler/telescope-luasnip.nvim) plugin.

See `h nvim_buf_set_text`

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*             For NVIM v0.8.0            Last change: 2023 July 18
+*luasnip.txt*           For NVIM v0.8.0           Last change: 2023 October 04
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
@@ -214,7 +214,7 @@ API                                                         *luasnip-node-api*
 - `get_buf_position(opts) -> {from_position, to_position}`:
     Determines the range of the buffer occupied by this node. `from`- and
     `to_position` are `row,column`-tuples, `0,0`-indexed (first line is 0, first
-    column is 0) and end-inclusive (see `:h api-indexing`, this is extmarks
+    column is 0) and end-inclusive (see |api-indexing|, this is extmarks
     indexing).
     - `opts`: `table|nil`, options, valid keys are:
         - `raw`: `bool`, default `true`. This can be used to switch between
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -230,12 +230,18 @@ local function snip_expand(snippet, opts)
 	-- position, which is just as bad) if text before the cursor, on the same
 	-- line is cleared.
 	if opts.clear_region then
+		-- Zer based indexing; this prevents negative values
+		local start_row = math.max(0, opts.clear_region.from[1])
+		local start_col = math.max(0, opts.clear_region.from[2])
+		local end_row   = math.max(0, opts.clear_region.to[1])
+		local end_col   = math.max(0, opts.clear_region.to[2]  )
+
 		vim.api.nvim_buf_set_text(
 			0,
-			opts.clear_region.from[1],
-			opts.clear_region.from[2],
-			opts.clear_region.to[1],
-			opts.clear_region.to[2],
+			start_row,
+			start_col,
+			end_row,
+			end_col,
 			{ "" }
 		)
 	end
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -233,8 +233,8 @@ local function snip_expand(snippet, opts)
 		-- Zer based indexing; this prevents negative values
 		local start_row = math.max(0, opts.clear_region.from[1])
 		local start_col = math.max(0, opts.clear_region.from[2])
-		local end_row   = math.max(0, opts.clear_region.to[1])
-		local end_col   = math.max(0, opts.clear_region.to[2]  )
+		local end_row = math.max(0, opts.clear_region.to[1])
+		local end_col = math.max(0, opts.clear_region.to[2])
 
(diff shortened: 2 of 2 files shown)


L3MON4D3:
Hmm, not 100% convinced that this is the best approach to fixing this:
I'd argue that if a caller has a bug where the region is unexpectedly wrong (ie. has negative positions), it would be better to throw an error, and not silently try to fix it (which might also lead to undesirable results).

How about we just check for negative positions and throw a more readable error?

hinell:
> How about we just check for negative positions and throw a more readable error?

`nvim_buf_set_text` already does that

This PR is a hotfix. `expand_params.clear_region`  is mishandled and it's your responsibility to provide proper values to the `nvim_buf_set_text` function. This PR simply normalizes input values so there are no negatives.

Otherwise error are thrown.

L3MON4D3:
Have you considered that it's the caller responsibility to provide ptoper values to the `snip_expand`-function?

hinell:
The PR is clear. It's a hotfix that protects you from shooting yourself into the foot. I didn't investigate why `expand_params.clear_region `  is mishandled. I can't fix it cause it's you who is author of the lines below:

https://github.com/L3MON4D3/LuaSnip/blob/ad089ed4580a65e0e4f89abb2876d7c132366713/lua/luasnip/init.lua#L326-L335

https://github.com/L3MON4D3/LuaSnip/blob/ad089ed4580a65e0e4f89abb2876d7c132366713/lua/luasnip/init.lua#L354-L361

L3MON4D3:
Okay, hotfix for what issue? I'd rather fix that then do something half-baked.
If the issue is in the telescope-plugin, I didn't find it. Could you send a direct link to the issue?

hinell:
1. Read `nvim_buf_set_text` docs and learn that it takes zero-based indexes.
2. If you look at the PR you find that it simply ensures that input to the `nvim_buf_set_text` is never below 0.

I'm busy with others things and I have no idea what's the purpose of the `expand_params.clear_region` and how should be it fixed in order for PR to be accepted. I just minimaze my efforts to ensure there are no silly errors.

hinell:
1. Read `nvim_buf_set_text` docs and learn that it takes zero-based indexes.
2. If you look at the PR you'll find that it simply ensures that input to the `nvim_buf_set_text` is never below 0.

I'm busy with others things and I have no idea what's the purpose of the `expand_params.clear_region` and how should be it fixed in order for PR to be accepted given your resistance. I just minimaze my efforts to ensure there are no weird errors.

L3MON4D3:
Alright, you can either
* implement my proposal of checking the region's validity before passing it on to `nvim_buf_set_text`, which would be better than the current implementation because the error can be more readable, or
* get on with you other things

I think I've been very clear in why I think silently fixing the mistake of the caller is a bad idea, and I really tire of arguing with you.

hinell:
> I think I've been very clear in why I think silently fixing the mistake of the caller is a bad idea

This PR ensures that text is never put below line 0. It will always be put at line 1. Nothing is silently breaking.

#### Why I won't investigate issue around `expand_params.clear_region ` myself:
1. `nvim_buf_set_text()` is buried deeply inside `luasnip.expand()`
2. `luasnip.expand()` is called by plugins without arguments
3. `luasnip.expand()` calls `nvim_buf_set_text` with arguments provided by internal code linked above; at certain circumstances they negate `clear_region` values
4. I'm not maintained and dunno `clear_region` intent and why it's mishandled

#### Why plugins has nothing to do with issue
Telescope-luasnip starts insertmode and then [simply calls `luasnip.expand()`](https://github.com/benfowler/telescope-luasnip.nvim/blob/master/lua/telescope/_extensions/luasnip.lua#L142-L152)

#### Why I won't implement checking
Given your resistance and constant arguing I suggest you merge this PR and fix region's validity yourself, cause, as I said earlier - I've no idea what `luasnip.expand()` does internally. I have no time to argue with you over reasoning, use cases, others weird considerations etc.. This is not serious.

This PR is obvious from the code itself.

L3MON4D3:
Can this be closed since https://github.com/benfowler/telescope-luasnip.nvim/pull/19 is merged?
We could have sorted this out much quicker if you would have told me what the underlying issue you're trying to solve is.

hinell:
@L3MON4D3 No, cause calling `luasnip.expand()` on empty first line of a new buffer may throw. Potentially - in other circumstances that I didn't investigate.  This is irrespective of the plugin...

L3MON4D3:
> No. Calling luasnip.expand() in insert mode may still cause nvim_buf_set_text() to throw when expanding snippet by using buffer content on a first line (?)

Very surely not the case, sorry

> or at some other circumstances (async calls to luasnip.expand() may be) I didn't investigate due to lack of time on hands.

If you know what the problem is/have something you can reproduce, feel free to open an issue

hinell:
Closing this PR won't prevent issues. In future, you will have to ensure `nvim_buf_set_text` calls are done properly elsewhere.  Otherwise you will get reports & PRs.

This is a waste of time.

PR #1033, opened 2023-10-06
Use libuv-file-watcher to update loaded snippet-collections.

L3MON4D3:
Currently we register an autocommands on BufWritePost for all files present when `load` is called. This has two undesirable side-effects:
* files that are added after the initial load are not loaded
* files are only reloaded by the neovim-instance that edited them

This PR is mainly for swapping the autocommand-mechanism for a libuv-based one, but will also include some refactorings of the loaders in general, for example the `cache`, which is a bit hard to grasp.
One disadvantage of the libuv-file-watcher is that it does not work for all directories, for example if a file on a nfs-share is edited by one machine, a different machine will not receive events for that edit (at least this is the case for my setup). This means we should probably have some kind of per-collection-fallback (maybe even back to autocommands), and/or a manual command for re-checking (which could also be called periodically..?)

One point I'm not yet sure about is whether snippets whose origin-file is removed should also be removed?

As of now, the mechanism works for the lua-loader, since it's the easiest one :D


bew:
Well, now nothing gets loaded ^^

On `main` running `LuaSnipListAvailable` gives all my snippets, but on this branch I get no snippet at all

L3MON4D3:
Oh xD
That shouldn't happen, could you try again with cca1293?

bew:
Running `nvim -u repro-init.lua` (the init file from my gist above) then calling `:lua require"luasnip".log.open()` gives me an empty buffer
(note: I deleted the log file before to make sure I see only the last logs)

L3MON4D3:
> gives me an empty buffer

Good hint! :D
Had a typo, path -> paths, which I didn't notice because my snippets are in a directory called "luasnippets", in rtp, which is loaded if `paths` (/`path`, previously) is not set.

> And I have an error everytime I save a Lua file (with my usual config):

Whoops, right, forgot to remove the autocommand-reload-mechanism for the lua-loader.. not sure how that didn't cause an error for me/how I missed it o.O

>  in from_lua.lua in LuaSnip's code you have a lot of mixed indentation, with tabs and spaces everywhere...

Really? Everything looks okay for me, eg. only tabs (that is, in neovim itself)

bew:
Hmm, I think you should try my repro code (adapted with your paths) as part of your test :grimacing:
`nvim -u repro-init.lua` immediately gives me this error:
```
Error detected while processing /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/init.lua:
E5113: Error while calling lua chunk: ...ugins/start/LuaSnip/lua/luasnip/loaders/tree_watcher.lua:97: attempt to index upvalue 'uv' (a nil value)
stack traceback:
        ...ugins/start/LuaSnip/lua/luasnip/loaders/tree_watcher.lua:97: in function 'tree_watcher'
        ...d-plugins/start/LuaSnip/lua/luasnip/loaders/from_lua.lua:177: in function 'new'
        ...d-plugins/start/LuaSnip/lua/luasnip/loaders/from_lua.lua:292: in function '_load'
        ...d-plugins/start/LuaSnip/lua/luasnip/loaders/from_lua.lua:298: in function 'load'
        /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/init.lua:19: in main chunk
```

L3MON4D3:
Oh right, `vim.uv` is new, no? Do you have that?

bew:
> Oh right, `vim.uv` is new, no? Do you have that?

Ah yes, I'm on latest stable `0.9.2`, don't really want to use nightly at the moment

L3MON4D3:
Sorry that you're a bit of a guinea-pig right now, normally the test-suite would catch all that, but it's pretty broken right now since I first want to get the new parts to work, and then look what stupid errors need fixing :D

L3MON4D3:
but maybe it works now :crossed_fingers:

bew:
`¯\_(ツ)_/¯`

No error anymore, but still nothing in logs 👀

bew:
With some print debugging it seems this function is never called..
https://github.com/L3MON4D3/LuaSnip/blob/e1fdca53fb61801365f785e86af411516c42f766/lua/luasnip/loaders/from_lua.lua#L181-L187

L3MON4D3:
> No error anymore, but still nothing in logs 👀

Huh!! :(
Oh, what's your loglevel, do you have it set to "info"?
`require("luasnip.util.log").set_loglevel("info")`
Make sure that's before any calls to `load/lazy-load`

> Maybe we can connect over jitsi some time, might be easier than adding tens of comments here

I think I'll just delete them once it works for you, we should be pretty close anyway

bew:
Indeed, snippets actually gets loaded, I think I was testing it wrong.
I tried one of my snippet, then edited the snippets in another neovim instance, it doesn't look like it gets reloaded in the original instance?
Here are the logs:
```
INFO  | lua-loader: Initialized snippet-collection at `/home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft`
INFO  | lua-loader: Loading lazy-load-snippets for filetype `lua`
INFO  | lua-loader: Loading lazy-load-snippets for filetype `all`
INFO  | lua-loader: Adding snippets for filetype `lua` from file `/home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft/lua.lua`
INFO  | loaders: Adding 5 snippets and 0 autosnippets from /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft/lua.lua to ft `lua`
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: 4913; change: nil; rename: true
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: 4913; change: true; rename: nil
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: 4913; change: nil; rename: true
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: lua.lua; change: nil; rename: true
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: lua.lua~; change: nil; rename: true
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: lua.lua; change: nil; rename: true
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: lua.lua; change: true; rename: nil
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: lua.lua; change: true; rename: nil
INFO  | tree-watcher: raw: root: /home/bew/tmpdir-large/nvim-luasnip-repro-for-1033/snips_for_ft; err: nil; relpath: lua.lua~; change: nil; rename: true
INFO  | lua-loader: Loading lazy-load-snippets for filetype ``
```

L3MON4D3:
Hah, just noticed that as well :D
Should work now!

bew:
Looks like it's working for the most common case now 👍

I might have an edge one: I use a `_utils.lua` file in the same directory as the snippets, with a number of helpers.
When I edit something in there / add a new helper, it's not available in the snippet files.

L3MON4D3:
Ah, that's an interesting case.
I think the easiest way to get that to reload would be using `loadfile` with the absolute path, since `require` (I assume you're using that right now?) caches the initial load forever, whereas `loadfile` (in neovim) uses a cache that updates on file-changes (so quick load on no changes, reload on changes)

Did this work correctly before? I'd be surprised if it did, but if it did, we'd have to replicate that :|

I'll assume you're by far not the only person to do something like this, so that's definitely something to add to the documentation

bew:
`loadfile` seems to work yes 👍

L3MON4D3:
Okay so we should be able to do this by substituting our own `loadfile` into snippet-files, and then build some kind of dependency-tree for the files.
Not sure if that won't end up being way overengineered though :D

L3MON4D3:
Okay was a bit wild to implement, and should be cleaned up a bit, but it should now be possible to load other files with `ls_tracked_dofile/dopackage` (one takes a path, the other a package-name, like require), and updating them should update the snippets that depend on their functions.
I've opted not to just replace require, since that would very likely have bad consequences (eg. reload all of luasnip if it is required in any snippet-files, which is still pretty likely) and since implementing all properties of require (same table passed to all who `require` it) would even more complex than what we have currently (and also not that useful, I doubt there is much state in these files)

L3MON4D3:
Changed some more things, there's now also an autocmd-backend for the filesystem-watcher, which is the lone default for now (so the libuv-watcher will only be used when specifically enabled in the `load`-call) (benefits of it are that it behaves correctly even if the filesystem/OS does not support the libuv-watcher, with the caveat that edits from outside the session will not be recognized)

Not sure if that should stay like this, enabling both would be more safe (less risk of missing changes), but may also be unnecessary (I, for example, seldomly edit snippet-files through another session, but almost always via `LuaSnipEditF` (think that's in `DOC.md`), and wouldn't need the libuv-watcher at all.. Not sure how many also have this workflow though :sweat_smile:)

L3MON4D3:
Alright, pretty sure this is done, I'm happy with tests, doc, etc.
@leiserfg (and others, ofc :D) Wanna take a cursory look? Or a closer look, if you have the time to spare? :sweat_smile:

lua/luasnip/session/snippet_collection/init.lua
```diff
@@ -254,7 +254,6 @@ function M.add_snippets(snippets, opts)
 				-- set source if it was passed, and remove from snippet.
```
bew, line 256: I think you need to update that comment?
```suggestion
				-- set source if it was passed.
```

plugin/luasnip.lua
```diff
@@ -83,20 +83,10 @@ require("luasnip.config")._setup()
...
-	end,
-	group = "_luasnip_lazy_load",
-})
-vim.api.nvim_create_autocmd({ "BufWinEnter", "FileType" }, {
-	callback = function(event)
-		require("luasnip.loaders.from_vscode")._load_lazy_loaded(event.buf)
+	callback = function()
+		require("luasnip.loaders").load_lazy_loaded(
+			tonumber(vim.fn.expand("<abuf>"))
+		)
```
bew, line 89: From the docs of [`nvim_create_autocmd`](https://neovim.io/doc/user/api.html#nvim_create_autocmd()):
image
So you should be able to do this:
```suggestion
	callback = function(event)
		require("luasnip.loaders").load_lazy_loaded(event.buf)
```

tests/helpers.lua
```diff
@@ -259,4 +259,58 @@ function M.check_global_node_refs(test_name, resolve_map, fn)
...
 end
 
+local scratchdir_path = ("%s/tests/scratch"):format(os.getenv("LUASNIP_SOURCE"))
+M.scratchdir_path = scratchdir_path
+
+function M.scratch_prepare()
+	-- clean (maybe a test was not able to clean up after itself) and re-create
+	-- scratch-directory.
+	os.execute(("rm -rf %s"):format(scratchdir_path))
+	os.execute(("mkdir %s"):format(scratchdir_path))
```
bew, line 269: You're not afraid of spaces in the scratchdir path?

bew:
With a quick look around, I'm surprised by the amount of new code in each `from_xxx` loaders.
And I see each have its own `Collection` object:
- https://github.com/L3MON4D3/LuaSnip/blob/6e1523f1591f3a63efd0d23d26b561e93c847a5e/lua/luasnip/loaders/from_lua.lua#L188
- https://github.com/L3MON4D3/LuaSnip/blob/6e1523f1591f3a63efd0d23d26b561e93c847a5e/lua/luasnip/loaders/from_snipmate.lua#L137
- https://github.com/L3MON4D3/LuaSnip/blob/6e1523f1591f3a63efd0d23d26b561e93c847a5e/lua/luasnip/loaders/from_vscode.lua#L267

Isn't there a way for you to refactor this Collection type to avoid duplication, and _only_ have loader-specific behavior as dedicated impl?
What's different between each Collection anyway?

lua/luasnip/session/snippet_collection/init.lua
L3MON4D3: Yes, TY!

plugin/luasnip.lua
L3MON4D3: Ohh, yeah good thing you point that out, accidentially just undid part of #1047 there :sweat_smile:

L3MON4D3:
> Isn't there a way for you to refactor this Collection type to avoid duplication, and _only_ have loader-specific behavior as dedicated impl? What's different between each Collection anyway?

So, the first big difference between the (lua and snipmate) and the vscode-loader is that the vscode-loader declares all files that belong to it, and to what filetype, in a manifest (package.json), while the other two scan a directory-tree, and the filenames determine the filetype.
Putting those two into one seems very impractical.

There's more potential with the lua and snipmate-loader, but those also have some differences, for example the handling of `extends` in snipmate needs quite a bit of code, while the lua-loader has to handle the `ls_tracked_dofile`-dependencies.
What is very similar in those two are `add_file`, `reload` and `stop`, but those methods don't amount to much compared to eg. `load_file`, or `new`, so merging them is probably not worth it.

L3MON4D3:
(What happened to your comment on the unquoted paths? Are they not actually an issue?)

PR #1043, opened 2023-10-17
Fix links to DOC.md in README.md

treequin:
Really tiny fix, the links to the configuration were outdated, and the link to DynamicNode led to FunctionNode instead.
Also, thank you for this plugin, I really like the Lua snippets.


L3MON4D3:
Ah, thank you!

PR #1047, opened 2023-11-04
Use event.buf directly to get bufnr

xu-cheng:
Due to the neovim's issue (https://github.com/neovim/neovim/issues/16416), using `vim.fn.expand` sometimes throws Keyboard interrupt error when pressing Ctrl-C (a common alternative to Esc).

This works around the above issue by simply getting the buffer number from event argument.


leiserfg (approved):

L3MON4D3:
Ahhh of course, thank you!

PR #1050, opened 2023-11-05
feat: publich to luarocks

mrcjkb:
### Summary

This PR is part of a push to get (neo)vim plugins on [LuaRocks](https://luarocks.org/labels/neovim),
so that they can be used with the WIP luarocks-based plugin manager, [`rocks.nvim`](https://github.com/nvim-neorocks/rocks.nvim).

* See [this blog post](https://mrcjkb.github.io/posts/2023-01-10-luarocks-tag-release.html), which follows up on [a series of posts](https://teto.github.io/posts/2021-09-17-neovim-plugin-luarocks.html) by @teto.

### Things done:

* Add a workflow that publishes tags to LuaRocks.org when a tag is pushed.

### Notes

* Tagged releases are installed locally and then published to LuaRocks.org.
  On PRs, a test release is installed, but not published.
* For the release workflow to work, someone with a LuaRocks.org account will have to add their [API key](https://luarocks.org/settings/api-keys) to this repo's [GitHub actions secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository).
* Due to a shortcoming in LuaRocks (https://github.com/luarocks/luarocks-site/issues/188), the `neovim` and/or `vim` labels have to be added  to the LuaRocks package manually (after the first upload), for this plugin to show up in https://luarocks.org/labels/neovim or https://luarocks.org/labels/vim, respectively.

__Adding the API key (screen shot)__
github-add-luarocks-api-key


L3MON4D3:
Hey, that's pretty cool!
Especially for luasnip, since we depend on jsregexp, which is available on luarocks, so the error-prone (or, somewhat platform-dependent) compilation-step can be circumvented, or at least outsourced, if luasnip itself is installed via luarock (Right? :sweat_smile:)

I don't have that much time right away, but I'll look into the blogposts you shared and this PR as soon as I have some free time on my hands :)

mrcjkb:
> (Right? 😅)

Kind of. By default, luarocks will use a platform-dependent compilation setup, but it is specified in the package's rockspec - and luarocks will output an appropriate error message if part of the toolchain required to build a dependency is missing.
That being said, it is possible [to host pre-built binaries](https://github.com/luarocks/luarocks/wiki/Hosting-binary-rocks), which we plan to implement for rocks.nvim, at least for dependencies that are more complicated to build (e.g. if they require a Rust or Node toolchain).

> I don't have that much time right away

Take your time.
I don't have much time for my side gigs this month either :smile:

L3MON4D3:
> Kind of. By default, luarocks will use a platform-dependent compilation setup, but it is specified in the package's rockspec (so end users shouldn't have to worry about it) - and luarocks will output an appropriate error message if part of the toolchain required to build a dependency is missing.
That being said, it is possible to host pre-built binaries, which we plan to implement for rocks.nvim, at least for dependencies that are more complicated to build (e.g. if they require a Rust or Node toolchain).

Cool, ok, as long as C can be compiled we should be fine :D

Is there a way to specify the version for dependencies? We currently don't support the newest release of jsregexp, but only 0.0.5

Also, do we need a rockspec? Or why don't we need one? :D

mrcjkb:
> Is there a way to specify the version for dependencies? We currently don't support the newest release of jsregexp, but only 0.0.5

Yes, we can specify something like `jsregexp >= 0.0.4, <= 0.0.5`. (what's the minimum version required?)

> Also, do we need a rockspec? Or why don't we need one? :D

The luarocks-tag-release action generates a rockspec based on GitHub metadata and the specified inputs.
In my experience, that works for most neovim plugins. In special cases, you can specify a rockspec template.

It can still be useful to have a `dev` rockspec in the repo root for people who want to install the latest head from the main branch, or if you want to use `busted` for tests.

L3MON4D3:
Ah, great.
We should probably fix jsregex == 0.0.5, it's the newest version with API we support :D

> The luarocks-tag-release action generates a rockspec based on GitHub metadata and the specified inputs.
In my experience, that works for most neovim plugins. In special cases, you can specify a rockspec template.

> It can still be useful to have a dev rockspec in the repo root for people who want to install the latest head from the main branch, or if you want to use busted for tests.

Makes sense, ty! Don't think we'll need that for now

mrcjkb:
> We should probably fix jsregex == 0.0.5, it's the newest version with API we support :D

Done :smile:

L3MON4D3:
Nice!
I'll look into making a release in a bit, so there actually is a luarock :D

L3MON4D3:
Dammit :|
https://github.com/L3MON4D3/LuaSnip/actions/runs/6849731673/job/18622507608

L3MON4D3:
I guess the rockspec-name should be lowercased?

mrcjkb:
I've pushed [a fix](https://github.com/nvim-neorocks/luarocks-tag-release/commit/5c55dc95abb6f09937ba3b82429994f17123d600). It should work if you rerun the workflow.

PR #1062, opened 2023-11-20
Use event.buf/event.file directly instead of vim.fn.expand

xu-cheng:
We don't need to use `vim.fn.expand(...)` to get the buffer or file for the autocmd callbacks. Instead, use `event.buf` or `event.file` directly from the callback function argument.

See also #1047.


L3MON4D3:
Nice, TY!
I'll assume that's all of them :D

PR #1066, opened 2023-11-22
Document ft_func with an example

Flimm:
`ft_func` and `load_ft_func` need to be configured together. This commit updates the example to include both settings at once.

I know it's a small change, but it took me quite some time to work this out. So I hope the example configuration will be helpful to others and will save them time.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1701,6 +1701,7 @@ Contains some utility functions that can be passed to the `ft_func` or
 
   ```lua
   ls.setup({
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Ah, sorry, it's hard to know wherebto omit for brevity and where just that could cost others time😅
Thank you for the change, I appreciate it!

Could you also add a comment that explains that without ft_func there will be no apparent effect on luasnips behaviour? (and squash, so it's one commit+doc-generation)

Flimm:
> Ping If you don't have time, I can do the small modifications as well

Yes, please! Thank you.

PR #1067, opened 2023-11-28
Update jsregexp to 0.0.6 [WIP]

L3MON4D3:
Title.
I don't think all package-managers/users will immediately run `make install_jsregexp` after an update, so to prevent breaking configs, we should support both 0.0.5 and 0.0.6.

I think the correct way to do this is to simply implement the api we need from 0.0.6 via 0.0.5, and expose that if the old version is detected, and otherwise just expose 0.0.6 (see `util/jsregexp.lua`). Since we need only like one function and three fields, that should be completely fine, and tests all pass with 0.0.5 and the facade.

I have not yet tested with 0.0.6, and have to find a good way to have both installed when doing tests, since ensuring 0.0.5-compatibility and preventing accidental breakage would be hard otherwise.

### DIFF
--- a/.github/workflows/luarocks.yml
+++ b/.github/workflows/luarocks.yml
@@ -15,4 +15,4 @@ jobs:
           LUAROCKS_API_KEY: ${{ secrets.LUAROCKS_API_KEY }}
         with:
           dependencies: |
-            jsregexp == 0.0.5
+            jsregexp >= 0.0.5, <= 0.0.6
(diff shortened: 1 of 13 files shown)


L3MON4D3:
After some more checking, turns out 0.0.6 does still support the same api as 0.0.5 :((
Wish I'd known that before looking into the compatibility-layer..

mrcjkb:
Don't forget to unpin the version in the luarocks workflow :smile:

mrcjkb:
> Right, yes! Does the luarock-notation support "either version 005 or 006, but no others"?

You can specify `>= 0.0.5, <= 0.0.6`.

PR #1072, opened 2023-12-02
add option to skip indent nodes

TwIStOy:


L3MON4D3:
Interesting!
What's the usecase?

TwIStOy:
Some LSP servers already take indens into consideration. In such cases, LuaSnip will add additional indents.

Before the option:

https://github.com/L3MON4D3/LuaSnip/assets/6359934/66ed9c61-8036-4707-b38b-f076fa8f7099

After the option:

https://github.com/L3MON4D3/LuaSnip/assets/6359934/f209df99-675c-4ef2-9a67-92b53e3920b7

Config in `nvim-cmp` can be:

```lua
snippet = {
  expand = function(args)
    local skip_indent_nodes = false
    if vim.api.nvim_get_option_value("filetype", { buf = 0 }) == "dart" then
      skip_indent_nodes = true
    end
    require("luasnip").lsp_expand(args.body, {
      skip_indent_nodes = skip_indent_nodes,
    })
  end,
},
```

L3MON4D3:
Ahh okay, pretty cool :+1:

lua/luasnip/init.lua
```diff
@@ -270,7 +270,8 @@ local function snip_expand(snippet, opts)
 	local snip_parent_node = snip:trigger_expand(
 		session.current_nodes[vim.api.nvim_get_current_buf()],
 		pos_id,
-		env
+		env,
+		opts.skip_indent_nodes
```
L3MON4D3, line 274: I think we should frame this the other way around, so instead of "skip indenting" with default false, we use just "indent" with default true.

L3MON4D3:
Could you also add this to the other documented options in `snip_expand` (waaayyyy at 
the bottom of `DOC.md` IIRC)? I think especially including the motivation with lsp would be pretty cool, could you do that?

TwIStOy:
All fixed!

DOC.md
```diff
@@ -3689,6 +3689,26 @@ These are the settings you can provide to `luasnip.setup()`:
       	return snip.insert_nodes[0]
       end
       ```
+    - `indent`: bool?, defaults to `true`. Whether LuaSnip will try to add
+      additional indents to fit current indent level in snippet expanding. This
+      option is useful when some LSP server already take indents into
+      consideration. In such cases, LuaSnip should not try to add additional
+      indents. If you are using `nvim-cmp`, sample config:
+      ```lua
```
L3MON4D3, line 3697: Very nice!!
Only our doc-converter seems to mess up here, could you try if adding a line between "..config:" and "```lua" fixes this?

TwIStOy: Seems good now!

L3MON4D3: Yup :ok_hand:

L3MON4D3:
Beautiful, could you squash the commits?

L3MON4D3:
Ah wait I can squash :D

PR #1091, opened 2023-12-17
feat(builtin): support vscode snippet variable CURRENT_TIMEZONE_OFFSET

masakichi:
`CURRENT_TIMEZONE_OFFSET` was added since [April 2023 (version 1.78)](https://code.visualstudio.com/updates/v1_78#_new-snippet-variable-for-timezone-offset) in vscode.

Also [Snippets in Visual Studio Code](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_snippet-syntax) the doc here mentions:

> CURRENT_TIMEZONE_OFFSET The current UTC time zone offset as +HH:MM or -HH:MM (example -07:00).

This PR is to support the variable in LuaSnip.

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*           For NVIM v0.8.0          Last change: 2023 December 14
+*luasnip.txt*           For NVIM v0.8.0          Last change: 2023 December 17
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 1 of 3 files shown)


L3MON4D3:
Hey, thanks for the contribution!
Could you exchange `os.data("%z")` with the more portable version described [here](http://lua-users.org/wiki/TimeZone)? Seems like windows doesn't support "%z" :/

L3MON4D3:
I think it'd be best to put that function into a new file, `util/time.lua`. But `util/util.lua` would also be fine

masakichi:
Hi @L3MON4D3 
Thanks for your review, I tried to update the code based on your feedback, could you please take a look again?

L3MON4D3:
Ah yes!
Looks perfect, thand you very much!

Issue #195, opened 2021-10-11
Adding auto import support


premell:
It would be great to have the same auto import functionality as you do with LSP expansions. For example in react, if I auto complete `useEffect` in LSP, `useEffect` gets auto imported. But when I use the luasnip snippet for `useEffect` it doesnt auto import.

Its probably a really hard feature to implement, and requires a lot of cooperation between you, the snippet engines and LSP. But I figured I'd make an feature request anyway since its a great feature, and because many people wonder about auto import, its good to have an issue regarding it. Just so people can see the progress towards auto import, or if its impossible, what the reasons are.

premell:
hmm I am not sure how I can use the ts server for this, since ts server has nothing to do with snippets. 

Considering the LSP suggestion has "Auto import from react" in the top, I'd assume that LSP has some api for imports. In that case, you could extend the snippet template with an optional field with the import, so that it doesnt break current snippets, also many snippets doesnt require imports. 

The problem is obviously that you have to change the snippet template which would make the snippet engines incompatible with other snippet managers. It's always really difficult create a new standard like this, but I am unsure of how to add auto imports otherwise. Maybe some way of extending 3rd party snippets, but that would make them very hard to maintain

L3MON4D3:
> I've tried to look at different code bases to see how they handle auto imports, but I am not familiar with LSP so its hard to follow. Btw what do you mean with querying all code actions?

Check [here] (https://github.com/neovim/neovim/blob/649b3160a10f0053747383f0703b0a5a94548570/runtime/lua/vim/lsp/buf.lua#L563) for the source of the code_action implementation for nvim, in the handler there is a loop going over the returned actions.

> "especially if multiple imports are possible". Cant you just make the import statements to an array in the snippet json, and loop over the array and applying the code action?

No :/
There may be code actions that have nothing to do with importing, or multiple Modules that provide this function (Rusts' Result iirc), importing all of them wouldn't make sense

L3MON4D3:
Going to close this, I don't think we'll add anything like this

weilbith:
I'm not yet 100% happy with it, but this is my current solution for an example of a React `useEffect`. It automatically adds the missing import statement (or nothing it it is already there) and also automatically inserts the dependencies based on the hook you wrote.

[...]

PR #1092, opened 2023-12-19
Allow callbacks via node-options.

L3MON4D3:
Currently the event-callbacks are only settable by providing the jump-index to the parent-node, which is pretty roundabout, and annoying for usage in eg. `fmt`.

```lua
s("qwer", {t'  ', i(1, "asdf", {node_callbacks = {[events.enter] = function()
	print("enter!!!")
end}})})
```

Calling it `node_callbacks` for now, since the option also applies to `snippetNode`, where  the key `callbacks` is already taken :/

Also, as of now the per-node-callbacks disable/override the parent-callbacks, maybe both should be executed?

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -172,6 +172,11 @@ Common opts:
   point in time, the snippet may contain each key only once. This means it is
   fine to return a keyed node from a `dynamicNode`, because even if it will be
   generated multiple times, those will not be valid at the same time.
+* `node_callbacks`: Define event-callbacks for this node (see
+  [events](#events)).  
+  Accepts a table that maps an event, e.g. `events.enter` to the callback
+  (essentially the same as `callbacks` passed to `s`, only that there is no
+  first mapping from jump-index to the table of callbacks).
 
 ## Api
 
@@ -3350,14 +3355,15 @@ The cache is located at `stdpath("cache")/luasnip/docstrings.json` (probably
 # Events
 
 Events can be used to react to some action inside snippets. These callbacks can
-be defined per snippet (`callbacks`-key in snippet constructor) or globally
-(autocommand).
+be defined per snippet (`callbacks`-key in snippet constructor), per-node by
+passing them as `node_callbacks` in `node_opts`, or globally (autocommand).
 
 `callbacks`: `fn(node[, event_args]) -> event_res`  
-All callbacks get the `node` associated with the event and event-specific
+All callbacks receive the `node` associated with the event and event-specific
 optional arguments, `event_args`.
 `event_res` is only used in one event, `pre_expand`, where some properties of
-the snippet can be changed.
+the snippet can be changed. If multiple callbacks return `event_res`, we only
+guarantee that one of them will be effective, not all of them.
 
 `autocommand`:
 Luasnip uses `User`-events. Autocommands for these can be registered using
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -247,6 +247,11 @@ Common opts:
     point in time, the snippet may contain each key only once. This means it is
     fine to return a keyed node from a `dynamicNode`, because even if it will be
     generated multiple times, those will not be valid at the same time.
+- `node_callbacks`: Define event-callbacks for this node (see
+    |luasnip-events|).
+    Accepts a table that maps an event, e.g. `events.enter` to the callback
+    (essentially the same as `callbacks` passed to `s`, only that there is no
+    first mapping from jump-index to the table of callbacks).
 
 
 API                                                         *luasnip-node-api*
@@ -3206,13 +3211,15 @@ The cache is located at `stdpath("cache")/luasnip/docstrings.json` (probably
 25. Events                                                    *luasnip-events*
 
 Events can be used to react to some action inside snippets. These callbacks can
-be defined per snippet (`callbacks`-key in snippet constructor) or globally
-(autocommand).
+be defined per snippet (`callbacks`-key in snippet constructor), per-node by
+passing them as `node_callbacks` in `node_opts`, or globally (autocommand).
 
-`callbacks`: `fn(node[, event_args]) -> event_res` All callbacks get the `node`
-associated with the event and event-specific optional arguments, `event_args`.
-`event_res` is only used in one event, `pre_expand`, where some properties of
-the snippet can be changed.
+`callbacks`: `fn(node[, event_args]) -> event_res` All callbacks receive the
+`node` associated with the event and event-specific optional arguments,
+`event_args`. `event_res` is only used in one event, `pre_expand`, where some
+properties of the snippet can be changed. If multiple callbacks return
+`event_res`, we only guarantee that one of them will be effective, not all of
+them.
 
 `autocommand`: Luasnip uses `User`-events. Autocommands for these can be
 registered using
--- a/lua/luasnip/nodes/node.lua
+++ b/lua/luasnip/nodes/node.lua
@@ -1,6 +1,5 @@
 local session = require("luasnip.session")
 local util = require("luasnip.util.util")
-local node_util = require("luasnip.util.util")
 local node_util = require("luasnip.nodes.util")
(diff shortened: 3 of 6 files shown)


L3MON4D3:
@weilbith You were interested in events on textNode/all kinds of node (see #195), would you motivate that again here so not all of the context of #195 is necessary? :D

weilbith:
I give it a try. 😄 

The current API decouples callback functions from their associated nodes. Nodes are references by some indexing mechanism. This is "hard" to get right, is prone error wen changing the snippet and is harder to read because the relation is not obvious. Being able to define callbacks directly at the node declaration solves all these issues. I think it is also fair to say that this tends to be an approach more and more "frameworks" follow. So it seem to be a good idea.

The second part is about the ability to define callbacks for any kind of node, also for non interactive ones. The stupid counter question is: why should there be an exception? It seems like there are valid use-cases to trigger callback functionality also for non-interactive nodes.
One practical example comes from the linked ticket which discusses an approach to apply language server actions on snippet nodes to do things like auto-importing a new symbol. Such functionality needs to know where in the buffer this text (node) is for the language server request. The current workaround to do this somewhat similar is by using a snippet "global" callback and find these nodes by specifically assigned keys. Having callbacks associated directly with any node resolves this issue very elegantly with a simple API and trivial usage with direct node access in the callback function.

---

@L3MON4D3 I must admit I'm unable to tell from the code. But in which order by which "event" do these node callbacks trigger now? 🤔 
Taking my very specific example order as I described in [this comment](https://github.com/L3MON4D3/LuaSnip/issues/195#issuecomment-1862533398) at the very bottom, would this work?

L3MON4D3:
ty!!
> The current API decouples callback functions from their associated nodes. Nodes are references by some indexing mechanism. This is "hard" to get right, is prone error wen changing the snippet and is harder to read because the relation is not obvious. Being able to define callbacks directly at the node declaration solves all these issues. I think it is also fair to say that this tends to be an approach that more and more "frameworks" follow. So it seem to be a good idea

Well put, I'd add that conceptually, one should be able to decouple the content of a snippet, the nodes, from the expansion-parameters, which are the other parameters passed to `s(...)`.
If callbacks have to be defined in the snippet-call, and not with the nodes, that distinction is weakened.

Regarding the textNode-callbacks: I have not yet implemented them😅
If I understand you correctly, they should trigger if they are "jumped over",  so in `{i(1), t"foo", i(2)}`, the enter/leave-callbacks of foo should be executed when the jump from i1 to i2 is performed? That might work, but I think that's hard to define generally, for example if insertNodes are not jumped through in textual order (`{i(2), t"foo", i(1)}`).
I think in general having to get nodes via key is "fine", (though  a bit better if it's avoidable).

Another idea for callbacks: you could make use of the existing infrastructure for functionNode, and just not return text, but do something else, eg. lsp-actions. That does not alleviate the issue with having to manually retrieve keyed nodes (right now, the only difficulty with exposing them to the callback is backward-compatibility) and you can get multiple nodes in the same callback, which is an inherent limitations of the event-callbacks.

weilbith:
>  one should be able to decouple the content of a snippet,

That is a really good point too! I used to do this and for some snippets I still do, also for re-usability sometimes. :+1:

> That might work, but I think that's hard to define generally, for example if insertNodes are not jumped through in textual order ({i(2), t"foo", i(1)}).

Yeah, that's what I was afraid of. So it can only work for nodes that have an explicit index. And we can't simply add such to text nodes or similar because that would be a huge breaking change to the API.
I still like the node callback improvement/feature. Though I currently have no use case for them. All my callbacks currently act on snippet level. But maybe one day I'll be happy to use it.

> Another idea for callbacks: you could make use of the existing infrastructure for functionNode, 

Thanks for the inspiration. I'm just not sure how this can work. But maybe you teach me.
First issue is the buffer location. I need to know were this piece of text is. Sticking to the original example of the related issue, if I have this piece of code (shortened into a single line) `useEffect(() => /* ... */, [])`, I need to know where `useEffect` is to get the import action for it. Similar with the second function parameter, the empty array. This gets the dependencies of the effect depending on the callback function I wrote as first parameter of the effect. I currently don't know how to get this buffer location information from within the function node itself. Furthermore, the function node would then return for example just `useEffect`? But I need this text already in the buffer before requesting the language server.
Second issue I have is the order. In the above snippet I can trigger the first action as soon as the `useEffect` call is in my buffer (ignoring the syncing delay to the server discussed in the issue). But the second action for the dependencies can only be triggered after I finished writing the callback function within the insert node. For this simple snippet this works because I can use the enter and leave event of the snippet (`[-1]`). And it might be very unlikely that I wanna trigger like three actions with two inserts in-between or similar.

I had the idea to use an insert node for this. That would give me a proper index and with the new node snippets, that would be great. The intention was to then just programatically jump to the next node within the callback of the node itself. But that doesn't work according to my experiments. Would be not really a "clean" way in terms of how the API is designed, but would do exactly what is needed. To some degree even more simple because the cursor would be at the right location and I can heavily simplify the code action request logic as I can use the native `vim.lsp.buf.code_action` function which works on the current cursor location.
But as I said, I tried to call `require('luasnip').jump(1)` within a callback function and it did nothing somehow. 🤷🏾

L3MON4D3:
> I currently don't know how to get this buffer location information from within the function node itself.

Yeah, you'd still have to resort to `get_keyed_node` for that, but we would be able to modify the behaviour of `functionNode` such that the passed callback is called with the nodes themselves, and not with their content.

> Furthermore, the function node would then return for example just useEffect?

Oh, no, I'd suggest to just use the functionNode as a callback, so it would not insert any text

```lua
s("foo", {i(1, {key=1}), i(2, {key=2}), t"bar", f(function(args) ... return "" end, {1,2})})
```

> Second issue I have is the order. In the above snippet I can trigger the first action as soon as the useEffect call is in my buffer (ignoring the syncing delay to the server discussed in the issue). But the second action for the dependencies can only be triggered after I finished writing the callback function within the insert node. For this simple snippet this works because I can use the enter and leave event of the snippet ([-1]). And it might be very unlikely that I wanna trigger like three actions with two inserts in-between or similar.

Ahh of course.. yes, for that the leave-callback is more appropriate :+1:

> I had the idea to use an insert node for this. That would give me a proper index and with the new node snippets, that would be great. The intention was to then just programatically jump to the next node within the callback of the node itself. 

Ah, so replace the textNode with an insertNode, so you can get the correct node in the callback, and then jump once more immediately?
Yeah, maybe, but afaict this would just avoid the `get_keyed_node`, right (ah, and simplify the lsp-callback)? Since `events.leave` on the insertNode before`useEffect`-insertNode is during the same interaction as `events.enter` on the `useEffect`-insertNode

> The intention was to then just programatically jump to the next node within the callback of the node itself

I *think* this may work if you `vim.defer_fn` it. Calling `jump` while another jump is being executed is not a good idea, but with defer it should work just as if you called it manually immediately after the initial jump was done.

>  And I have to "maintain" it across NeoVim core runtime changes

Feel ya xD

weilbith:
> I think this may work if you vim.defer_fn it. Calling jump while another jump is being executed is not a good idea, but with defer it should work just as if you called it manually immediately after the initial jump was done.

Thank you so much! That is amazing.
Using it plus this branch (using it locally), my code shrinks enormously and becomes very simple. Which I highly appreciate. 
There is one little downside and that is that the defer delay also for the language server request (you remember the sync issue) affects the user experience. Though most servers I tried so far work quite well. It is not blazing fast, but pretty great.

L3MON4D3:
> It is not blazing fast

We're not writing Rust here after all :P

Glad it works better!

weilbith:
> We're not writing Rust here after all :P

I'm afraid Rust would not solve this problem. 😉 

---

Thank you very much to implement this feature so fast. Looking forward to get it merged. 😊

L3MON4D3:
(I've decided to change this a bit from the initial version: now all callbacks that are associated with some node are executed, not just one of them)

PR #1097, opened 2023-12-26
Minor DOC.md modifications for panvimdoc lists

Ote-Leo:
Pandoc fails to parse markdown lists unless they come after an empty line. Thus the panvimdoc generated lists are crammed in a single paragraph.


L3MON4D3:
Ah nice, thank you!!

PR #1117, opened 2024-01-27
Typehinting luasnip

michaeldebetaz:
See #1110

As Luasnip uses lazy loading in the code logic, the type hinting chain breaks before getting to the user. This is the first step towards a better type hinted Luasnip, leveraging the `lua-ls` LSP server and its use of the annotations (e. g. `---@params name string`).

I don't see another way to ensure type hinting than to cast a type to the lazy table at the end of `lua/luasnip/init.lua`. But I am sure we can still improve the type hiniting from functions that are using too ofter an `any` type. Typically for the `opts` or for the `snippets`.

This first try put the type hints a bit all over the place, so the author can better understand to what part of the code they are linked. It would be advisable to put the `---@alias` in a `types.lua` file in the future, so that the type hints don't pollute too much of the code.

I apologize if it is sometimes messy, I wanted to do this first iteration quickly before I get lost into the type hinting madness ^^ I am of course available if you have questions or if I can help in any way (I'm sure you will 😅).

### DIFF
--- a/lua/luasnip/config.lua
+++ b/lua/luasnip/config.lua
@@ -11,7 +11,7 @@ local function set_snip_env(target_conf_defaults, user_config)
 	-- either "set" or "extend", make sure it does not appear in the final snip_env.
 	local snip_env_behaviour = user_config.snip_env.__snip_env_behaviour ~= nil
 			and user_config.snip_env.__snip_env_behaviour
-		or "extend"
+			or "extend"
 	assert(
 		snip_env_behaviour == "set" or snip_env_behaviour == "extend",
 		"Unknown __snip_env_behaviour, `" .. snip_env_behaviour .. "`"
@@ -34,7 +34,16 @@ end
 -- declare here to use in set_config.
 local c
 
+---@class UserConfig
+---@field ext_opts table<string, table>
+---@field history boolean
+---@field update_events string[]
+---@field updateevents string[]
+
+---@alias SetConfig fun(user_config: UserConfig)
+
 c = {
+	---@type SetConfig
 	set_config = function(user_config)
 		user_config = user_config or {}
 		local conf = vim.deepcopy(conf_defaults)
@@ -43,13 +52,13 @@ c = {
 		ext_util.clear_invalid(conf.ext_opts)
 		conf.ext_opts = ext_util.child_complete(conf.ext_opts)
 		user_config.ext_opts =
-			ext_util.child_complete(user_config.ext_opts or {})
+				ext_util.child_complete(user_config.ext_opts or {})
 		ext_util.child_extend(user_config.ext_opts, conf.ext_opts)
 
 		-- use value from update_events, then updateevents.
 		-- also nil updateevents, don't spill it into the main config.
 		user_config.update_events = user_config.update_events
-			or user_config.updateevents
+				or user_config.updateevents
 		user_config.updateevents = nil
 
 		set_snip_env(conf, user_config)
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -17,6 +17,9 @@ local luasnip_data_dir = vim.fn.stdpath("cache") .. "/luasnip"
 
 local log = require("luasnip.util.log").new("main")
 
+---@alias GetActiveSnip fun(): table|nil
+
+---@type GetActiveSnip
 local function get_active_snip()
 	local node = session.current_nodes[vim.api.nvim_get_current_buf()]
 	if not node then
@@ -39,6 +42,9 @@ local function match_snippet(line, type)
 	)
 end
 
+---@alias GetSnippets fun(ft: string|nil, opts: table|nil): table
+
+---@type GetSnippets
 -- ft:
 -- * string: interpreted as filetype, return corresponding snippets.
 -- * nil: return snippets for all filetypes:
@@ -56,6 +62,15 @@ local function get_snippets(ft, opts)
 	return snippet_collection.get_snippets(ft, opts.type or "snippets") or {}
 end
 
+---@class SnipInfo
+---@field name string
+---@field trigger string
+---@field description string
+---@field wordTrig boolean
+---@field regTrig boolean
+
+---@alias SnipInfoFunction fun(snip: Snippet): SnipInfo
+
 local function default_snip_info(snip)
 	return {
 		name = snip.name,
@@ -66,6 +81,9 @@ local function default_snip_info(snip)
 	}
 end
 
+---@alias Available fun(snip_info: SnipInfoFunction|nil): SnipInfo[]
+
+---@type Available
 local function available(snip_info)
 	snip_info = snip_info or default_snip_info
 
@@ -166,6 +184,7 @@ local function jump(dir)
 		return false
 	end
 end
+
 local function jump_destination(dir)
 	-- dry run of jump (+no_move ofc.), only retrieves destination-node.
 	return safe_jump_current(dir, true, { active = {} })
@@ -225,6 +244,16 @@ local function _jump_into_default(snippet)
 	return util.no_region_check_wrap(snippet.jump_into, snippet, 1)
 end
 
+---@class Snippet
+---@field name string
+---@field trigger string
+---@field trigger_expand unknown
+---@field description string
+---@field wordTrig boolean
+---@field regTrig boolean
+---@field captures table
+
+---@alias SnipExpand fun(snippet: Snippet, opts: table|nil): Snippet
 -- opts.clear_region: table, keys `from` and `to`, both (0,0)-indexed.
 local function snip_expand(snippet, opts)
 	local snip = snippet:copy()
@@ -305,8 +334,11 @@ local function snip_expand(snippet, opts)
 	return snip
 end
 
+---@class ExpandOpts
+---@field jump_into_func table
+
 ---Find a snippet matching the current cursor-position.
----@param opts table: may contain:
+---@param opts ExpandOpts|nil: may contain:
 --- - `jump_into_func`: passed through to `snip_expand`.
 ---@return boolean: whether a snippet was expanded.
 local function expand(opts)
@@ -390,6 +422,9 @@ local function expand_or_jump()
 	return false
 end
 
+---@alias LspExpand fun(body: string, opts: table|nil): nil
+
+---@type LspExpand
 local function lsp_expand(body, opts)
 	-- expand snippet as-is.
 	snip_expand(
@@ -440,6 +475,9 @@ local function change_choice(val)
 	session.current_nodes[vim.api.nvim_get_current_buf()] = new_active
 end
 
+---@alias SetChoice fun(choice_indx: integer): nil
+
+---@type SetChoice
 local function set_choice(choice_indx)
 	local active_choice =
 		session.active_choice_nodes[vim.api.nvim_get_current_buf()]
@@ -457,11 +495,15 @@ local function set_choice(choice_indx)
 	session.current_nodes[vim.api.nvim_get_current_buf()] = new_active
 end
 
+---@alias GetCurrentChoices fun(): string[]
+
+---@type GetCurrentChoices
 local function get_current_choices()
 	local active_choice =
 		session.active_choice_nodes[vim.api.nvim_get_current_buf()]
 	assert(active_choice, "No active choiceNode")
 
+	---@type string[]
 	local choice_lines = {}
 
 	active_choice:update_static_all()
@@ -523,6 +565,9 @@ local function active_update_dependents()
 	end
 end
 
+---@alias StoreSnippetDocstrings fun(snippet_table: any): nil
+
+---@type StoreSnippetDocstrings
 local function store_snippet_docstrings(snippet_table)
 	-- ensure the directory exists.
 	-- 493 = 0755
@@ -557,6 +602,9 @@ local function store_snippet_docstrings(snippet_table)
 	vim.loop.fs_write(docstring_cache_fd, util.json_encode(docstrings))
 end
 
+---@alias LoadSnippetDocstrings fun(snippet_table: table): nil
+
+---@type LoadSnippetDocstrings
 local function load_snippet_docstrings(snippet_table)
 	-- ensure the directory exists.
 	-- 493 = 0755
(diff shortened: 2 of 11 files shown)


lua/luasnip/util/parser/init.lua
```diff
@@ -7,6 +7,9 @@ local functions = require("luasnip.util.functions")
 local util = require("luasnip.util.util")
 local extend_decorator = require("luasnip.util.extend_decorator")
 
+---@class Parser
+---@field parse_snippet fun(context: table|number|string, body: string, opts: table|nil): table
+---@field parse_snipmate fun(context: table|number|string, body: string, opts: table|nil): table
```
bew, line 12: Note that class annotations are usually only used when there are fields with values, here it is not really a class but a module with only functions.
Also don't duplicate function signature between here and the docstring above each function, it'd be very annoying to maintain.

michaeldebetaz: Thanks for the feedback! I used class indeed in order to restrict the fields you can invoke from a module. Is there a way to the same with a table?

That's very helpful, because the starting point was pretty annoying to get done. I would be happy to implement the changes you want if you don't want to do it yourselves 👍

For the docstrings, I tried not to touch the original code too much, because I am not familiar enough with the code. I was hoping someone more expert would help tidy that up.

If there is a better way than submitting PRs to improve the type hints incrementally, I would be happy to follow. But I don't think we can avoid making those kind of roundtrips, because I am too much of a LuaSnoob for understanding what's going on in the code base.

You can count on me however to produce some starting point and do the chores, as it's time-consuming. 

I hope my contributions can be useful to you! Let me know how I can help 😊

L3MON4D3:
Hey @michaeldebetaz, really happy about the quick PR and you tackling this at all, a big Thank You! already for that :D
I'm a bit busy this week and the next and don't have much time to spend on luasnip, but I'll check your changes as soon as I have free time on my hands :)

michaeldebetaz:
Thank you for your kind words and thank you for helping me getting started.

Take your time of course, there is no rush. 

If my code is too messy (I rushed it sometimes I think), we can speak on discord and pair program. Anyway, I'm here to help 👍

bew:
Hey, I've been reading up on LuaLS annotations, and found `@module` did you try to use it?
https://luals.github.io/wiki/annotations/#module

From the 2nd example, it looks like you could define a lazy local var as the target module, and later fill it and use it. :thinking: 
This way the 'complexity' of lazyness could be only where the lazyness happens, and all other modules would be _clean_ without the need to define endless aliases and pseudo classes for modules.

lua/luasnip/util/util.lua
```diff
@@ -334,7 +336,9 @@ local function redirect_filetypes(fts)
 
 	return snippet_fts
 end
-
+---@generic T : table
```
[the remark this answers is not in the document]
michaeldebetaz: As I understand it, Lua only have table as a data structure. Like for your previous comment about class vs. table, I would say it boils down to how it is more convenient for you to typehint your code.

`any[] ` is the same as `table<integer, any>` (which is better than table and I prefer too the array type). But I think `any` should be avoided if we can restrict) narrow the type.

Do you have a better type? Usually if I let any or unknown, that means that I couldn't guess the type.

michaeldebetaz:
@bew Yeah! I saw it but didn't find a useful application in our case. Maybe you see something that I missed 💡

Anyway, I don't see a way around type casting/forcing when we use the lazy_table. It would be easier if we could capture a type with some kind of `typeof`, but no idea how to do that.

lua/luasnip/util/util.lua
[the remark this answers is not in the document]
bew: Well in this case `deduplicate` is very generic and handles any kind of list without value type restriction, there is not much we can do.
Maybe it's only used with some specific types, but that's beside the point of type hinting, typing a function should represent the function's interface and limitation not how it's going to be used.

When you have `T: table` it's basically `table<any, any>` so restricting to lists of any already explains that the function is limited to processing lists and not any kind of tables (as it's using `ipairs` to iterate and not `pairs`)
`¯\_(ツ)_/¯`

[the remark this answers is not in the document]
michaeldebetaz: Alright, let's change for `any[]` then 👍

lua/luasnip/config.lua
```diff
@@ -34,7 +34,16 @@ end
...
+---@class UserConfig
+---@field ext_opts table<string, table>
+---@field history boolean
+---@field update_events string[]
+---@field updateevents string[]
+
+---@alias SetConfig fun(user_config: UserConfig)
+
 c = {
+	---@type SetConfig
```
L3MON4D3, line 46: Is the SetConfig-alias necessary? `---@type fun ...` seems more direct :D

michaeldebetaz: I'm grateful you find the time to watch my contribution :) Don't hesitate if you wanna clear things up vocally, I am reachable 👍

Usually, I use alias when I need to use the same type twice. Isn't ConfigSetup used somewhere else?

L3MON4D3:
I've looked a bit into the lazy_table-issue, and it seems like `@generic`, used as shown [here](https://stackoverflow.com/questions/74336921/add-type-annotations-to-lua-function-that-merges-two-tables-and-make-the-lsp-tra), is a good solution.
```lua
---@generic T1: table<string, fun(): any>
---@generic T2: table<string, fun(): any>
---@param lazy_t T1
---@param lazy_defs T2
---@return T1 | T2
return function(lazy_t, lazy_defs)
...
end
```
One issue with this is that the function-arguments (to eg. `insert_node` in `require("luasnip").insert_node`), are not passed through, maybe it needs some more tricks, or luals just can't do that (yet?) :/

lua/luasnip/config.lua
L3MON4D3: Ah, maybe for `c.setup`? But that seems like something luals should be able to resolve automatically

> Don't hesitate if you wanna clear things up vocally, I am reachable 👍

Cool, I'll keep it in mind :+1:

L3MON4D3: (e.g. quick search didn't show any other usage :P)

michaeldebetaz: Sorry, I am on my phone and it's a bit difficult to search right now, but if it's alone, it's possible I deleted it and then didn't update the alias.

lua/luasnip/init.lua
```diff
@@ -39,6 +42,9 @@ local function match_snippet(line, type)
 	)
 end
 
+---@alias GetSnippets fun(ft: string|nil, opts: table|nil): table
```
L3MON4D3, line 45:
```suggestion
---@alias GetSnippets fun(ft: string|nil, opts: table|nil): table<string, Snippet[]>
```

Just a tad more information: the table maps filetypes to Snippets, like `{"all" = {snip1, snip2, ...}, "lua" = {snip3, ...}, ...}`, maybe it helps :D

lua/luasnip/init.lua
```diff
@@ -225,6 +244,17 @@ local function _jump_into_default(snippet)
 	return util.no_region_check_wrap(snippet.jump_into, snippet, 1)
 end
 
+---@class Snippet
```
L3MON4D3, line 247: Oh, this gave me a few thoughts:
* I think the annotations should only communicate official API, which for a snippet means relatively few fields, I think `trigger`, `name`, `description`, `captures` and `env` would be official, or are so commonly-used that they won't be changed, while stuff like `id` and `wordTrig` should not (yet) be relied upon.
Also, there are a few interesting functions for snippet, see [here](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#api-1)
* This is kind of a weird place to define `Snippet`, we should probably use a separate file for basic classes like that (see `lua/luasnip/_types.lua` or `lua/luasnip/extras/_extra_types.lua` for examples of this) (also, looking at them, it may make sense to namespace our classes :sweat_smile:)
* Not sure if you recall my little digression on snippet vs expandable vs addable in the discussion you opened, but `snip_expand` would take an `expandable`. But since this distinction exists in very few places, and is yet another thing newcomers would have to deal with, it's probably fine, especially for now, to just make most things `Snippet`

lua/luasnip/init.lua
```diff
@@ -225,6 +244,17 @@ local function _jump_into_default(snippet)
...
+---@field id integer
+---@field name string
+---@field trigger string
+---@field trigger_expand unknown
+---@field description string
+---@field wordTrig boolean
+---@field regTrig boolean
+---@field captures table
+
+---@alias SnipExpand fun(snippet: Snippet, opts: table|nil): Snippet
```
L3MON4D3, line 257: Maybe for the future: It would be cool to also have the `opts` to all the functions type-hinted :D
I think luals is capable of eg. suggesting keys if a table is passed, and that would help discovery of options

L3MON4D3: Ah, like you've done in `expand`, nice!

lua/luasnip/init.lua
```diff
@@ -613,6 +662,9 @@ local function unlink_current_if_deleted()
 	end
 end
 
+---@alias ExitOutOfRegion fun(node: table): nil
```
L3MON4D3, line 665: Yeah, I think we'll also need a `Node`-type, first for arguments like this, and second to communicate API (which is currently described [here](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#api))

lua/luasnip/init.lua
```diff
@@ -711,14 +768,22 @@ local function setup_snip_env()
...
 local function get_snip_env()
 	return session.get_snip_env()
 end
 
+---@type GetIdSnippet
 local function get_id_snippet(id)
 	return snippet_collection.get_id_snippet(id)
 end
 
+---@alias AddSnippets fun(ft: string, snippets: Snippet[], opts: table|nil): nil
```
L3MON4D3, line 784: Again, addable vs expandable vs snippet.
I think I'll take care of that distinction after this PR, I pretty much know what changes to make :D

lua/luasnip/init.lua
L3MON4D3: (correction: `id` can also be official :sweat_smile: Just saw that we have `get_id_snippet`, and that'd be pretty stupid without some way to get the `id`)

lua/luasnip/config.lua
L3MON4D3: Ahh okay :+1:

lua/luasnip/init.lua
michaeldebetaz: Yeah, I did in some places, but as it was too much of a task for a first PR and for my noobness with Luasnip too typehint the expected types!

lua/luasnip/util/extend_decorator.lua
```diff
@@ -8,8 +11,8 @@ local function default_extend(arg, extend)
 end
 
 ---Create a new decorated version of `fn`.
----@param fn The function to create a decorator for.
----@vararg The values to extend with. These should match the descriptions passed
+----@param fn The function to create a decorator for.
+----@vararg The values to extend with. These should match the descriptions passed
```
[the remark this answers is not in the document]
michaeldebetaz: Oh yeah, I don't know where comes from. Maybe because lua-ls was hinting an error? lua-ls uses `---@` to understand the types, so it throws an error if the declaration doesn't match what is expected.

lua/luasnip/init.lua
michaeldebetaz: Yeah, I read your explanation several times, but I guess I didn't end to understand it well if that didn't click when doing the work.

L3MON4D3: Yeah, it's a pretty fine distinction, and you need quite a bit of context to really motivate why it's useful to have

michaeldebetaz:
Hi @L3MON4D3 👋 I was away the pas weeks, but I am still available if I can help! What's the course of action so far? does my PR bring any help?

I tried to merge incrementally the solution from @bew (https://github.com/L3MON4D3/LuaSnip/pull/1117#discussion_r1468457076) but I see it creates a conflict with the autogenerated docs.

In any case, I am still here if I can help :)

L3MON4D3:
> does my PR bring any help?

Oh yeah, it certainly does :D

> I see it creates a conflict with the autogenerated docs.

Oh, don't worry about that, those will be generated anew anyway :)

Did you try my suggestion for the `lazy_table` on your end? Did you run into the same issue? It seems like it would reduce the burden of manually keeping things in sync a bit :D

michaeldebetaz:
I did work on it, but I can't find a way with my knowledge of leveraging lua-ls :/ I don't think it's avoidable. Type casting still is the best solution I found.

L3MON4D3:
Oh hey, sorry, I totally forgot to actually answer you :grimacing:

I'm fine with just typecasting, it's not very maintainable, but then again, there probably won't be many more big additions, so it won't cause too much additional work in the future, and will give us the (probably) best results right now :D

michaeldebetaz:
Okay, no problem at all :) what's the next step then? Should I go on with the PR? Or will you take it and adapt the changes with with your preferences before merging?

L3MON4D3:
Hey, back again after a break :sweat_smile:
I think I'll do a few of the changes that need more intimate knowledge of luasnip, the low-hanging but tedious fruits you've addressed nicely :)

michaeldebetaz:
That's great! Let me know if I can do anything to help. Otherwise we can close this PR after you're done with the implementation :)

tmillr:
Ahhh here it is. I was wondering why this plugin didn't have any types/completion. Since it's June now and there still aren't any types, perhaps this could be addressed in an incremental fashion? E.g. for the time being just start by implementing the most common/central interfaces for creating snippets (like `s()`, `i()`, and `t()`)?

Also FWIW, I have a decent amount of experience with emmylua/luals type annotations and can help or give an opinion if needed (though the luals impl can be a bit tricky and quirky at times, so the result is practically never going to be 100% perfect, especially in comparison to other languages that are statically-typed by default like rust or typescript). With that said, having types or some types/hints is usually better than having none at all (in which case you have to constantly cross-reference with a browser window or another document containing documentation).

L3MON4D3:
I'm still planning to finish this, but I don't have that much time for it right now :/
Sorry for not giving some kind of update @michaeldebetaz

@tmillr, feel free to take a look, I'd appreciate input from someone who has experience with lua-annotations :) (mine is limited to what I've done here in luasnip, which isn't much :D)

Issue #1118, opened 2024-01-28
error[lua loder from path]: attempt to call global 'k' (a nil value)


rathod-sahaab:
```lua
require("luasnip.loaders.from_lua").lazy_load({ paths = { "~/.config/nvim/LuaSnip/" } })
```

Manual import fixes this
```lua
local k = require("luasnip.nodes.key_indexer").new_key
```

L3MON4D3:
Ah, PR is merged, closing this :)

PR #1119, opened 2024-01-28
add missing key_indexer shorthand: Update default_config.lua

rathod-sahaab:
This fixes bug #1118  where k is nil, in case of lua load from folder as it's not declared explicitly by the user.


L3MON4D3:
Hey!
In case you're not aware, there is `snip_env`, a config-option which customizes the variables exposed in files loaded by the lua-loader (it's described [here](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#config-options)).
By setting it to `{k =  require("luasnip.nodes.key_indexer").new_key}`, you can also get this variable :)

But since key-indexer is such a nice tool, you're probably right in wanting it in the default-env too, so TY for this PR :D

L3MON4D3:
(Merged via cherry-pick in 04266a9)

Issue #1161, opened 2024-04-18
Snippets do not exit at exit node or i(0)


gokberkgunes:
After commit https://github.com/L3MON4D3/LuaSnip/commit/d9cb6ab976bca0bbf0ce790bc3c56af5ae80f4b8, when the user reach an exit nodes, the snippet does not exit. Two screenrecordings are given below to demonstrate problems.

![first](https://github.com/L3MON4D3/LuaSnip/assets/96139011/44c81192-65ef-40bd-92de-7dc5da6ebd3d)
![second](https://github.com/L3MON4D3/LuaSnip/assets/96139011/10524440-d3bf-4b2e-85ac-653df9039d88)

L3MON4D3:
Hi :)
Riiiight, I did not implement a substitute for this behaviour after switching from `history` to the more granular `keep_roots, link_roots, link_children`.

I believe the best fix for this would be to add a new option for toggling this behaviour, something like `exit_roots` (don't disable the snippet if it's a child-snippet), and have that default to `history` to restore the old behaviour.

PR #1162, opened 2024-04-18
Properly exit snippets when exit node is reached

gokberkgunes:
This pull request fixes the bug introduced in https://github.com/L3MON4D3/LuaSnip/commit/d9cb6ab976bca0bbf0ce790bc3c56af5ae80f4b8. Basically, when user reaches exit nodes, user can keep jumping to earlier nodes unless another action, e.g. creating another snippet outside of the current snippet.

This pull request fixes https://github.com/L3MON4D3/LuaSnip/issues/1161.

The logic behind commit that fixes nested snippet behavior, d48cb22f4849a5b0d48119020c97700322328767, appears flaky. Is there any better way to do this?

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -154,10 +154,6 @@ Since there would almost always be more jumps if the roots are linked, regular
 `<Tab>` would not work almost all the time, and thus `link_roots` has to stay
 disabled.
 
-If the snippet is not a child, but a root, it can be exited when reaching to
-last node, `$0`, if `exit_roots` is enabled. This setting may avoid unexpected
-behavior by disallowing to jump earlier (finished) snippets.
-
 # Node
 
 Every node accepts, as its last parameter, an optional table of arguments.
@@ -3506,8 +3502,11 @@ These are the settings you can provide to `luasnip.setup()`:
   [Basics-Snippet-Insertion](#snippet-insertion) for more context.
 - `link_roots`: Whether snippet-roots should be linked. See
   [Basics-Snippet-Insertion](#snippet-insertion) for more context.
-- `exit_roots`: Whether snippet-roots should exit at reaching at their `$0`.
-  See [Basics-Snippet-Insertion](#snippet-insertion) for more context.
+- `exit_roots`: Whether snippet-roots should exit at reaching at their last
+  node, `$0`. This setting is only valid for root snippets, not child snippets.
+  This setting may avoid unexpected behavior by disallowing to jump earlier
+  (finished) snippets. Check [Basics-Snippet-Insertion](#snippet-insertion) for
+  more information on snippet-roots.
 - `link_children`: Whether children should be linked. See
   [Basics-Snippet-Insertion](#snippet-insertion) for more context.
 - `history` (deprecated): if not nil, `keep_roots`, `link_roots`, and
--- a/DOC.md
+++ b/DOC.md
@@ -154,6 +154,10 @@ Since there would almost always be more jumps if the roots are linked, regular
 `<Tab>` would not work almost all the time, and thus `link_roots` has to stay
 disabled.
 
+If the snippet is not a child, but a root, it can be exited when reaching to
+last node, `$0`, if `exit_roots` is enabled. This setting may avoid unexpected
+behavior by disallowing to jump earlier (finished) snippets.
+
 # Node
 
 Every node accepts, as its last parameter, an optional table of arguments.
@@ -3502,11 +3506,14 @@ These are the settings you can provide to `luasnip.setup()`:
   [Basics-Snippet-Insertion](#snippet-insertion) for more context.
 - `link_roots`: Whether snippet-roots should be linked. See
   [Basics-Snippet-Insertion](#snippet-insertion) for more context.
+- `exit_roots`: Whether snippet-roots should exit at reaching at their `$0`.
+  See [Basics-Snippet-Insertion](#snippet-insertion) for more context.
 - `link_children`: Whether children should be linked. See
   [Basics-Snippet-Insertion](#snippet-insertion) for more context.
 - `history` (deprecated): if not nil, `keep_roots`, `link_roots`, and
-  `link_children` will bet set to the value of `history`.  
-  This is just to ensure backwards-compatibility.
+  `link_children` will be set to the value of `history`, and
+  `exit_roots` will set to inverse value of `history`. This is just to ensure
+  backwards-compatibility.
 - `update_events`: Choose which events trigger an update of the active nodes'
   dependents. Default is just `'InsertLeave'`, `'TextChanged,TextChangedI'`
   would update on every change.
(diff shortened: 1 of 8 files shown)


L3MON4D3:
Hmm, I don't think I'd want to approach this by a conditional in the exitNode-code, but rather by checking the current node after jumping, and then, if `history`(or `exit_roots`, which I propose in your issue) is set, simply `nil` the current node, which should make this code behave the same as before d9cb6ab.

What flaky logic are you referring to? Maybe I can shed a light on it, or improve it :)

gokberkgunes:
I've been trying to fix tests with current approach, but it fails to satisfy some conditions in tests. The flakiness is happening when the ExitNode:jump_into is introduced only by itself. So basically, when just https://github.com/L3MON4D3/LuaSnip/pull/1162/commits/5d736483b5aef80636ffc1d8dd489ccfc7f0c48b is applied to LuaSnip code, below behavior appears.


The node inside the function is skipped by introducing a snippet to there. I had to introduce https://github.com/L3MON4D3/LuaSnip/pull/1162/commits/d48cb22f4849a5b0d48119020c97700322328767 then https://github.com/L3MON4D3/LuaSnip/pull/1162/commits/7f405f93a36022ea40120727e605bd78aa22d32a to fix this problem.

I think my first try is getting too complicated, I will just get things together and put it away. I will instead try the solution you proposed.

lua/luasnip/config.lua
```diff
@@ -58,6 +58,7 @@ c = {
 		if user_config.history ~= nil then
 			conf.keep_roots = user_config.history
 			conf.link_roots = user_config.history
+			conf.exit_roots = not user_config.history
```
L3MON4D3, line 61: Ah, this also needs to be added in `default_config.lua`. (default true IIUC)

DOC.md
```diff
@@ -154,6 +154,10 @@ Since there would almost always be more jumps if the roots are linked, regular
 `<Tab>` would not work almost all the time, and thus `link_roots` has to stay
 disabled.
 
+If the snippet is not a child, but a root, it can be exited when reaching to
+last node, `$0`, if `exit_roots` is enabled. This setting may avoid unexpected
```
L3MON4D3, line 158: I'm not sure we should describe this option at that place, it isn't that strongly related to snippet-insertion as the other 3 that replaced `history`.
I think just shifting this description to the config-section should be fine (maybe with a short recap on snippet-roots, or a hint to snippet-insertion-section on tips for that)

L3MON4D3:
I think a test for this would be pretty nice, would you like to add it? I can give you some hints if you haven't worked with the neovim-test-framework before

gokberkgunes:
For sure I would like to add a test for this. I have no experience with test framework so please give me hints.

L3MON4D3:
Cool:
The general idea is to simulate key-inputs/lua-calls, s.t. a snippet in a desired state is produced, and then we take a "screenshot" (copy all visible lines) and compare it to a previously taken one to ensure consistency.

I'd suggest you start out by copying the test `exit_out_of_region activates last node of snippet-root` in `snippet_basics_spec.lua`, and modify it s.t. the new setting is tested.
You can generate a `screen:expect`-call via `screen:snapshot_util()`, it will be printed to stdout when running tests.
For running tests, I'd suggest you use `TEST_FILE=/home/simon/Code/luasnip/tests/integration/snippet_basics_spec.lua TEST_07=false TEST_09=false make test`, (and adjust the paths to your luasnip-repo), since that will run the tests for only nvim-master and not versions 0.7 and 0.9.
This is rather a rough overview, if you have any questions, ask here or on matrix :)

DOC.md
gokberkgunes: I resolved this, should be good now.

lua/luasnip/init.lua
```diff
@@ -159,8 +159,15 @@ end
 local function jump(dir)
 	local current = session.current_nodes[vim.api.nvim_get_current_buf()]
 	if current then
-		session.current_nodes[vim.api.nvim_get_current_buf()] =
-			util.no_region_check_wrap(safe_jump_current, dir)
+		local next_node = util.no_region_check_wrap(safe_jump_current, dir)
+
+		if session.config.exit_roots then
+			if next_node.pos == 0 and next_node.parent.parent_node == nil then
```
L3MON4D3, line 165: Ah, I checked the failing test: it's possible that next_node itself is nil (happens when a jump caused some error, then `safe_jump_current` may return nil), could you handle that here? Just making sure it's non-nil before accessing .pos should be enough :)

gokberkgunes: I checked the problem, and tried to implement what you suggested, but this did not help with tests. Removing exit_roots implementation allows tests to pass. I suspect if test needs to be updated. Tomorrow, I will check out the tests manually, then I will report back.

L3MON4D3: Some tests also rely on exit_roots=false, passing that in addition should make them pass

gokberkgunes: Is returning `false` and setting `current_node` to `nil` if we have no `next_node` OK?

gokberkgunes:
Sigh, all tests passed on my local machine but not here. Checking the tests again on my machine.

Also, for some reason running tests twice appears to make tests behave differently.

L3MON4D3:
Ugh, yeah that one test is notorious for being flaky, I have issues with it myself often :/
No idea how to make it pass deterministically :(

lua/luasnip/init.lua
L3MON4D3: I think we currently return `true` because there was a jump.. I think we'd best stick to that

tests/integration/session_spec.lua
```diff
@@ -392,6 +392,10 @@ describe("session", function()
 		jump(1)
 	end)
 	it("Deleting nested snippet only removes it.", function()
+		ls_helpers.session_setup_luasnip({
+			setup_extend = { exit_roots = false },
+			hl_choiceNode = true,
```
L3MON4D3, line 397: Can `hl_choiceNode` be omitted?

gokberkgunes: During my tests, it was not possible. I think this is because `before_each` function call adds `hl_choice = true`, then calling `ls_helpers.session_setup_luasnip` here discards this `hl_choice = true` setting.

However, if you want I can change snippet to work without `hl_choiceNode`.

gokberkgunes:
Addition of one more test where we test `exit_roots = false` left for this commit.

tests/integration/session_spec.lua
L3MON4D3: Ohhh, yeah ofc... you could also theoretically pass `exit_roots = false` in `before_each`, all test in `session` should work with that.
Actually, I think I'd prefer that, seems a bit cleaner :D

L3MON4D3:
Sounds good :+1:

L3MON4D3:
Hey, I missed your commits until now, sorry about that :sweat_smile:
I think this looks really good now, thank you very much for sticking through until the end :D
I'll squash all commits, that should be fine given that everything is pretty self-contained :)

PR #1167, opened 2024-05-06
Use lazyredraw to avoid flicking

xzbdmw:
based on https://github.com/hrsh7th/nvim-cmp/blob/8f3c541407e691af6163e2447f3af1bd6e17f9a3/lua/cmp/utils/feedkeys.lua#L15
before:

https://github.com/L3MON4D3/LuaSnip/assets/97848247/2dc8666b-0ab5-4bc6-a611-2182e1b7ea5b

after:

https://github.com/L3MON4D3/LuaSnip/assets/97848247/d722b3f2-bfb2-4b88-b1dc-78f566178234

This is especially obvious in nvim 0.10

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*            For NVIM v0.8.0            Last change: 2024 April 30
+*luasnip.txt*             For NVIM v0.8.0             Last change: 2024 May 06
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Hey!
TBH, I've never encountered this before, the "slight flicker" I refer to is when the cursor momentarily jumps next to the INSERT or VISUAL, not that the whole buffer fllickers.
Do you know of some way I could reproduce this?

xzbdmw:
Hmm it is caused by a plugin redrawing screen when I type enter

PR #1175, opened 2024-05-16
Use filetypes instead of langs from treesitter

leiserfg:


L3MON4D3:
Oh cool, I'd worried about having to hardcode something like this!
I'm a bit unsure about 1. `get_filetypes` returns a list, so  `return { ..get_filetypes(..)}` seems wrong, and 2. I don't know about availability (wrt nvim-version) of `get_filetypes`, could you make sure it's only called when available (`vim.F.if_nil` should work splendidly)

leiserfg:
Sure, we could fallback to 'lang' if missing, regarding returning the file types, I'm not sure if we should if should join them with '.' or keep only the first

lua/luasnip/extras/filetype_functions.lua
```diff
@@ -1,27 +1,34 @@
+local function fts_from_ts_lang(lang)
+	-- In case of someone using nvim <= 0.9
+	if vim.treesitter.language and vim.treesitter.language.get_filetypes then
+		return vim.treesitter.language.get_filetypes(lang)
+	end
+	return { lang }
```
L3MON4D3, line 6: I _think_ we should also return `lang` in the first case, there are probably users who have their latex-snippets as latex due to the current behaviour, and those would no longer be loaded like this.

Also, I hope that this does not cause issues where a filetype is loaded when it shouldn't be (manually disassociating them is not really possible), but since it's a better default, let's wait with making this configurable until there is a concrete issue-case.

leiserfg: I will added `lang` to the `fts` if it's not there already (most of the time it will be).

L3MON4D3: Sounds good :ok_hand:

L3MON4D3:
Yeaah that's the stuff :ok_hand: :D

PR #1177, opened 2024-05-18
Handmade list_contains

leiserfg:


leiserfg:
Seems like vim.list_contains is "new" so let's do it ourselves.

L3MON4D3:
Aw man, srry, I completely missed that you prepared a fix for this :(
Do you see any downsides with letting deduplicate deal with this later on?

leiserfg:
True, if there is a dedup in the pipeline then it's not a problem.

PR #1180, opened 2024-05-21
fix(util): Path.parent now works on windows (#1168)

### LINKED ISSUE #1168: Windows: friendly snippets not loaded
I added friendly snippets by the documentation
``` lua
 require("luasnip.loaders.from_vscode").lazy_load()
```
it works on my WSL enviornment but it doesn't work on Windows build of nvim,
Then I changed the config and force to use slash instead of backslash in the path
``` lua
 require("luasnip.loaders.from_vscode").lazy_load({paths = vim.fn.stdpath('data') .. "/lazy/friendly-snippets"})
```
then it works.

So is there any problem on the default behavior of path separator?

xudyang1:
On Windows, both forward slash `/` and backslash `\\` work as the path separator. Linux and MacOS can have backslash as a valid filename character.

Unit tests are also provided for each platform because `Path.parent` depends on the local variable `sep` which depends on `jit.os`.

Fixes #1168

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*             For NVIM v0.8.0             Last change: 2024 May 18
+*luasnip.txt*             For NVIM v0.8.0             Last change: 2024 May 22
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
--- a/lua/luasnip/util/path.lua
+++ b/lua/luasnip/util/path.lua
@@ -193,14 +193,72 @@ function Path.components(path)
 	return vim.split(path, sep, { plain = true, trimempty = true })
 end
 
-function Path.parent(path)
-	local last_component = path:match("%" .. sep .. "[^" .. sep .. "]+$")
-	if not last_component then
-		return nil
+---Get parent of a path, without trailing separator
+---if path is a directory or does not have a parent, returns nil
+---Example:
+---  On platforms that use "\\" backslash as path separator, e.g., Windows:
+---    Path.parent("C:/project_root/file.txt") -- returns "C:/project_root"
+---    Path.parent([[C:\project_root\file.txt]]) -- returns [[C:\project_root]]
+---
+---    -- the followings return `nil`s
+---    Path.parent("C:/")
+---    Path.parent([[C:\]])
+---    Path.parent([[C:\project_root\]])
+---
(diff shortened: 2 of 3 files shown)


L3MON4D3:
Hey, thanks for tackling that :)

This looks good (differentiating on definition-level, not at runtime is cool :D), and I'd like to propose two improvements/changes:
1. `parent` should retain the previous behaviour of removing the trailing sep from the path, that fits with the remainder of the current code (join always appends sep, for example)
2. Regarding testing: I like that you added tests for both platforms, but since these tests can be run on either platform, it would be cool to also do this. For example, you could (in the definition of `parent`) allow some global variable that is unlikely to be set during normal operations, maybe `___LUASNIP_TEST_SEP_OVERRIDE`, to override what we detect via `sep`, and then set that to `/` or `\` depending on which test is currently running

xudyang1:
> Hey, thanks for tackling that :)
> 
> This looks good (differentiating on definition-level, not at runtime is cool :D), and I'd like to propose two improvements/changes:
> 
> 1. `parent` should retain the previous behaviour of removing the trailing sep from the path, that fits with the remainder of the current code (join always appends sep, for example)
> 2. Regarding testing: I like that you added tests for both platforms, but since these tests can be run on either platform, it would be cool to also do this. For example, you could (in the definition of `parent`) allow some global variable that is unlikely to be set during normal operations, maybe `___LUASNIP_TEST_SEP_OVERRIDE`, to override what we detect via `sep`, and then set that to `/` or `\` depending on which test is currently running

If I understand that correctly:

1. the previous behavior of `parent` returns parent directory without trailing sep for files only, but for directories, for example `/home/user/`, it returns `nil` but not `/home`. To retain the previous behavior, `require("luasnip.util.path").parent("/somefile.txt")` should return `""` (empty string). Although it looks weird, this is a valid parent because empty string in lua is a truthy value. I will also add some docs to the `parent` function to clarity the above behaviors.
2. Instead of making tests platform dependent, we should make them path separator dependent so that the tests can even run on other platforms that use either forward slash or backslash as path separator.

L3MON4D3:
Nice, this looks great!
Thanks for adding the disclaimer on `C:` for windows, I doubt there will ever be a case where a snippet-repo lives in a root-directory, but this should find it :) (and if that ever happens, maybe we can just switch over to the fs-module built into neovim :) )

Thank you!!

PR #1189, opened 2024-06-03
Add spell-checker

leiserfg:


L3MON4D3:
Oho, great idea!
Do you know if there's some linux-tool for this too? Something to potentially call from the Makefile, so we have quicker feedback? (although, for spellchecking that doesn't seem that important since it'll likely only need to run once :D)

leiserfg:
Yes, the action is jus a wrapper of a cli. The PR is in draft because I need to check another one that is maybe better.

L3MON4D3:
Superseded by #1239.
Thanks for initiating this :D

PR #1192, opened 2024-06-05
Declare ls variable

raffaem:


L3MON4D3:
Ah, yes, nice, ty! :D

Issue #1200, opened 2024-06-27
Build breaks when make's path has spaces


mcustiel:
When updating the plugin using Lazy I get the following error because make is stored in a path containing spaces. 

![Screenshot 2024-06-27 at 07 06 47](https://github.com/L3MON4D3/LuaSnip/assets/3268370/bc3cbd38-2fba-4aea-b1c6-4f759423f23c)

Because of security reasons I am not admin of my work laptop, so this path is provided by the company I work on and I am not able to change it.

L3MON4D3:
Closed in #1201 (thank you for providing the fix :+1:)

PR #1201, opened 2024-06-27
fix: support spaces in make's path

mcustiel:
Fix for issue #1200 
Surrounded make variable with quotes


L3MON4D3:
Looks good, TY!

PR #1210, opened 2024-07-13
add "\" for wordtrig

yuukibarns:
I think it should not be difficult to add API for configuration of `wordTrig`, but I don't know how to do it.
When writing latex, I still need to use `\` for some uncommon symbols and I have to care about whether it will trigger any snippet, which is quite disturbing.


L3MON4D3:
Hey :)
I think you can also use `condition` for this (essentially, do the check you've modified in there), I'm not sure if it's such a good idea to change the definition of `wordTrig`

One issue with condition is that it's a bit unwieldly, but you can use [extend_decorator](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#extend-decorator) for this.

PR #1211, opened 2024-07-15
Improved syntax file and ftplugin for SnipMate snippets.

Vanillma:
Hi again :))
I've made a few changes to improve the syntax highlighting and folding behavior for code snippets.

I noticed that both the ftplugin and syntax files in vsnip and luasnip are copied from vim-snipmate, and the issue is that they both have syntax errors from the start. I had asked a question about this on Reddit (see link below).
[Link to Reddit question](https://www.reddit.com/r/vim/comments/1cxt55r/check_a_syntax_file/)

List of changes:

Changes to syntax file:
1. Removed the syntax for the "version" keyword. This is related to vim-snipmate, which has two different versions of the snippet engine.

2. Fixed a syntax error in the snipKeyword definition.
In regular expressions, there are two types of groups: capturing groups and non-capturing groups.
In Vim, since regular expressions are in magic mode by default, groups must be escaped with a backslash.
For better syntax highlighting, it's also recommended to use non-capturing groups because they don't store unnecessary substrings and are faster.

The capturing group syntax is:
   ```
   \(\)
   ```
And the non-capturing group syntax is:
   ```
   \%(\)
   ```
It's also important to note that the OR operator in Vim's magic mode regular expressions is:
   ```
   \|
   ```
We also don't need `me=s+8`. This prevents the "autosnippet" keyword from being highlighted correctly. Without it, everything works fine, so there's no need to write extra code. :)
See the screenshot for a visual representation of the changes.

**Changes to ftplugin file:**
1. I personally use spaces instead of tabs, and I think it's better to leave it to the user's default settings. So, I removed the part about using tabs.

2. I slightly improved the folding functionality. It now works for both files written with hard tabs and soft tabs.
Also, the words "expand," "priority," and comments are not folded.
I wanted to submit these changes to vsnip in the past, but I needed more time to learn regex in Vim. By the time I was ready, I was using luasnip, so I'm sending them to you instead. :))

**A few notes:**
* Please test them yourself. I've tried them on several files from the vim-snippet project.
* I also don't know if it's better by default to have the code folded or unfolded when the file is opened. So, I didn't use foldlevel in the ftplugin file.

I hope this is helpful. Thank you! :))
Before the changes:
Screenshot_20240715-193941_Termux
After the changes in the syntax file (pay attention to the problem with the word autosnippet, Because of this piece of code me=s+8)
Screenshot_20240715-194102_Termux
And examples after complete changes
Screenshot_20240715-194147_Termux
Screenshot_20240715-195248_Termux
Screenshot_20240715-200915_Termux

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*             For NVIM v0.8.0            Last change: 2024 June 28
+*luasnip.txt*             For NVIM v0.8.0            Last change: 2024 July 21
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
--- a/ftplugin/snippets.vim
+++ b/ftplugin/snippets.vim
@@ -5,10 +5,9 @@ if exists("b:did_ftplugin")
 endif
 let b:did_ftplugin = 1
 
-let b:undo_ftplugin = "setl et< sts< cms< fdm< fde<"
+let b:undo_ftplugin = "setl cms< fdm< fde<"
 
-" Use hard tabs
-setlocal noexpandtab softtabstop=0
+setlocal foldmethod=expr foldexpr=getline(v:lnum)=~'^#\\\|^e\\\|^p'?0:getline(v:lnum)!~'^\\t\\\|^\ \\\|^$'?'>1':1
 
 setlocal commentstring=#\ %s
 setlocal nospell
--- a/syntax/snippets.vim
+++ b/syntax/snippets.vim
@@ -8,14 +8,13 @@ syn match snipCommand '\%(\\\@<!\%(\\\\\)*\)\@<=`.\{-}\%(\\\@<!\%(\\\\\)*\)\@<=`
 syn match snippet '^snippet.*' contains=multiSnipText,snipKeyword
 syn match snippet '^autosnippet.*' contains=multiSnipText,snipKeyword
 syn match snippet '^extends.*' contains=snipKeyword
-syn match snippet '^version.*' contains=snipKeyword
 syn match snippet '^priority.*' contains=snipKeyword,priority
 syn match priority '\d\+' contained
 syn match multiSnipText '\S\+ \zs.*' contained
-syn match snipKeyword '^(snippet|extends|version|autosnippet|priority)'me=s+8 contained
+syn match snipKeyword '^\%(snippet\|extends\|autosnippet\|priority\)' contained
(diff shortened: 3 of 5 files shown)


L3MON4D3:
Hey, I didn't have time to look at anything until now, and now it's too late here to properly test this, I just already wanted to say Thank You for the clean PR, I really appreciate that you went into some detail on the changes because I've not written a syntax-file in my life :D (+1 for good screenshots, very nice :ok_hand:)
I'll doublecheck the changes tomorrow :)

L3MON4D3:
Hello again, had some time to look over the changes, AFAICT everything works just as you say :ok_hand:
I'll write and push a quick test to this branch so that we can sure the highlighting does not change from now :)

L3MON4D3:
Was a bit of a hassle (syntax/ftplugin files didn't load automatically for some reason) but here we are.
If you don't have ideas for other tests (right now I'm doing a very simple define colorscheme -> check the colors match, and check whether the folding works) I'll merge :)

Vanillma:
> Was a bit of a hassle (syntax/ftplugin files didn't load automatically for some reason) but here we are. If you don't have ideas for other tests (right now I'm doing a very simple define colorscheme -> check the colors match, and check whether the folding works) I'll merge :)


Hi, Sorry for the late reply. I've been migrating from Vim to NeoVim, which involved a lot of learning and setup. I committed to moving all my plugins to lazy.nvim this past week, and my editor was unusable for a few days.

Almost everything is working well with my new NeoVim configuration now.

I'm glad the changes are working as expected. Unfortunately, I still don't know anything about the test directory and how it works.

I tried running make test after installing many packages on Debian, but I couldn't figure out how it was supposed to help me.

So, I reverted to the manual approach and modified the syntax and ftplugin files directly. Then, I opened tests/data/syntaxtest.snippets in NeoVim, and it worked fine.

So I think it's ready for merge and I don't have any other ideas to make it better at the moment.

I'm still new to NeoVim, i'm sorry :))

L3MON4D3:
Oh, no worries about replying a few days later :P
All functional tests essentially work by giving input to the editor and then checking its state.
The tests I've written here just check that the highlight remains consistent and that folding works as expected, so if your manual inspection is fine, the tests should be enough :)
Thanks for working on this, this is a pretty nice addition :D

eeeXun:
Hi, I would like to know if folding could be optional? I personally don't like files folded when opening them.

L3MON4D3:
I think you can adjust `'foldlevel'` to make sure folds are opened initially

PR #1231, opened 2024-09-12
Add a way to get the current log path through lua, add time formatting to logs

Diegovsky:
So, for some specific debug functionality I'm adding to my config, I needed a way to get the log path used by LS, and figured other people might want to use it too.

Also, added time information to help debug other problems and/or improve understanding of when stuff happens.


L3MON4D3:
These changes are good, Thanks!
Also +1 for making the time-fmt configurable :)

PR #1239, opened 2024-10-03
Add spell checker action WordWarden

gevhaz:
Hi I was searching for something else in the issues and noticed you guys were [considering options](https://github.com/L3MON4D3/LuaSnip/pull/1189#issuecomment-2146066603) for spellcheckers in #1189. So I thought I'd suggest my own action. It basically does pre-processing by converting to HTML and removing code and links, then runs aspell on it. It's possible to run offline by using the script https://github.com/gevhaz/Word-Warden/blob/main/spellcheck.py

My action seems to discover more spelling issues than the one in #1189, for better or for worse (it can be annoying with too many false positives). I tried either adding them to a dictionary or 'correcting' them so that the job passes. In the case of words that are arguably 'code', such as as `trigEngine`, I fix the issue by adding backticks, since inline code will be removed in the preprocessing.

 These things can be very subjective and I'm not attached to any choice between 'fixing' a word or adding it to the dictionary.

If you think that the changes I made to the markdown files are an improvement, consider merging this PR. Otherwise feel free to close it or ask for changes.


L3MON4D3:
Hi :)
Sorry for taking so long to look at your PR, I was busy :sweat_smile:

This seems very usable, especially that there is a cli-tool so one doesn't have to wait for the action to process :+1:
One potential issue is changing the visible text of a link to a heading: it will be used to generate the link for the vim-help directly, so changing the text may break it. Can that, or only the potential for breaking, be avoided? Or is it already avoided? :D

@leiserfg WDYT?

gevhaz:
Hey, thanks for checking it!

The link in `DOCS.md` with the visible text 

~~~
`lsp-snippets-transformations`
~~~

looks like this in the generated docs:

```
|luasnip-`lsp-snippets-transformations`|
```

The backticks were added by me, and they explains why it breaks. I also agree that backticks is not the 'right' way to solve this 'spelling error'. Basically there are three other choices:

1. Add lowercase 'lsp' to the dictionary. The drawback is that you will not discover if someone writes lsp in lowercase from then on.
2. Uppercase LSP in the link ('LSP-snippets-transformations'). Would this work?
3. WordWarden also supports custom sed preprocessing as a last resort. You could allow the link to stay the same by doing a `s/lsp-snippets-transformations//` in such a preprocessing script. It can't be done by adding that full string to the dictionary since aspell doesn't support dashes in dictionary words.

I'll also check for similar broken link cases after hearing what solution you prefer.

leiserfg:
Option 2 seems fine to me, let's wait for @L3MON4D3

L3MON4D3:
Thank you for the detailed reply!

I agree, option 2 is fine :)
Especially because the link generated in the helptext is lowercased anyway (eg. `Variables-Environment Namespaces` -> `luasnip-variables-environment-namespaces`), so this doesn't even affect it.

gevhaz:
Great! I've updated the PR. I also double checked that there are no other similar issues.

leiserfg:
Is there a way to exempt links (the same way it happens with inline code)? So other links don't need to be changed (like vsnip.vim).

gevhaz:
It's possible to add a pre-processing sed script. So it would be possible to replace the pattern `[.*](` with `[](`, which would have the effect of excluding link texts. 

I think the easiest thing would be to just add `vsnip` and similar to the dictionary though. It's unlikely that you ever misspell something as 'vsnip' specifically in the future, so there is no harm in it. And then you will also continue to benefit from spellchecking in other cases. For example, I changed the link text in the following line:

```markdown
This can also be done much cleaner, with all the benefits that come with using a loader, by using the [loader for Lua](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#lua)
```

so that it's "Lua" instead of "lua".

L3MON4D3:
Yeah I'd agree, spellchecking in links ist useful, if they contain a Word that is misspelled it should be corrected.

gevhaz:
I've added a fixup commit which removes backticks from repo names. If it looks good, I can squash.

L3MON4D3:
Good choice, that seems more consistent👍
I'm happy, feel free to squash :)

gevhaz:
Done

L3MON4D3:
Awesome, thank you very much for the contribution! :)

PR #1240, opened 2024-10-07
Fix `apply_common_opts()` wrong substring when `opts.max_len` is set

Bekaboo:
When `opts.max_len` is set and `opts.max_len` > `#line_to_cursor`, `#line_to_cursor - opts.max_len + 1` will be negative which gives the wrong substring, see doc of `string.sub`


L3MON4D3:
Oh, yeah of course!
Thanks for fixing it!

PR #1252, opened 2024-11-10
fix(jsregexp): fix install_jsregexp on Windows

xudyang1:
Users on Windows may install neovim by many means:

1. `scoop install neovim`: `which nvim` shows `~/scoop/shims/nvim`
2. `winget install neovim` or manually download Windows installer and set env: `which nvim` shows `C:/Program Files/Neovim/bin/nvim`

Either way, we can use `lua51.dll` in the original `nvim` directory (`$(scoop prefix neovim)/bin` or `C:/Program Files/Neovim/bin/nvim`)

Care must be taken when path contains spaces. Quotes are needed to let gcc understand `L"C:/Program Files/..."` such directories.

> [!NOTE]
> Users may also need to set:
>   SHELL=C:/path/to/Git/usr/bin/sh.exe # if git sh.exe is not in PATH
>   .SHELLFLAGS=-c # if git sh.exe is not in PATH
>   CC=gcc # if default cc is not set to any compiler
>   NEOVIM_BIN_PATH=C:/path/to/Neovim/bin # if Neovim/bin is not detected by the Makefile

Related: #759 #986 #1130 #1190 #1220

PR #1255, opened 2024-11-12
fix(Makefile): defensively validate scoop output path

xudyang1:
`scoop prefix package` on Git Bash returns 0 and writes error to stdout even if scoop does not find the `package`. (<https://github.com/ScoopInstaller/Scoop/issues/6228>)

Both #1252 and this change will work if future scoop returns 1 for unknown `package`

---

#1252 will fail when the followings happen:

1. scoop still returns 0 for unknown `package` **AND**
2. `grep -v 'Could not find app path for'` fails by
  2.1. scoop changes error message of `scoop prefix package`, **OR**
  2.2. `scoop prefix neovim` returns a path that contains `Could not find app path for` (very rare)

The following change enables defensive path validation against the output. Works on either forward slash or backslash as path separator.

```diff
-               if (scoop prefix neovim | grep -v 'Could not find app path for') >/dev/null 2>&1; then \
+               if (scoop prefix neovim | grep '^[A-Z]:[/\\]') >/dev/null 2>&1; then \
```

### DIFF
--- a/.github/data/project-dictionary.txt
+++ b/.github/data/project-dictionary.txt
@@ -111,6 +111,5 @@ CC's
 usr
 gcc
 exe
-Makefile's
 MSYS
 SHELLFLAGS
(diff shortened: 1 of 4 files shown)


L3MON4D3:
Ahh I see, this should be more safe :+1:
TY!!

Issue #1260, opened 2024-11-18
Build breaks when make's path has spaces 


mcustiel:
Duplicate of #1200 
A recent update left an unquoted make command in the Makefile.

PR #1261, opened 2024-11-18
fix: support spaces in make's path

mcustiel:
Fix for issue #1260 
Surrounded make variable with quotes


L3MON4D3:
Oh, I'd assumed that make/gmake handles this case by itself since all examples in their documentation have this variable unquoted.
Thanks for the fix! :)

PR #1264, opened 2024-11-27
Add indent string conversion in fmt for multiline string

xudyang1:
- Problem:

The following code outputs spaces instead of tabs `'\t'`:

```lua
fmt([[
  line1 (no spaces after dedent)

    line3: -> '  line3' (two spaces)
      line4: -> '    line4' (4 spaces)
]], {}, { dedent = true })
```

The following does not output tabs:

```lua
fmt([[
  line1 (no spaces after dedent)

  \tline3: -> '\\tline3'
  \t\tline4: -> '\\t\\tline3'
]], {}, { dedent = true })
```

- Solution:

This PR adds option `indent_string` to `fmt`, default to empty string (disabled for backward compatibility).

Users can specify `indent_string` to let `fmt` convert them to tabs `'\t'`:

```lua
fmt([[
  line1 (no spaces after dedent)

  \tlin3: 1 indent ('\\t' -> '\t')
  \t\tlin4: 2 indent ('\\t\\t' -> '\t\t')
]], {}, { indent_string = [[\t]] })
```

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1526,6 +1526,23 @@ ls.add_snippets("all", {
   }, {
     repeat_duplicates = true
   }))
+  s("example5", fmt([[
+    line1: no indent
+
+      line3: 2 space -> 1 indent ('\t')
+        line4: 4 space -> 2 indent ('\t\t')
+  ]], {}, {
+    indent_string = "  "
+  }))
+  -- NOTE: [[\t]] means '\\t'
+  s("example6", fmt([[
+    line1: no indent
+
+    \tline3: '\\t' -> 1 indent ('\t')
+    \t\tline4: '\\t\\t' -> 2 indent ('\t\t')
+  ]], {}, {
+    indent_string = [[\t]]
+  }))
 })
 ```
 
@@ -1560,6 +1577,9 @@ any way, correspond to the jump-index of the nodes!
   	when passing multiline strings via `[[]]` (default true).
   * `dedent`: remove indent common to all lines in `format`. Again, makes
   	passing multiline-strings a bit nicer (default true).
+  * `indent_string`: convert `indent_string` at beginning of each line to unit
+        indent ('\t'). This is applied after `dedent`. Useful when using
+        multiline string in `fmt`. (default empty string, disabled)
   * `repeat_duplicates`: repeat nodes when a key is reused instead of copying
         the node if it has a jump-index, refer to [Basics-Jump-Index](#jump-index) to
         know which nodes have a jump-index (default false).
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*           For NVIM v0.8.0          Last change: 2024 November 27
+*luasnip.txt*           For NVIM v0.8.0          Last change: 2024 November 28
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
@@ -1462,7 +1462,7 @@ Simple example:
       s("example5", fmt([[
         line1: no indent
     
-          line3: two space -> 1 indent ('\t')
+          line3: 2 space -> 1 indent ('\t')
             line4: 4 space -> 2 indent ('\t\t')
       ]], {}, {
         indent_string = "  "
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*           For NVIM v0.8.0          Last change: 2024 November 18
+*luasnip.txt*           For NVIM v0.8.0          Last change: 2024 November 27
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
@@ -1459,6 +1459,23 @@ Simple example:
       }, {
         repeat_duplicates = true
       }))
+      s("example5", fmt([[
+        line1: no indent
+    
+          line3: two space -> 1 indent ('\t')
+            line4: 4 space -> 2 indent ('\t\t')
+      ]], {}, {
+        indent_string = "  "
+      }))
+      -- NOTE: [[\t]] means '\\t'
+      s("example6", fmt([[
+        line1: no indent
+    
+        \tline3: '\\t' -> 1 indent ('\t')
+        \t\tline4: '\\t\\t' -> 2 indent ('\t\t')
+      ]], {}, {
+        indent_string = [[\t]]
+      }))
     })
 <
 
@@ -1487,6 +1504,9 @@ any way, correspond to the jump-index of the nodes!
         when passing multiline strings via `[[]]` (default true).
     - `dedent`: remove indent common to all lines in `format`. Again, makes
         passing multiline-strings a bit nicer (default true).
+    - `indent_string`: convert `indent_string` at beginning of each line to unit
+        indent (’). This is applied after `dedent`. Useful when using
+        multiline string in `fmt`. (default empty string, disabled)
     - `repeat_duplicates`: repeat nodes when a key is reused instead of copying
         the node if it has a jump-index, refer to |luasnip-basics-jump-index| to
         know which nodes have a jump-index (default false).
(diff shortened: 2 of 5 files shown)


L3MON4D3:
Hi :)
Am I understanding correctly that the issue is essentially that the `[[]]`-strings can't represent `<Tab>` as `\t`?
What about `<C-v><Tab>`, which inserts a literal tab? Or a manual `([[...]]):gsub("\\t", "\t")`? Under which circumstances should a tab not be converted? (Mhmm, maybe if the string itself should insert a `"\\t"`)

Thinking about it a bit more, what do you think about `fmt` taking an argument which makes it handle control-characters like plain old `""`? So we convert `[[\t]]` to an actual tab, and `[[\\t]]` to `[[\t]]`. To me, it seems a bit simpler than sometimes converting tabs and sometimes not :)

xudyang1:
> Am I understanding correctly that the issue is essentially that the `[[]]`-strings can't represent `<Tab>` as `\t`?
> What about `<C-v><Tab>`, which inserts a literal tab? Or a manual

Yes, the issue is the difficulty in representing `\t` in `[[]]`, but it's beyond that.

When I tried to create a custom snippet using `fmt`, I would like it to auto convert the leading two-space indent to tab when snippet is expanded, so that neovim can determine how to represent indentation in a target buffer (configs related to `shiftwidth` and `expandtab`).

For example, let `T` be a real tab and `S` be a real space, and with lua config `vim.opt.shiftwidth=2`, `vim.opt.softtabstop=2` and `vim.opt.expandtab=true`, the following snippet:

```lua
-- cpp.lua for cpp
s({ trig = "test_if" }, {
  sn(1, fmt([[
SSSSif {} then
SSSSSS{}
SSSSend
  ]], { i(1), i(2) },{}))
})
-- when expanded, I would expect
---cpp
---if {} then 
---T{}
---end
-- after that, neovim would expand `T` to certain number of spaces based on configs `softtabstop` and `expandtab` (users may have autocmds to set different value of `softtabstop` for some language)
```

Currently, LuaSnip does not have spaces-to-tab conversion. I tried to explicitly use `\t`, but `[[\t]]` does not work.

- `<C-v><Tab>` inserts a true tab character, which works. However, the tedious and complicated key input may introduce unexpected mix of spaces and tabs as indentation, and debugging is difficult as white space characters are not rendered on the screen.
- `([[...]]):gsub("\\t", "\t"` converts all `"\\t"` to `"\t"`, not only those at the beginning.
- `Under which circumstances should a tab not be converted`: this PR only converts `indent_string` at the beginning of each line, so the following

```lua
-- S is a REAL space, T is a REAL tab
-- \t is just [[\t]]

-- with { indent_string="SS" }, example1 expands to:
---echo "hello\tworld\n"
---Techo "new tab \t"

-- with { }, example1 does nothing after dedent:
---echo "hello\tworld\n"
---SSecho "new tab \t"
s({ trig = "example1" }, {
  sn(1, fmt([[
SSSSecho "hello\tworld\n"
SSSSSSecho "new tab \t"
  ]], {},{ indent_string = "SS"}))
})

-- users may want to explicitly showing the indent by \t

-- with { indent_string="SS" }, example2 expands to:
---echo "hello\tworld\n"
---Techo "new tab \t"

-- with { }, example2 does nothing after dedent:
---echo "hello\tworld\n"
---\techo "new tab \t"
s({ trig = "example2" }, {
  -- S is space
  sn(1, fmt([[
TTSSecho "hello\tworld\n"
TTSS\techo "new tab \t"
  ]], {},{ indent_string = [[\t]]}))
})
```

- I am not sure letting `fmt` handle control character may be easier for this case. `[[\t]]` to `"\t"` may be expected, but `[[string.format("%s\n%s", ...)]]` to `[[string.format("%s]] .. "\n" .. [[%s", ...)]]` when expanded may be undesired:

```
string.format("%s
%s", ...)
```

I may not explain that clearly. This PR intends to ensure custom snippets can be formatted in a neat style when using spaces as indent when `expandtab=true`. (without explicit `\t` in front of each line or doing multiple text nodes `t({ "\t" })`)

L3MON4D3:
Ah, thank you, that helped me understand the issue :)

I honestly think the best workflow for handling this problem (restore tab-keypress from spaces when expandtab is set) is 
inserting real tabs. The debugging-issue you mention can be fixed by using `'listchars'`, for example
```vim
set listchars+=tab:→\ ,trail:␣,space:·
noremap <silent><leader>l :set invlist<cr>
```
which makes showing whitespace-characters one keybind away.

I'd also prefer real tabs because they can be differentiated from real spaces, in case they should be mixed in the snippet:
```c
int main(int argc,
·········char*[] argv) {
→    printf("Hello World!")
}
```
where the spaces are dots and the arrow+spaces are a tab
(not possible with `indent_string`, right?)

Can you see big downsides with this approach? I don't think it's that tedious to press `<C-v>`, a keybind most programmers should be comfortable with :P

xudyang1:
> Ah, thank you, that helped me understand the issue :)
>
> I honestly think the best workflow for handling this problem (restore tab-keypress from spaces when expandtab is set) is inserting real tabs. The debugging-issue you mention can be fixed by using `'listchars'`, for example
>
> ```viml
> set listchars+=tab:→\ ,trail:␣,space:·
> noremap <silent><leader>l :set invlist<cr>
> ```
>
> which makes showing whitespace-characters one keybind away.
>
> I'd also prefer real tabs because they can be differentiated from real spaces, in case they should be mixed in the snippet:
>
> ```c
> int main(int argc,
> ·········char*[] argv) {
> →    printf("Hello World!")
> }
> ```
>
> where the spaces are dots and the arrow+spaces are a tab (not possible with `indent_string`, right?)
>
> Can you see big downsides with this approach? I don't think it's that tedious to press `<C-v>`, a keybind most programmers should be comfortable with :P

I get your point, but the example is possible with `indent_string` because indent_string by default does no conversion (`indent_string = ""`)

```lua
fmt([[
int main(int argc,
·········char*[] argv) {
→    printf("Hello World!")
}
]], {}, {})
```

It only offers an opt-in conversion if the user intends to convert something (spaces or `[[\t]]`) to tab:

```lua
fmt([[
int main(int argc,
·········char*[] argv) {
····printf("Hello World!")
}
]], {}, { indent_string = "····"})
-- will become:
[[
int main(int argc,
→→·char*[] argv) {
→printf("Hello World!")
}
]] -- looks neat and it uses tabs for indent and spaces for alignment

-- but below is not possible, as every 4 spaces become a tab
[[
int main(int argc,
→·····char*[] argv) {
→printf("Hello World!")
}
]]
```

L3MON4D3:
Ahh good point :)

Okay, I'm convinced this is a worthwhile addition, thank you!

L3MON4D3:
Okay, I've looked over the code, this looks very solid, I like the tight loops instead of cobbling something together out of `:gsub`, very minimal :)
Tests also look great (thanks for adding some for dedent), I've no complaints at all

Thank you for implementing this feature!

xudyang1:
> Okay, I've looked over the code, this looks very solid, I like the tight loops instead of cobbling something together out of `:gsub`, very minimal :) Tests also look great (thanks for adding some for dedent), I've no complaints at all
> 
> Thank you for implementing this feature!

Thank you for merging this PR. The path I learned how to create custom LuaSnip snippets was from:

```lua
s("trigger", {
  t("text node "), i(1, "insert node"), t({ "", "this is next line" }), ...
})
```

to:

```lua
s("trigger", {
  sn(1, fmt([[
  text node {}
  this is next line
  ]], { i(1, "insert node") }, {}))
})
```

For multiline snippets, a `fmt` node avoids text nodes and insert nodes interleaving and provides an organized presentation. I really like how it uses a single string as the snippet skeleton and fills the organs with specific nodes, more concise than vscode's multi-string style.

L3MON4D3:
Oh yeah, 100%, `fmt` makes snippets much more readable :D
I also like to use `ls.parser.parse_snippet(nil, <str>)` for this when the snippet(Node) is trivial, for example

```lua
s("for", fmt([[
    for {} do
        {}
    end
]], {
    c(1, {
        parse(nil, "${1:k}, ${2:v} in pairs($3)"),
        parse(nil, "${1:k}, ${2:v} in ipairs($3)"),
        parse(nil, "${1:val} in $2"),
        parse(nil, "${1:i} = $2, $3"),
    }),
    i(2)
}))
```

PR #1279, opened 2024-12-30
Duplicate examples in `luasnip-extra-match` docs

caleskog:
Found duplicate examples and single-line code examples without explanations in `luasnip-extras-match`. This commit fixes this for better user readability.

### DIFF
--- a/DOC.md
+++ b/DOC.md
@@ -1349,7 +1349,7 @@ something other than strings).
 
 Examples:
 
-* `match(n, "^ABC$", "A")`, where `n = 1`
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Hi :)
I agree, this is better :+1:
Could you remove the "where n=1"-explanations? I think they confuse rather than help ("Why would these have to be 1?")

caleskog:
Yea, you're correct.
Its now been rectified.

L3MON4D3:
Much appreciated, thank you for the contribution :)

PR #1314, opened 2025-04-01
fix: avoid deprecated usage of `vim.validate`

carlosala:
`vim.validate` in its spec form is deprecated.
This fixes it.

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*           For NVIM v0.8.0           Last change: 2025 January 04
+*luasnip.txt*            For NVIM v0.8.0            Last change: 2025 April 01
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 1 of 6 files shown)


leiserfg (approved):

nenahp:
Wait... are you sure this is usable below v0.11.0? Just tested:
```
E5108: Error executing lua [string ":lua"]:1: opt: expected table, got string
stack traceback:
        [C]: in function 'error'
        vim/shared.lua: in function 'validate'
        [string ":lua"]:1: in main chunk
```

leiserfg:
If it's just a warning and does not work in 0.10 then we better don't do it. Not everybody uses arch/nixos and many people are still in 0.10 and even older.

zachfi:
Can the plugin be made to detect the version to straddle the api changes?

L3MON4D3:
Hi @carlosala, thank you for going ahead and implementing a fix, I've gone ahead and implemented a version of this that does the version-check @zachfi suggested.
Also a ty @leiserfg for taking a look :)

Thank you for your initiative, nonetheless, and please excuse my late response here :sweat_smile:

Superseded by #1338, I'll give you guys a chance to take a quick look at it before I merge it (probably 1-2 days)

carlosala:
Thanks for the great work!

Issue #1025, opened 2023-10-02
Feature request: Expose types of snippet globals for nicer development


MariaSolOs:
When writing custom snippets, it would be really helpful to use type annotations for `s`, `t`, and other global snippet constructors for better editing support :)

bew:
Related #378 (but incomplete)

L3MON4D3:
Oh yeah, for sure!
As far as I understand, this would basically be whatever is in `DOC.md`, but as emmylua-annotation, so lua-language-server can parse it?
If so, I'd say there is a choice to make:
* either just copy+transform what is currently in `DOC.md` to the source-files (which could be done relatively quickly, but having to keep both in sync manually seems suboptimal, and mistakes could happen), or
* create a proper build-step for extracting these definitions and inserting them into `DOC.md` (initial setup seems much harder, but this seems cleaner at least)

I'd honestly be fine with either option since I don't really anticipate many major additions to luasnip, so the additional long-time-work induced by the first option may not outweigh one-time work and complexity introduced by the second one (though the second one may be fun to figure out)

Unfortunately, I really don't use annotations at all though, so my motivation for doing either is pretty low :sweat_smile:

MariaSolOs:
@L3MON4D3 that's fair. I think that if no major API changes are predicted, then option 1 is fine.

Any reason why #378 hasn't been merged? Despite being incomplete, it's still an improvement.

L3MON4D3:
Only that it's marked as draft, and adding annotations for a small number of api-functions only seemed a bit weird.
But you're right, better that than nothing.
I'll look into getting it merged

MariaSolOs:
Sounds good. I think that splitting the work into smaller PRs will be less overwhelming than documenting the entire API at once (I think no one has the motivation for that hehe). 

I’m also happy to help with this btw!

L3MON4D3:
> Sounds good. I think that splitting the work into smaller PRs will be less overwhelming than documenting the entire API at once

Ah, yeah, probably also true :D

> I’m also happy to help with this btw!

Yay, hoped for just that :P
Best wait until I'm done with #941 (shouldn't be long now), lots of changes to the files where api is defined in there

UtkarshVerma:
Hi, I am also interested in this. Since the other PR is merged. Is there anything else blocking this?

bew:
[...]

PR #1318, opened 2025-04-05
doc(types): Allow `ls` to be typed for all static & lazy functions

bew:
This is the trick I mentioned in https://github.com/L3MON4D3/LuaSnip/issues/1025#issuecomment-2779569951

The difficulty with defining type hints for the lazy code was getting the lazily loaded documentation to be loaded by LuaLS & collect them in a single type that can be applied on `ls`.

The trick here is using the fact that LuaLS explores all the code, **even when it is never actually executed**.
This allows to type the lazy part, then we can have a type for the static part and finally merge the 2 for the finale `ls` value.

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*             For NVIM v0.8.0             Last change: 2025 May 02
+*luasnip.txt*             For NVIM v0.8.0             Last change: 2025 May 03
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Hey, finally getting to this :D

Your solution is pretty clever, it solves all issues I had with previous versions of this :)

I've gone ahead and rebased your changes on the current master, should I merge them or do you want to add some documentation in this PR already?

bew:
> should I merge them or do you want to add some documentation in this PR already?

Let's merge this, this was we have a solid minimal base on which we can incrementally contribute all the rest (by modules/topics/..)

L3MON4D3:
Alright :+1:

L3MON4D3:
Thank you for the quality contribution :)

PR #1320, opened 2025-04-05
Add editorconfig to have uniform settings for every contributors

bew:
TSIA

Reference: https://editorconfig.org
The file is automatically read by Neovim since a few versions already ✨

This is useful as I personally use 2 spaces for Lua indents, but this project uses tabs exclusively and I had to run `:set shiftwidth=2 noexpandtab` to avoid making a completely different formatting as the rest of the project.
Now I only have to open a Lua file, and Neovim is automatically configured to use tabs

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,4 +3,4 @@ root = true
 [*.lua]
 insert_final_newline = true
 indent_style = tab
-max_line_length = 70
+max_line_length = 80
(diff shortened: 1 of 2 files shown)


L3MON4D3:
Good idea!
I'll bump `max_line_length`/`textwidth` to 80, I think that's what I've been using so far and most longer comments adhere to it.
If you're fine with that, I'll merge :)

bew:
Yes, let's go ;)

L3MON4D3:
A'ight :)

PR #1338, opened 2025-05-03
fix: account for deprecated format of vim.validate

L3MON4D3:
Supersedes #1314, and stays backward-compatible via a version-check.


L3MON4D3:
Ah, good point, yeah I'll get on it :)

Issue #1220, opened 2024-08-06
Unable to install jsregexp (OS: Windows)


hrcHarry:
Hi, I have check [#569](https://github.com/L3MON4D3/LuaSnip/issues/569) but I have a slightly different situation.

OS: Windows 10
Neovim Version:
NVIM v0.10.1                                                                                                            
Build type: Release                                                                                                     
LuaJIT 2.1.1713484068

The installation of Luasnip via [Lazy](https://github.com/folke/lazy.nvim) gives the following message
```
git submodule init
process_begin: CreateProcess(NULL, uname, ...) failed.
process_begin: CreateProcess(NULL, pwd, ...) failed.
process_begin: CreateProcess(NULL, pwd, ...) failed.
git submodule update
make "CC=cc" "INCLUDE_DIR=-I'/deps/lua51_include/'" LDLIBS="" -C "deps/jsregexp"
make[1]: Entering directory `C:/Users/User/AppData/Local/nvim-data/lazy/LuaSnip/deps/jsregexp'
cc -I'/deps/lua51_include/' -O2 -fPIC -c jsregexp.c -o jsregexp.o
process_begin: CreateProcess(NULL, cc -I/deps/lua51_include/ -O2 -fPIC -c jsregexp.c -o jsregexp.o, ...) failed.

make[1]: *** [jsregexp.o] Error 2
make[1]: Leaving directory `C:/Users/User/AppData/Local/nvim-data/lazy/LuaSnip/deps/jsregexp'
make: *** [jsregexp] Error 2
```

Additional info:
Before installing Luasnip, gcc, g++, zig, and clang (Visual Studio) is installed on my machine.

Could anyone help? Thanks.

JohnWilliston:
I have the very same issue with a clean install of Neovim and clean pull of the plugin.

JohnWilliston:
Wow. And at the end of the day, it doesn't make one whit of difference. I _finally_ fought through all the nonsense necessary to get the make files and everything building without errors in the right directory and everything. The following screenshot is of me using the MSYS2 UCRT64 command window to do an uninstall and a reinstall in the very LuaSnip folder:

![2024-10-24 19_36_38-_c_Users_John_AppData_Local_nvim-data_lazy_LuaSnip](https://github.com/user-attachments/assets/44514de7-c0b8-47a3-9a2e-9c9993f872e7)

[...]

PR #1343, opened 2025-05-10
fix(Makefile): use %Git%/bin/sh.exe if no SHELL is provided on Windows

xudyang1:
Some Windows users installed Git for Windows by Windows installer and did not add `path_to_Git/bin/` to `PATH`. We can by default use `C:/Program Files/Git/bin/sh.exe` when `SHELL` is not set by the user.

Related: #1220  #1341


L3MON4D3:
LGTM, Thank you for another fix! :)

PR #1345, opened 2025-05-12
doc,loaders: add section on fixing diagnostics from lua-language-server.

L3MON4D3:
I've found a rather nice way of getting `lua-language-server` to shup up about missing globals (and even provide correct `go to definition`) in snippet-files for the lua-loader: One can define the `snip_env` like any other framework using a definition file, add it to a directory in `runtimepath`, and add the correct `---@module <path>` in the snippet-files, which results in correct definitions :)

While it would be nice to just put the default `snip_env` into `lua/luasnip/`, but there are issues: If one also puts a personal `snip_env` definition file in eg. `~/.config/nvim/lua/`, if some variable-names appear in both files (eg. `fmt`), `gd` will list both the `fmt` in our provided file, as well as the one in the personal file. So far, I couldn't find a way to exclude the provided file if it is not explicitly requested via `---@module`.

I'm opening this as a PR because I'd like to see if anyone has an idea on how to do this a bit more comfortably, and maybe have it working by just adding the `---@module`-line to snippet files, without thepotential drawbacks for users with custom `snip_env`.

### DIFF
--- a/.github/data/project-dictionary.txt
+++ b/.github/data/project-dictionary.txt
@@ -1,10 +1,11 @@
-personal_ws-1.1 en 109 utf-8
+personal_ws-1.1 en 113 utf-8
 AbsoluteIndexer
 Autosnippets
 Cfigure
 Cimg
 DeVries
 ECMAscript
+Env
 GIFs
 IDEs
 IndentSnippetNode
@@ -62,6 +63,7 @@ extmark
 extmarks
 filetype
 filetypes
+fmt
 fn
 functionNode
 functionNodes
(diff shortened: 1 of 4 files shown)


L3MON4D3:
Alright, as it turns out, this isn't quite as strightforward as I'd hoped: luals loads all definition-files it can find in the paths in `workspace.library` immediately, and globally, and not just for files with the `---@module` line at the top.

So, one would have to change the content of `workspace.library` during lsp setup to only include the definition-file when a snippet-file is being edited, for example via `.luarc.json`, I'd guess, which is not annoying as long as there is only maybe one big snippet repository.

PR #1282, opened 2025-01-07
feat(Makefile): enable auto detect compiler

xudyang1:
`CC` on Windows is usually not set to the correct compiler. #1252 needs manually defined `CC` to work (i.e., `make install_jsregexp CC=compiler`). Many users complained about this extra work (#1220 #1278 ).

This PR enables compilers detection in the following order:
- manually defined `CC`
- `gcc`
- `clang`
- `zig`

PR #1346, opened 2025-05-15
fix(Makefile): use correct git shell path and add semicolons

xudyang1:
### Problem:

PR #1343 did not use the correct git shell path as the space is not escaped.

### Solution:

1. Adding a backslash to escape the space does not fix the issue as `SHELL` still expands to `sh.exe` even after assignment. Instead of directly copying `$(WIN_GIT_SHELL)`, use `$(wildcard $(WIN_GIT_SHELL))` returned from the Makefile function to overwrite `SHELL`.

2. Similar to PR #1282 on `$(shell %COMMAND%)`, semicolons are added after each command to make sure `make` can call `CreateProcess` correctly:

- Without semicolon: calls `CreateProcess(NULL,%COMMAND%,...)`
- With semicolon: calls `CreateProcess(NULL,$(SHELL) $(.SHELLFLAGS) "%COMMAND%;",...)`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,9 @@ NVIM_0.9_PATH=${NVIM_PATH}/${NVIM_0.9_PATH_REL}
 ${NVIM_PATH}:
 # fetch current master and 0.7.0 (the minimum version we support) and 0.9.0
 # (the minimum version for treesitter-postfix to work).
-	git clone --bare --depth 1 https://github.com/neovim/neovim ${NVIM_PATH}
-	git -C ${NVIM_PATH} fetch --depth 1 origin tag v0.7.0
-	git -C ${NVIM_PATH} fetch --depth 1 origin tag v0.9.0
+	git clone --bare --depth 1 https://github.com/neovim/neovim ${NVIM_PATH}; \
+	git -C ${NVIM_PATH} fetch --depth 1 origin tag v0.7.0; \
+	git -C ${NVIM_PATH} fetch --depth 1 origin tag v0.9.0;
 # create one worktree for master, and one for 0.7.
 # The rationale behind this is that switching from 0.7 to master (and
 # vice-versa) requires a `make distclean`, and full clean build, which takes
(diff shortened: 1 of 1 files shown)


L3MON4D3:
Hi!
Would you add a comment which explains why (or mentions that) the `CreateProcess` fails with multiple lines? It might be forgotten when creating new build steps.

L3MON4D3:
Of course :)

xudyang1:
@L3MON4D3 I believe this PR is ready to be merged, as another user has provided positive feedback: https://github.com/L3MON4D3/LuaSnip/issues/1220#issuecomment-3024947809.

Also, what do you think about switching from Makefile's `$(shell ....)` function to the shell's native command substitution in the following section? I have not verified this change, so I'm leaving it as-is for now.

```diff
diff --git a/Makefile b/Makefile
index fff60df..9205201 100644
--- a/Makefile
+++ b/Makefile
@@ -157,8 +157,8 @@ test_nix: nvim install_jsregexp
 spellcheck:
 # grabbed from word-warden.
 	# Misspelled words will appear inside the following tests:
-	if [ -n "$(shell aspell --home-dir . --encoding=utf-8 --mode markdown --lang en_US --personal ./data/project-dictionary.txt list < README.md)" ]; then exit 1; fi;
-	if [ -n "$(shell aspell --home-dir . --encoding=utf-8 --mode markdown --lang en_US --personal ./data/project-dictionary.txt list < DOC.md)" ]; then exit 1; fi;
+	if [ -n "$$(aspell --home-dir . --encoding=utf-8 --mode markdown --lang en_US --personal ./data/project-dictionary.txt list < README.md)" ]; then exit 1; fi;
+	if [ -n "$$(aspell --home-dir . --encoding=utf-8 --mode markdown --lang en_US --personal ./data/project-dictionary.txt list < DOC.md)" ]; then exit 1; fi;
 
 spellcheck_interactive:
 	aspell --home-dir . --encoding=utf-8 --mode markdown --lang en_US --personal ./data/project-dictionary.txt check DOC.md;
```

PR #1348, opened 2025-05-25
fix: use non-deprecated overload of `vim.str_utfindex`

MariaSolOs:
The current use of `vim.str_utfindex` was deprecated in https://github.com/neovim/neovim/pull/30915. Now the encoding parameter is required and return value is an integer (instead of the table from the deprecated version corresponding to the `utf-32` and `utf-16` encodings).

### DIFF
--- a/lua/luasnip/util/util.lua
+++ b/lua/luasnip/util/util.lua
@@ -120,10 +120,13 @@ local function move_to_mark(id)
 	set_cursor_0ind(new_cur_pos)
 end
 
-local function bytecol_to_utfcol(pos)
-	local line = vim.api.nvim_buf_get_lines(0, pos[1], pos[1] + 1, false)
-	-- line[1]: get_lines returns table.
-	return { pos[1], vim.str_utfindex(line[1] or "", pos[2]) }
+local function str_utf32index(s, index)
+	if vim.fn.has("nvim-0.11") == 1 then
+		return vim.str_utfindex(s, "utf-32", index)
+	else
+		local utf32_index, _ = vim.str_utfindex(s, index)
+		return utf32_index
+	end
 end
 
 local function multiline_equal(t1, t2)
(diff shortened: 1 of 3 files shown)


L3MON4D3:
Hi :)
It seems like our code was a bit misleading, the pre-0.11 `str_utfindex` returns `utf32_index, utf16_index`, not the other way around (see [here](https://github.com/neovim/neovim/blob/a74e869ffa503cc9c2d21836e24fec7a7ffca147/runtime/lua/vim/lsp/util.lua#L122-L141)).
Since we only really need the utf32-index in both of these cases, could you extract the added version-check logic into one function in `util/util.lua`, maybe `str_utf32index`, and call that in both situations?

I'm also fine with me doing this, just let me know :D

MariaSolOs:
> the pre-0.11 `str_utfindex` returns `utf32_index, utf16_index`, not the other way around (see [here](https://github.com/neovim/neovim/blob/a74e869ffa503cc9c2d21836e24fec7a7ffca147/runtime/lua/vim/lsp/util.lua#L122-L141)). 

Oh yes, that has been the code I've been referring to but I clearly switched the order 🤦🏻‍♀️ 

> Since we only really need the utf32-index in both of these cases, could you extract the added version-check logic into one function in `util/util.lua`, maybe `str_utf32index`, and call that in both situations?

I can do that, but the existing code does seem to use the `utf16-index`, that's what `bytecol_to_utfcol` returns. Am I missing something?

> I'm also fine with me doing this, just let me know :D

I'm happy to do this, no worries :)

lua/luasnip/util/util.lua
```diff
@@ -120,10 +120,13 @@ local function move_to_mark(id)
 	set_cursor_0ind(new_cur_pos)
 end
 
-local function bytecol_to_utfcol(pos)
```
MariaSolOs, line 123: Removing this function as it is unused.

L3MON4D3: Oh yeah, good spot :+1:

MariaSolOs:
> Alright, looks good :D Thank You for the addition!

Thanks to you for the review!

PR #1353, opened 2025-06-17
Generate `DOC.md` from annotations

L3MON4D3:
Some preliminary progress to unifying annotations and information living currently in `DOC.md`.
This uses [`luals-mdgen`](https://github.com/L3MON4D3/luals-mdgen) to read code for rendering markdown from lua-codeblocks.

This PR will implement the framework and add annotation for the functions implemented in `lua/luasnip/init.lua`, and maybe one or two of the snippet-constructor functions so that there is a good idea on whether this is feasible for the entire documentation :D


lua/luasnip/nodes/choiceNode.lua
```diff
@@ -45,7 +49,47 @@ function ChoiceNode:init_nodes()
 	self.active_choice = self.choices[1]
 end
 
-local function C(pos, choices, opts)
+---@class LuaSnip.Opts.ChoiceNode: LuaSnip.Opts.Node
```
bew, line 53: Any specific reason this doc isn't on top of `local ChoiceNode = Node:new()`'s line above at L13  ?

It might work because class docs can be extended afaik, but it's weird to do so in the same file 🤔

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -45,7 +49,47 @@ function ChoiceNode:init_nodes()
...
+---end
+---```
+---Consider passing this override into `snip_env`.
+---@field node_callbacks table<"change_choice"|"enter"|"leave", fun(node:LuaSnip.Node)>
+---Specify functions to call after changing the choice, or entering or leaving
+---the node. The callback receives the `node` the callback was called on.
+
+---Create a new choiceNode.
+---@param pos integer Jump-index of the node (See
+---[Basics-Jump-Index](../../../DOC.md#jump-index)).
```
bew: Could we avoid relative path like this and instead work from root of repo, thus links would be only `DOC.md#jump-index` ?
We could even simplify to `#jump-index` since everything will be in the same file 🤔

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -45,7 +49,47 @@ function ChoiceNode:init_nodes()
...
+---@param choices table<integer, LuaSnip.Node|LuaSnip.Node[]> A list of nodes
+---that can be switched between. If a list of nodes is passed as a choice, it
+---will be turned into a snippetNode.
+---Jumpable nodes that generally need a jump-index don't need one inside the
+---choiceNode, they inherit its jump-index. Additionally, one should make sure
+---the cursor has a position to stop at inside every choice, since changing the
```
bew: random question: Is that still true? I know I can have a choice be a simple text node and the choice selection will stop on it already (no need for an i(1))

bew:
Glad to see this 🤩

Is it at any kind of ready state ?
Or what are you tweaking / what is still missing ?

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: Ah, that class is for the options that can be passed to `C`, which is a superset of the options that can be passed to any node (e.g. `C` additionally accepts `restore_cursor`). I think it makes sense to put those `.Opts`-classes immediately above the functions that accept them.

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: Ahhh, check out https://github.com/L3MON4D3/luals-mdgen/commit/5bfd01ab28a5b8939aa448639d3d63ec8a5663fd, `luals-mdgen` will parse the link and make it relative to the generated markdown file :D
That way, the destination is correct in the source file (and one could actually jump to it, if the editor supports it) and the documentation :)

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: Good point, that's an exception, I'll amend the doc :+1:

lua/luasnip/nodes/choiceNode.lua
bew: Aah sorry I misread and skipped the Opts part of the type ><

I like this idea of having a kind an 'opts' namespace for all types related to function options, I might steal that for my own configs :D

lua/luasnip/nodes/choiceNode.lua
bew: > and one could actually jump to it, if the editor supports it

I didn't think of this, it's actually a good idea 👍 nice!

DOC.md
```diff
@@ -747,22 +747,24 @@ ChoiceNodes allow choosing between multiple nodes.
...
-  - `restore_cursor`: `false` by default. If it is set, and the node that was
-    being edited also appears in the switched to choice (can be the case if a
-    `restoreNode` is present in both choice) the cursor is restored relative to
-    that node.  
-    The default is `false` as enabling might lead to decreased performance. It's
-    possible to override the default by wrapping the `choiceNode` constructor
-    in another function that sets `opts.restore_cursor` to `true` and then using
-    that to construct `choiceNode`s:
+`c(pos, choices, opts): LuaSnip.ChoiceNode`: Create a new choiceNode.
+* `pos: integer` Jump-index of the node (See [Basics-Jump-Index](../../../DOC.md#jump-index)).
```
bew: this link is wrong 👀

It should be:
```suggestion
* `pos: integer` Jump-index of the node (See [Basics-Jump-Index](#jump-index)).
```
🤔

lua/luasnip/init.lua
```diff
@@ -688,17 +848,20 @@ local function exit_out_of_region(node)
...
 end
 
 -- ft string, fts table of strings.
-local function filetype_set(ft, fts)
+function API.filetype_set(ft, fts)
 	session.ft_redirect[ft] = util.deduplicate(fts)
 end
 
-local function cleanup()
+---Clear all loaded snippets. Also sends the `"User LuasnipCleanup"`
```
bew: Why the d-quotes here?
```suggestion
---Clear all loaded snippets. Also sends the `User LuasnipCleanup`
```
(same comment for others like this)

doc/luasnip.txt
```diff
@@ -756,43 +756,44 @@ ChoiceNodes allow choosing between multiple nodes.
...
-choiceNodes’.
-
-As it is only possible (for now) to change choices from within the
-`choiceNode`, make sure that all of the choices have some place for the cursor
-to stop at!
-
-This means that in `sn(nil, {...nodes...})` `nodes` has to contain e.g. an
-`i(1)`, otherwise LuaSnip will just "jump through" the nodes, making it
-impossible to change the choice.
+`c(pos, choices, opts): LuaSnip.ChoiceNode`: Create a new choiceNode. * `pos:
```
bew: Very bad generation here, this is completely unreadable now 😬

doc/luasnip.txt
```diff
@@ -3626,189 +3627,221 @@ GENERAL ~
...
-    - `strict`, `bool?`: If set, throw an error if the node under the cursor can’t
-        be jumped into. If not set, fall back to any node of the snippet and enter
-        that instead.
-    - `select`, `bool?`: Whether the text inside the node should be selected.
-        Defaults to true.
+- `jump(dir): boolean`: Jump forwards or backwards
+    - `dir: (1|-1)` Jump forward for 1, backward for -1.
+    This function returns:
+    - `jumped: boolean` `true` if a jump was performed, `false` otherwise.
+- `jump_destination(dir): LuaSnip.Node`: Find the node the next jump will end up
```
bew: I think each function should be separated by blank line, and actually have a help tag to help jump to that specific doc 🤔

lua/luasnip/nodes/node.lua
```diff
@@ -6,8 +6,30 @@ local events = require("luasnip.util.events")
...
+---node.
+---This is useful for [Key Indexer](../../../DOC.md#key-indexer) or for finding the node at
+---runtime (See [Snippets-API](../../../DOC.md#snippets-api)
+---These keys don't have to be unique across the entire lifetime of the snippet,
+---but every key should occur only once at the same time. This means it is fine
+---to return a keyed node from a dynamicNode, because even if it will be
+---generated multiple times, the same key not occur twice at the same time.
+---@field node_callbacks table<"enter"|"leave", fun(node:LuaSnip.Node)>
+---Specify functions to call after changing the choice, or entering or leaving
+---the node. The callback receives the `node` the callback was called on.
```
bew: note: For big multiline field documentations I think it'd help readability (a topic I love to think about :D) to have a blank line to separate the fields 🤔
```suggestion
---to return a keyed node from a dynamicNode, because even if it will be
---generated multiple times, the same key not occur twice at the same time.
---
---@field node_callbacks table<"enter"|"leave", fun(node:LuaSnip.Node)>
---Specify functions to call after changing the choice, or entering or leaving
---the node. The callback receives the `node` the callback was called on.
```
WDYT?

### Alternative

An alternative could be to just indent a bit the continuation documentation lines?

```suggestion
---    to return a keyed node from a dynamicNode, because even if it will be
---    generated multiple times, the same key not occur twice at the same time.
---@field node_callbacks table<"enter"|"leave", fun(node:LuaSnip.Node)>
---    Specify functions to call after changing the choice, or entering or leaving
---    the node. The callback receives the `node` the callback was called on.
```

lua/luasnip/nodes/snippet.lua
```diff
@@ -66,8 +66,22 @@ local stored_mt = {
...
 local Snippet = node_mod.Node:new()
 
+-- very approximate classes, for now.
+---@alias LuaSnip.SnippetID integer
+
+---@class LuaSnip.Addable
+---Anything that can be passed to ls.add_snippets().
+
+---@class LuaSnip.ExpandedSnippet
+---Represents an expanded snippet.
```
bew: Documentation should be _above_ the class to appear in lsp hover popup
```suggestion
---Anything that can be passed to ls.add_snippets().
---@class LuaSnip.Addable

---Represents an expanded snippet.
---@class LuaSnip.ExpandedSnippet
```
(with lua-language-server anyway)

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -9,6 +8,11 @@ local session = require("luasnip.session")
 local sNode = require("luasnip.nodes.snippet").SN
 local extend_decorator = require("luasnip.util.extend_decorator")
 
+---@class LuaSnip.Choice: LuaSnip.Node
```
bew: Having Choice & ChoiceNode be different and still both inheriting from Node feels very wrong 😬
Maybe rename that one to ChoiceEntryNode?

lua/luasnip/init.lua
```diff
@@ -66,19 +71,38 @@ local function default_snip_info(snip)
...
+---```lua
+---function(snip)
+---    return {
+---        name = snip.name,
+---        trigger = snip.trigger,
+---        description = snip.description,
+---        wordTrig = snip.wordTrig and true or false,
+---        regTrig = snip.regTrig and true or false,
+---    }
+--end
```
bew:
```suggestion
---end
```

lua/luasnip/init.lua
```diff
@@ -176,28 +206,49 @@ local function jump(dir)
...
+---cursor-position, or whether it's possible to jump forward from the current
+---node.
+---@return boolean is_expand_or_jumpable
+function API.expand_or_jumpable()
+	return API.expandable() or API.jumpable(1)
 end
 
-local function in_snippet()
+---Determine whether the cursor is within a snippet.
+---@return boolean is_inside
```
bew: What is this `is_inside` 'keyword' that I saw you added ~everywhere for returns? is that a new thing from the Rust-based Lua lsp?
(I find it very weird here and in the generated docs, and feels redundant with the surrounding doc strings)

lua/luasnip/init.lua
```diff
@@ -66,19 +71,38 @@ local function default_snip_info(snip)
...
+---    return {
+---        name = snip.name,
+---        trigger = snip.trigger,
+---        description = snip.description,
+---        wordTrig = snip.wordTrig and true or false,
+---        regTrig = snip.regTrig and true or false,
+---    }
+--end
+---```
+---@return table<string, T[]> available_info Table mapping filetypes to list of
```
bew: I guess you don't know, but there's a shorthand way to write that kind of type: `{[string]: T[]}`
(this is what I use everywhere in my configs, I like that it keeps the type short and not repeating `table` everywhere)

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -45,7 +49,49 @@ function ChoiceNode:init_nodes()
...
+---```
+---Consider passing this override into `snip_env`.
+---@field node_callbacks table<"change_choice"|"enter"|"leave", fun(node:LuaSnip.Node)>
+---Specify functions to call after changing the choice, or entering or leaving
+---the node. The callback receives the `node` the callback was called on.
+
+---Create a new choiceNode.
+---@param pos integer Jump-index of the node (See
+---[Basics-Jump-Index](../../../DOC.md#jump-index)).
+---@param choices table<integer, LuaSnip.Node|LuaSnip.Node[]> A list of nodes
```
bew: Why this type instead of a `(LuaSnip.Node|LuaSnip.Node[])[]` ? 🤔

bew (commented):
Another partial pass
It's pretty neat the way you insert stuff in the doc template by writing moar Lua code ^^

DOC.md
L3MON4D3: Ah, yeah, the doc-generation workflow is still the one from `master`, so it does not include the newer changes yet😬
I'll see if I can run this locally and push, I'm currently away from my PC

doc/luasnip.txt
L3MON4D3: Ah, yeah that's also an issue that should (at least partially) go away with the newer versions of the doc-generation.
panvimdoc, which generates the vimhelp needs an empty line before lists, I didn't respect that initially. Maybe this is also something we should report upstream

doc/luasnip.txt
L3MON4D3: Mhmmmmm yeah that'd be optimal, we'd only need `##` headers with the function name... Doable!

lua/luasnip/nodes/node.lua
L3MON4D3: Code is read more often than it is written☝️ :D
I agree, empty lines give a bit more readability👌

lua/luasnip/init.lua
L3MON4D3: Ah, that's just for giving the return-values names. Maybe a bit excessive at that point, true😅
I'll look into making that a bit simpler :)

lua/luasnip/init.lua
L3MON4D3: Oh cool, will apply it👍👍

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: Aha, yeah your's is better👍
I was debating between `(Node[]|Node[][])` and didn't think one step further :D

DOC.md
```diff
@@ -3908,10 +3929,9 @@ These are the settings you can provide to `luasnip.setup()`:
     * `pos: (integer,integer)?` Position at which the snippet should be inserted. Pass as
       `(row,col)`, both 0-based, the `col` given in bytes.
     * `indent: boolean?` Whether to prepend the current lines' indent to all lines of the snippet.
-      `true` by default.  
-      Turning this off is a good idea when a LSP server already takes indents into consideration. In
-      such cases, LuaSnip should not add additional indents. If you are using `nvim-cmp`, this could
-      be used as follows:
+      `true` by default. Turning this off is a good idea when a LSP server already takes indents
+      into consideration. In such cases, LuaSnip should not add additional indents. If you are using
+      `nvim-cmp`, this could be used as follows:
```
bew: note: even if Markdown doesn't really care, I think we should keep / have newlines between sentences, it helps a lot with git diffing & improves raw-file readability as well ;)

I can do a PR someday if you want

lua/luasnip/init.lua
bew: my suggestion would be to just remove them entirely 👀

doc/luasnip.txt
bew: That would give nice table of content in the `DOC.md` file as well 😀

doc/luasnip.txt
```diff
@@ -756,42 +756,55 @@ ChoiceNodes allow choosing between multiple nodes.
...
+
+- `pos: integer` Jump-index of the node (See |luasnip-basics-jump-index|).
+- `choices: table<integer,(LuaSnip.Node|LuaSnip.Node[])>` A list of nodes that
+    can be switched between. If a list of nodes is passed as a choice, it will be
+    turned into a snippetNode. Jumpable nodes that generally need a jump-index
+    don’t need one when used as a choice, they inherit the choiceNode’s
+    jump-index. Additionally, one should make sure the cursor has a position to
+    stop at inside every choice, since changing the choice is generally only
+    possible when inside the choiceNode. This means that in `sn(nil,
+    {...nodes...})` `nodes` has to contain an `i(1)`, otherwise LuaSnip will just
```
bew: ⚠️ Well the generation is _much_ better now, but the newline broke the inline code block 🤔

L3MON4D3: Mhmmm, that's on panvimdoc, unfortunately :/

DOC.md
L3MON4D3: Tbh, not sure.. we can have forced newlines by inserting two spaces at the line-end in the annotations. But I think the unit of change can be a paragraph, not necessarily a sentence

bew:
I just wanted to make sure it wasn't forgotten 😅
No rush, I probably won't do much coding this week-end

lua/luasnip/init.lua
L3MON4D3: Yeah, that's it :)

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: Mhmm true.. maybe making the choices their own class is already a bit premature, they have like two extra fields. But I think it'd still be a good idea, though I don't like `ChoiceEntryNode`, maybe `ChoiceNode.Choice`? It's fitting, since the choices are a kind of subclass of (ie. private to) `Choice`

lua/luasnip/init.lua
L3MON4D3: 9611fc1

lua/luasnip/nodes/snippet.lua
L3MON4D3: Ahh, important! Fixed by 357800b

lua/luasnip/init.lua
L3MON4D3: e15ad06 :)

L3MON4D3: (totally agree, `{[` is a bit more easily ignored than `table<` :D)

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: 5857a31

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: ea7ba6d

L3MON4D3:
Okay, that were quite a few changes for me saying I'm done :D
I've decided to implement the api-doc as headings, since panvimdoc can [turn those into proper vimdoc-links](https://github.com/kdheepak/panvimdoc/issues/16#issuecomment-1447490485). One issue with that is that doing that properly requires an annotation like `{doc=luasnip-api-get_snippets}`, which shows up in markdown.
I've circumvented that by having `luals-mdgen` accept a `mode`-flag, which is passed through to the render-regions. So, `DOC.md` can be generated with the `doc=...`-annotations for panvimdoc, and without for the real `DOC.md`.
Finally, I've forked panvimdoc and made some small adjustments, codespans should no longer be split over multiple lines.

@bew, please take a look and let me know of any issues you find :)

doc/luasnip.txt
L3MON4D3: https://github.com/L3MON4D3/panvimdoc/commit/2bd11e62234e0c410e3b215d5f7762e11aac7804

doc/luasnip.txt
L3MON4D3: 11a250d9a0174bde74ecb040194c8c825af7677d :)

lua/luasnip/init.lua
```diff
@@ -66,19 +71,38 @@ local function default_snip_info(snip)
...
+---    return {
+---        name = snip.name,
+---        trigger = snip.trigger,
+---        description = snip.description,
+---        wordTrig = snip.wordTrig and true or false,
+---        regTrig = snip.regTrig and true or false,
+---    }
+---end
+---```
+---@return {[string]: T[]} available_info Table mapping filetypes to list of
```
bew: You didn't remove all 'return keyword' like `available_info` here, is that intentional?
(I had understood that you removed them 👀)

👉 I removed the remaining ones in #1354

lua/luasnip/nodes/choiceNode.lua
bew: I think that since it's a Node type, the choice item's type name should also have the `Node` suffix like `ChoiceNode`..
But `ChoiceNode.ChoiceNode` is kinda even more confusing 😬
What about `ChoiceNode.EntryNode` ? `ChoiceNode.ItemNode` ?

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -45,7 +49,52 @@ function ChoiceNode:init_nodes()
...
+---then using that to construct `choiceNode`s:
+---```lua
+---local function restore_cursor_choice(pos, choices, opts)
+---    if opts then
+---        opts.restore_cursor = true
+---    else
+---        opts = {restore_cursor = true}
+---    end
+---    return c(pos, choices, opts)
+---end
```
bew: NOTE: A simpler impl of that function:
```suggestion
---local function restore_cursor_choice(pos, choices, opts)
---    opts = opts or {}
---    opts.restore_cursor = true
---    return c(pos, choices, opts)
---end
```

👉 Now included in #1354

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -45,7 +49,52 @@ function ChoiceNode:init_nodes()
...
+---into a snippetNode.  
+---Jumpable nodes that generally need a jump-index don't need one when used as a
+---choice, they inherit the choiceNode's jump-index. Additionally, one should
+---make sure the cursor has a position to stop at inside every choice, since
+---changing the choice is generally only possible when inside the choiceNode.
+---This means that in `sn(nil, {...nodes...})` `nodes` has to contain an
+---`i(1)`, otherwise LuaSnip will just "jump through" the nodes, making it
+---impossible to change the choice after switching to it. Using an `insertNode`
+---or `textNode` directly as the choice is also fine, the latter are
+---special-cased to have a jump-point at the beginning of their text.
```
bew: Slight rephrase, I think it reads a bit better, and help separate the various inline code blocks in markdown/vimdoc
```suggestion
---This means that in `sn(nil, {...nodes...})` the given `nodes` have to contain an
---`insertNode` (e.g. `i(1)`), otherwise LuaSnip will just "jump through" the nodes
---making it impossible to change the current choice after switching to it.
---Using an `insertNode` or `textNode` directly as the choice is also fine, the latter
---is special-cased to have a jump-point at the beginning of its text.
```
👉 Now included in #1354

lua/luasnip/nodes/node.lua
```diff
@@ -6,8 +6,30 @@ local events = require("luasnip.util.events")
...
 local Node = {}
 
+---@alias LuaSnip.NodeExtOpts {["active"|"passive"|"visited"|"unvisited"|"snippet_passive"]: vim.api.keyset.set_extmark}
+
+---@class LuaSnip.Opts.Node
+---@field node_ext_opts LuaSnip.NodeExtOpts? Pass these opts through to the
+---underlying extmarks representing the node. Notably, this enables highlighting
+---the nodes, and allows the highlight to be different based on the state of the
+---node/snippet. See [ext_opts](../../../DOC.md#ext_opts)
+---@field merge_node_ext_opts boolean? Whether to use the parents' `ext_opts` to
```
bew, line 19: Something I use a lot in my configs is to mark fields/params as optional, and the following type now only describes the expected type **when it is given**.
I think this is usually a nicer way to describe class fields as the actual value being nil doesn't have any real meaning.

Here it would look like this:
```suggestion
---@field merge_node_ext_opts? boolean Whether to use the parents' `ext_opts` to
```
(the same kind of thing can be applied in many places)


👉 Made a PR for you ;) #1354

bew (commented):
Starting to look really good!

Few tweaks still, and a small PR for your branch: #1354

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: Yes, nice!

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: Right, yeah, not being able to distinguishing adjacent codespans in vimdoc was also something I noticed, this is a good solution.
And I agree, better phrasing :+1:

lua/luasnip/nodes/node.lua
L3MON4D3: Very good, as of c49fdb2 also visible in the doc :)

bew (approved):
Re-reading parts of this PR, I missed a few files during my last set of indent changes 👀 They can be done later though.
You'll have to regen de final DOC.md (I didn't play with that) I think before merging.

doc/luasnip.txt
```diff
@@ -1,4 +1,4 @@
-*luasnip.txt*             For NVIM v0.8.0            Last change: 2025 June 10
+*luasnip.txt*           For NeoVim 0.7-0.11          Last change: 2025 June 23
```
bew: The range is nice ✨

doc/luasnip.txt
```diff
@@ -756,43 +756,54 @@ ChoiceNodes allow choosing between multiple nodes.
...
+                opts.restore_cursor = true
                 return c(pos, choices, opts)
             end
         <
-
-Jumpable nodes that normally expect an index as their first parameter don’t
-need one inside a `choiceNode`; their jump-index is the same as the
-choiceNodes’.
-
-As it is only possible (for now) to change choices from within the
```
bew, line 789: This whole section that was _readable_ before is now cramed into the `@param choices`'s documentation which doesn't render very well here 👀

Since the choices are the main goal of the choice node, I think we should move the global explanation / quirks details in the top-documentation of the function, with multiple paragraphs to keep it easily readable 🤔

WDYT?

doc/luasnip.txt
```diff
@@ -3628,197 +3673,423 @@ GENERAL ~
...
-        values, it’s safe to remove the text manually.
-    - `expand_params`: table, override `trigger`, `captures` or environment of the
-        snippet. This is useful for manually expanding snippets where the trigger
-        passed via `trig` is not the text triggering the snippet, or those which expect
-        `captures` (basically, snippets with a non-plaintext `trigEngine`).
-        One example:
+
+get_active_snip(): LuaSnip.Snippet?              *luasnip-api-get_active_snip*
+
+Get the currently active snippet. This function returns:
```
bew: note: `This function returns` should be forced to be on its own line.

The Markdown rendering looks like:
image
Panvimdoc just skips the ending 2 spaces that markdown renderers handle as 'put-newline-here'.

👉 Could you force a full blank line there? (and remove ending spaces)

I think it'll just improve everything and actually be more consistent in github, because a text then a list adds a _medium_ vertical spacing before the list, that is completely absent when using newline via spaces at the end.. So the line of text introducing the list is mashed into the previous paragraph 😬

<details>
<summary>Compare this (click to open)</summary>

Compare this:

Get the currently active snippet.
And some arbitrary more details about the function here....
This function returns:

- `some thing here`
- other list item

to this:

Get the currently active snippet.
And some arbitrary more details about the function here....

This function returns:

- `some thing here`
- other list item

</details>

We could also **bold** or _italicize_ the `This function returns` to make it _stand out_ more when there are long function description (that might be in paragraphs if we develop this further!).

---

(note: When a function has params it's already good though)

doc/luasnip.txt
```diff
@@ -3628,197 +3673,423 @@ GENERAL ~
...
+- `dir: (1|-1)` Test jumping forwards/backwards.
+
+
+snip_expand(snippet, opts?): LuaSnip.ExpandedSnippet *luasnip-api-snip_expand*
+
+Expand a snippet in the current buffer.
+
+- `snippet: LuaSnip.Snippet` The snippet.
+- `opts?: LuaSnip.Opts.SnipExpand?` Optional additional arguments. Valid keys
+    are:
```
bew: I think we should have the same forced-on-new-line for `Valid keys are:`

Markdown looks like:
image

doc/luasnip.txt
```diff
@@ -3628,197 +3673,423 @@ GENERAL ~
...
+
+- `LuaSnip.ExpandedSnippet` The snippet that was inserted into the buffer.
+
+
+expand(opts?): boolean                                    *luasnip-api-expand*
+
+Find a snippet whose trigger matches the text before the cursor and expand it.
+
+- `opts?: LuaSnip.Opts.Expand?` Subset of opts accepted by `snip_expand`. Valid
+    keys are:
```
bew: another example (for testing ;) )

doc/luasnip.txt
```diff
@@ -3628,197 +3673,423 @@ GENERAL ~
...
+because the text of a snippet was deleted), the snippet is removed from the
+jumplist and the current node set to the end/beginning of the next/previous
+snippet.
+
+- `node: LuaSnip.Node`
+
+
+filetype_extend(ft, extend_ft)                   *luasnip-api-filetype_extend*
+
+Add `extend_ft` filetype to inherit its snippets from `ft`. Example:
```
bew: 🤔 This `Example` should be on its own line as well.

This one is surprising though, because I know I left a blank line before it in the Lua doc annotation:
image

But the markdown is all mashed together (will be an issue as we add more top documentation to functions & stuff!):
image

👉 In general, I think that blank lines in Lua docs should stay as full blank lines in the md generation.

... And what's going on with the ``  `sh`/`zsh`/`bash` `zsh`/`bash` `bash` buffers. `` at the end 👀
Did your md generator invent some stuff here? (is it sentient??? 😱)

doc/luasnip.txt
```diff
@@ -3628,197 +3673,423 @@ GENERAL ~
...
+
+
+add_snippets(ft?, snippets, opts?)                  *luasnip-api-add_snippets*
+
+Add snippets to luasnip’s snippet-collection. NOTE: Calls `refresh_notify` as
+needed if enabled via `opts.refresh_notify`.
+
+- `ft?: string?` The filetype to add the snippets to, or nil if the filetype is specified in
+    `snippets`.
+- `snippets: (LuaSnip.Addable[]|{ [string]: LuaSnip.Addable[] })` If `ft` is nil a table mapping a
```
bew: note: I think we're missing some docs about what `LuaSnip.Addable` means, can be 'added' (👀) later though.

doc/luasnip.txt
```diff
@@ -3628,197 +3673,423 @@ GENERAL ~
...
+clean_invalidated(opts?)                       *luasnip-api-clean_invalidated*
+
+Clean invalidated snippets from internal snippet storage. Invalidated snippets
+are still stored; it might be useful to actually remove them as they still have
+to be iterated during expansion.
+
+- `opts?: LuaSnip.Opts.CleanInvalidated?` Additional, optional arguments.
+
+
+activate_node(opts?)                               *luasnip-api-activate_node*
```
bew: the opts type for `clean_invalidated` is not expanded 🤔

bew (commented):
> I'll merge after you give it one last look @bew :)

Be careful what you wish for, I love reviewing PRs and finding things to improve, so it can take a while 😅

Let's say this is my last big review though, it's _almost_ ready for me!

> > Re-reading parts of this PR, I missed a few files during my last set of indent changes 👀 They can be done later though.
> 
> Oh, which ones? I thought `init.lua` and `choiceNode.lua` were the only really relevant files? But yeah, let's do that later 👍

I was looking at the other files of that PR that change some type annotations, like `_extra_types.lua`, `node.lua`, `snippet.lua`, `util.lua`, ...

I didn't read the CI stuff in details, but you're going to 'live with it' so it gotta be good :D 

The vimhelp rendering is generally much much better than before (and inline code blocks seems all perfect now! No cut-off in the middle with a newline) ✨
Params description is still not great but for everything else it's a big positive!

I did a full pass on the generated vimhelp file, and noticed a few issues that surfaces small issues with the md doc generation.

doc/luasnip.txt
L3MON4D3: TY! :)

doc/luasnip.txt
L3MON4D3: Yeah I see what you mean, some of that info in the `choices`-description are more like general behaviour..
I'll move that stuff into the general description :+1:

doc/luasnip.txt
L3MON4D3: Good suggestion, this is more consistent :+1:
Maybe it'd be better to have some kind of keyword, like **Return-values**, as opposed to the current simple text. But I'd say that's something for later :D

doc/luasnip.txt
L3MON4D3: 3ccbb4f

doc/luasnip.txt
L3MON4D3: a70f977

doc/luasnip.txt
L3MON4D3: Yeah, this was a bigger issue, panvimdoc collapsed `  \n` into a single whitespace, so I had to dig into that a bit :sweat_smile:
But it should work now!

L3MON4D3: 5be06e2a47f931dd5ec9978d1b6a9e5f98323f39

doc/luasnip.txt
L3MON4D3: 5be06e2a47f931dd5ec9978d1b6a9e5f98323f39

doc/luasnip.txt
L3MON4D3: Again, this needed a few larger changes, 68faade forces an empty line between two paragraphs, 2354788 replaces the very naive previous algorithm for keeping something like `` (`sh`) `` together with something that is more robust.
So, it's not about to become Skynet, I just appended all `[^%s]`-chars to code-spans, which includes other code-spans xD

L3MON4D3: Oh, and totally agree, full empty lines are paragraph-breaks, they should be preserved :+1:

bew: Thanks for saving the world this time 🙏😄

Yay for newlines 🚀

L3MON4D3: Didn't even have to time travel :muscle:

doc/luasnip.txt
L3MON4D3: Yeah yeah, see what you did there :D
I'll add a small explanation for now (in short, `LuaSnip.Addable` is a container for one or more expandable snippets, it was introduced to support `multiSnippet` where adding one object could require adding more than one snippet)

doc/luasnip.txt
L3MON4D3: Good catch!
fbbd855

doc/luasnip.txt
L3MON4D3: d8cee89

data/DOC-template.md
```diff
@@ -3784,6 +3784,9 @@ local opts_table = {
     activate_node = {type_expand = {
         ["LuaSnip.Opts.ActivateNode?"] = {explain_type = "LuaSnip.Opts.ActivateNode"},
     }},
+    clean_invalidated = {type_expand = {
+        ["LuaSnip.Opts.CleanInvalidated?"] = {explain_type = "LuaSnip.Opts.CleanInvalidated"},
```
bew, line 3788: huh, I didn't read that very carefully, I hadn't realized those where manually listed here

It's a good first step I guess but I feel like it gives a bit too much flexibility that isn't _really_ needed here 🤔
Oh well, it'll get refactored later once it's annoying enough `¯\_(ツ)_/¯` (story of my life :D)

lua/luasnip/nodes/snippet.lua
```diff
@@ -66,8 +66,22 @@ local stored_mt = {
...
+---@alias LuaSnip.SnippetID integer
+
+---Anything that can be passed to ls.add_snippets().
+---@class LuaSnip.Addable
+
+---Represents an expanded snippet.
+---@class LuaSnip.ExpandedSnippet
+
+---@class LuaSnip.Snippet: LuaSnip.Addable, LuaSnip.ExpandedSnippet
+---@class LuaSnip.SnippetNode: LuaSnip.Node
```
bew, line 82: ⚠️ `LuaSnip.Snippet` is already defined above at line 69 😬, I think that other definition is wrong though and should be called `LuaSnip.SnippetNode` like it is here.
I think you just have to replace that earlier definition with the one here.

bew (approved):
Anything else on your radar ?

Appart from this comment about conflicting `@class` definition it's good to go for me! 🎉 🚀 🚢

data/DOC-template.md
L3MON4D3: Ah, yeah I've gone the verbose-but-precise way for now :D
Once we have enough usecases we can simplify that, for sure :+1:

lua/luasnip/nodes/snippet.lua
L3MON4D3: Ah! Yes, good catch!

L3MON4D3: bd39313

L3MON4D3:
Aaaalright, that's it for now :D
Thank you for spending time on these thorough reviews, they really helped! :)
Definitely feel free to tinker with this too, just let me know if you're tackling something so we don't start on the same thing accidentally :D

PR #1354, opened 2025-06-23
Few annotation tweaks for PR #1353

bew:
- Update some types to mark 'optional' at field/param-level
- Cleanup remaining keywords for @return annotations
- Add few top-level documentation
- Separate field/param with newlines for better readability
- Include other suggested changes from my comments on #1353

### DIFF
--- a/lua/luasnip/init.lua
+++ b/lua/luasnip/init.lua
@@ -20,7 +20,7 @@ local log = require("luasnip.util.log").new("main")
 ---@class LuaSnip.API
 local API = {}
 
----Get the currently active snippet.
+--- Get the currently active snippet.
 ---@return LuaSnip.Snippet? _ The active snippet if one exists, or `nil`.
 function API.get_active_snip()
 	local node = session.current_nodes[vim.api.nvim_get_current_buf()]
@@ -46,19 +46,21 @@ end
 
 ---@class LuaSnip.Opts.GetSnippets
 ---@field type? "snippets"|"autosnippets" Whether to get snippets or
----autosnippets. Defaults to "snippets".
+---  autosnippets. Defaults to "snippets".
 
----Retrieve snippets from luasnip.
+--- Retrieve snippets from luasnip.
 ---@param ft? string Filetype, if not given returns snippets for all filetypes.
 ---@param opts? LuaSnip.Opts.GetSnippets Optional arguments.
----@return LuaSnip.Snippet[]|{[string]: LuaSnip.Snippet[]} _ Flat array when `ft`
----is non-nil, otherwise a table mapping filetypes to snippets.
+---@return LuaSnip.Snippet[]|{[string]: LuaSnip.Snippet[]} _ Flat array when
+---  `ft` is non-nil, otherwise a table mapping filetypes to snippets.
 function API.get_snippets(ft, opts)
 	opts = opts or {}
 
 	return snippet_collection.get_snippets(ft, opts.type or "snippets") or {}
 end
 
+---@param snip LuaSnip.Snippet
+---@return any
 local function default_snip_info(snip)
 	return {
 		name = snip.name,
@@ -69,25 +71,27 @@ local function default_snip_info(snip)
 	}
 end
 
----Retrieve information about snippets available in the current file/at the
----current position (in case treesitter-based filetypes are enabled).
+--- Retrieve information about snippets available in the current file/at the
+--- current position (in case treesitter-based filetypes are enabled).
+---
 ---@generic T
 ---@param snip_info? (fun(LuaSnip.Snippet): T) Optionally pass a function that,
----given a snippet, returns the data that is returned by this function in the
----snippets' stead. By default, this function is
----```lua
----function(snip)
----    return {
----        name = snip.name,
----        trigger = snip.trigger,
----        description = snip.description,
----        wordTrig = snip.wordTrig and true or false,
----        regTrig = snip.regTrig and true or false,
----    }
----end
----```
+---  given a snippet, returns the data that is returned by this function in the
+---  snippets' stead.  
+---  By default, this function is
+---  ```lua
+---  function(snip)
+---      return {
+---          name = snip.name,
+---          trigger = snip.trigger,
+---          description = snip.description,
+---          wordTrig = snip.wordTrig and true or false,
+---          regTrig = snip.regTrig and true or false,
+---      }
+---  end
+---  ```
 ---@return {[string]: T[]} _ Table mapping filetypes to list of data returned by
----snip_info function.
+---  snip_info function.
 function API.available(snip_info)
 	snip_info = snip_info or default_snip_info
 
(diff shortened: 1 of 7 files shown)


L3MON4D3:
Ah, good idea :+1:

lua/luasnip/init.lua
```diff
@@ -50,11 +50,10 @@ end
...
 
 ---Retrieve snippets from luasnip.
----@param ft string? Filetype, if not given returns snippets for all filetypes.
----@param opts LuaSnip.Opts.GetSnippets? Optional arguments.
----@return LuaSnip.Snippet[]|{[string]: LuaSnip.Snippet[]} snippets Flat
----array when `ft` is non-nil, otherwise a table mapping filetypes to
----snippets.
+---@param ft? string Filetype, if not given returns snippets for all filetypes.
+---@param opts? LuaSnip.Opts.GetSnippets Optional arguments.
+---@return LuaSnip.Snippet[]|{[string]: LuaSnip.Snippet[]} Flat array when `ft`
```
L3MON4D3: Unfortunately, if there is a comment for the return-param, it also has to have a name:

screen

vs

screen

L3MON4D3: I think we can tolerate that small repetition :)

lua/luasnip/init.lua
```diff
@@ -74,7 +73,7 @@ end
 ---Retrieve information about snippets available in the current file/at the
 ---current position (in case treesitter-based filetypes are enabled).
 ---@generic T
----@param snip_info (fun(LuaSnip.Snippet): T)? Optionally pass a function that,
+---@param snip_info? (fun(LuaSnip.Snippet): T) Optionally pass a function that,
```
L3MON4D3: Oh that's pretty cool, makes it easier to see which parameters are optional!
It seems like the `?` is still appended to the type (like before), so if we want to have a description like `available(snip_info?): {[string]: T}`, we'd have to do a tiny tweak to `luals-mdgen`. Seems very feasible :)

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -81,18 +78,18 @@ end
 ---
 ---@param choices (LuaSnip.Node|LuaSnip.Node[])[] A list of nodes that can be
 ---switched between. If a list of nodes is passed as a choice, it will be turned
----into a snippetNode.  
+---into a snippetNode.
```
L3MON4D3: Oh, this paragraph-break was intentional, it'll be preserved in the generated `DOC.md`, and causes a linebreak when viewed on github

lua/luasnip/session/snippet_collection/init.lua
```diff
@@ -225,6 +225,8 @@ end
 local current_id = 0
 -- snippets like {ft1={<snippets>}, ft2={<snippets>}}, opts should be properly
 -- initialized with default values.
+---@param snippets {[string]: LuaSnip.Addable[]}
+---@param opts LuaSnip.Opts.AddSnippets
```
L3MON4D3: These are a tiny bit narrower than the opts of the exposed `add_snippets`: `type` may no longer be `nil`.
I don't have a great solution for this though, we could have `LuaSnip.Opts.AddSnippets` inherit from `LuaSnip.Opts.SessionAddSnippets` and override the `type`, but that'd mean we also duplicate its documentation.
OTOH, writing this out, this does not seem like a big concession, probably fine :)

bew: I'll make another type then

lua/luasnip/init.lua
bew: Odd 👀
What do you think about using `_` as the keyword? Then it's easy to ignore it in doc generation

lua/luasnip/init.lua
bew: sounds great 👍

lua/luasnip/init.lua
L3MON4D3: Yeah, that's fine too :+1:

bew: Note that is practice, it's pretty uncommon to add comment for the `@return` line, since the rest of the function's documentation usually already explain what it does / what it returns `¯\_(ツ)_/¯`

lua/luasnip/nodes/choiceNode.lua
bew: 🤬 Markdown... 🤬
I'd argue that to have better flow for the descriptions, we could have full blank lines to better split lines, but that's for another time I guess..

bew:
There is something I do in all my config as well, is to use this format for docstrings:
```lua
--- This is the doc, with a space before it
---@param any But all `@something` don't have the space after `---`
```
For `@param` long line continuation I indent them with 2 (or 4) spaces like:
https://github.com/bew/dotfiles/blob/acbc80844c/nvim/lua/mylib/keymap_system.lua#L15
Sometimes I also just put the entire docstring on the next line:
https://github.com/bew/dotfiles/blob/acbc80844c/nvim/lua/mycfg/hl_patterns/init.lua#L43
(but LuaLS hover has very bad support for multiline in general, I might switch to the rust version that you're using now 🤔)

At the moment, most of your docstrings use `---This is the doc, without space` format.
👉 Are you interested in using this kind of format everywhere? I can make these changes if you want 🤔

lua/luasnip/session/snippet_collection/init.lua
bew: NOTE: I had to create a new table to call the session add snippet function, as opts would have had conflicting type

L3MON4D3:
So far I haven't given much though to indenting this stuff (except deleting the one space Neogen inserts before descriptions :D)
I think that indenting with two spaces is a good idea, that gives us (incidentally) good alignment in the hover-window, and makes sense in the code (2 space indent is common enough).

screen

I'd be glad if you made those changes, I'm happy with one space for the initial description, and then two spaces for multi-line param/field descriptions :)

And yeah, incomplete handling of multiline params is why I switched `luals-mdgen` to the rust version, it seems to handle that better. I'm still on good ol' `lua-language-server` for lsp though, there it still works very well, even if it is not blazingly fast :P

lua/luasnip/init.lua
L3MON4D3: That's true, there are a few cases where I like it, eg. `expand` mainly expands the snippet, but also returns whether something was expanded. In that case the return value is more detached than in a function like `get_snippets`, and putting the description into the return-annotation feels appropriate

lua/luasnip/init.lua
L3MON4D3: https://github.com/L3MON4D3/luals-mdgen/commit/896453c24b1d2170e546be282d61fd9edf2a5de6

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: xD
A full blank line sounds like it'd be easier to read, I agree, but also a tad harder to parse than hard line breaks `  `, since these are just nodes in treesitter.. another time :P

lua/luasnip/session/snippet_collection/init.lua
L3MON4D3: That's fine :+1:
Maybe we'll have to overthing dumping all the optional-param classes into `LuaSnip.Opts`, but so far it seems fine :)

bew:
Done, not that I only updated doc annotations in files that are relevant for this PR, more indent tweaks will need to be applied in all other files of the codebase as we go through them to add doc annotations.

PR #1137, opened 2024-03-04
Allow self-dependent dynamicNode

L3MON4D3:
Deal with/Explicitly allow dynamicNodes being updated due to a change inside of them.
This is to enable stuff like #1096 and #1102, where the text of an insertNode should be updated while typing.

One important limitation this has (for now??) is that the updates have to reach some equilibrium (text is unchanged after a second update), otherwise we just keep updating.

One example:
```lua
ls.setup_snip_env()

local opt_arg = require("luasnip.nodes.optional_arg").new_opt

ls.add_snippets("all", {
	s("test", {
		t":",
		d(1, function(args)
			if args[1] == nil then
				return sn(nil, { i(1, "fff", {key = "asd"}) })
			else
				for i, str in ipairs(args[1]) do
					args[1][i] = str:lower()
				end
				return sn(nil, { i(1, args[1], {key = "asd"}) })
			end
		end, { opt_arg(k("asd")) }),
		t":"
	})
}, {key = "qwe"})
```

Note the other new addition, `opt_arg` (optional argnodes, we currently don't run the dynamicNode-function if an argnode is missing), for allowing the initial creation of the `sn` when the `"asd"`-node does not yet exist.

This is missing tests and doc for now, and I'm not certain it handles all cases correctly, but I'd like to show this since I think it should work for simple ones.

PR #1396, opened 2025-10-31
WIP: Add MANY type annotations 🚀

bew:
Hola @L3MON4D3, it's been a while o/

I've recently updated my neovim to 0.11 & reworking my config a bit, and since the _big_ PR #1353 that started adding type annotations & DOC.md generation, I didn't touch my snippets at all...
The nvim upgrade was the nudge, and [I started replacing the few type annotations I had manually crafted in my config](https://github.com/bew/dotfiles/commit/78d7f38ed20b696d606a2df7e9b9d356f28f9249), replacing them with the ones from the plugin!..
...
And then I kinda went down the rabbit hole 🤪
Porting documentation, reading the codebase while adding a few types for the things I understood, and adding MOAR types....

:point_right: So here is a first draft of ~1k LoC of doc additions & a few small reworks 🚀🚀

It took a few iterations of codebase exploration to kinda understand the various relationships between the different kind of snippet, node, and what 'layer'/'stage' of the snippets definition process I was on

I'm still far from understanding everything, and I might have made some mistakes, put some fields in the wrong class or failed a few inheritance relations, but I think it's pretty good for a start!

There are still many things I saw and could work on / refactor with your permissions, but I got a
good chunk going on and I'm quite happy with it!

> [!NOTE]
> I left a few `FIXME`/`IDEA` in the PR with questions for you

> [!WARNING]
> I didn't test the DOC.md generation at all, I only focused on the Lua-side of things

### Some things I typed/documented

- Node, InsertNode, FunctionNode, DynamicNode, ChoiceNode, with (most of) their internal fields ✨
- Snippet, SnippetNode & most of the internal snippet-initialization functions (their input opts & their normalized types)

  I managed to build a sort of hierarchy of classes that when combined build the final Snippet & SnippetNode types (with extra fields if needed), I'd need careful review of this part, I tried to follow the calls and build the structures but I might be wrong here and there..

  I know there are a few fields missing but they're kind of overwelming by their number, and the
  lack of internal documentation in such a big codebase 😅

- MultiSnippet, SnippetString
- some util functions here and there
- fmt's main functions
- AbsoluteIndexer, KeyIndexer, NodeRef

### Other things

- moved `copy3` impl to util, to avoid duplication between `snippet.lua` & `fmt.lua`
- rewrote the `events.lua` & `types.lua` for nicer typing without duplication

---

👉 In the end, most core files type-checks correctly, with only a few warnings that make sense and
that would need to be addressed (some of them challenge the class hierarchy in place, and I made a
few comments on ideas to improve the situation)

As for where to go with the types design, I think the codebase would benefit from splitting the Snippet class into multiple classes to reflect the various stages of use (barebone for internal use, snippetnode, snippet, expandedsnippet, (..?)). For the actual type annotations, maybe you'd like to have a public/private interface to avoid exposing the huge internals of the nodes to the users 🤔

I was also quite surprised that you used `Node:new` both to create new node types & new node instances, would you welcome some changes around this to segregate the types/functions more by use-cases?
(this can probably come later though)

> [!NOTE]
> This PR targets the `self-dependent-dNode` branch, since this is the branch I'm on normally.
> :grey_question: I'm wondering, do you think we could merge it soon or do you still have planned work on it?
> I'm asking because otherwise I'd need to rework some parts of this PR to be able to rebase on `master` 👀

Let me know what you think, and how you'd want to tackle this!
I'm available to do a day/night pair-programming/review session if you want :upside_down_face:

### DIFF
--- a/lua/luasnip/extras/init.lua
+++ b/lua/luasnip/extras/init.lua
@@ -122,7 +122,7 @@ return {
 	--- s("extras4", { i(1), t { "", "" }, extras.rep(1) })
 	--- ```
 	---
-	---@param node_ref LuaSnip.NodeRef a single [Node Reference](#node-reference).
+	---@param node_ref LuaSnip.NodeRef a single [Node Reference](../../../DOC.md#node-reference).
 	---@return LuaSnip.FunctionNode
 	rep = function(node_ref)
 		return F(function(args)
--- a/lua/luasnip/extras/init.lua
+++ b/lua/luasnip/extras/init.lua
@@ -116,11 +116,18 @@ end
 return {
 	lambda = lambda,
 	match = match,
-	-- repeat a node.
-	rep = function(node_indx)
+	--- Repeat a node, by inserting the text of the passed node.
+	---
+	--- ```lua
+	--- s("extras4", { i(1), t { "", "" }, extras.rep(1) })
+	--- ```
+	---
+	---@param node_ref LuaSnip.NodeRef a single [Node Reference](#node-reference).
+	---@return LuaSnip.FunctionNode
+	rep = function(node_ref)
 		return F(function(args)
 			return args[1]
-		end, node_indx)
+		end, node_ref)
 	end,
 	-- Insert the output of a function.
 	partial = function(func, ...)
--- a/lua/luasnip/nodes/choiceNode.lua
+++ b/lua/luasnip/nodes/choiceNode.lua
@@ -122,7 +122,12 @@ function ChoiceNode:subsnip_init()
 	for _, choice in ipairs(self.choices) do
 		choice.parent = self.parent
 		-- only insertNode needs this. (?)
-		if vim.list_contains({types.textNode, types.insertNode, types.functionNode}, choice.type) then
+		if
+			vim.list_contains(
+				{ types.textNode, types.insertNode, types.functionNode },
+				choice.type
+			)
+		then
 			choice.pos = self.pos
 		end
 	end
--- a/lua/luasnip/nodes/choiceNode.lua
+++ b/lua/luasnip/nodes/choiceNode.lua
@@ -123,7 +123,7 @@ function ChoiceNode:subsnip_init()
 		choice.parent = self.parent
 		-- only insertNode needs this. (?)
 		if
-			util.list_contains(
+			vim.tbl_contains(
 				{ types.textNode, types.insertNode, types.functionNode },
 				choice.type
 			)
--- a/lua/luasnip/nodes/choiceNode.lua
+++ b/lua/luasnip/nodes/choiceNode.lua
@@ -61,7 +61,7 @@ end
 ---  ```
 ---  Consider passing this override into `snip_env`.
 ---
----@field node_callbacks? {["change_choice"|"enter"|"leave"]: fun(node:LuaSnip.Node)}
+---@field node_callbacks? {[LuaSnip.EventType]: fun(node:LuaSnip.Node)}
 ---  Specify functions to call after changing the choice, or entering or leaving
 ---  the node. The callback receives the `node` the callback was called on.
 
--- a/lua/luasnip/nodes/choiceNode.lua
+++ b/lua/luasnip/nodes/choiceNode.lua
@@ -123,7 +123,7 @@ function ChoiceNode:subsnip_init()
 		choice.parent = self.parent
 		-- only insertNode needs this. (?)
 		if
-			vim.list_contains(
+			util.list_contains(
 				{ types.textNode, types.insertNode, types.functionNode },
 				choice.type
 			)
--- a/lua/luasnip/nodes/choiceNode.lua
+++ b/lua/luasnip/nodes/choiceNode.lua
@@ -458,6 +458,7 @@ function ChoiceNode:extmarks_valid()
 	return node_util.generic_extmarks_valid(self, self.active_choice)
 end
 
+---@param opts LuaSnip.Opts.NodeSubtreeDo
 function ChoiceNode:subtree_do(opts)
 	opts.pre(self)
 	self.active_choice:subtree_do(opts)
--- a/lua/luasnip/nodes/choiceNode.lua
+++ b/lua/luasnip/nodes/choiceNode.lua
@@ -11,17 +11,17 @@ local feedkeys = require("luasnip.util.feedkeys")
 local log = require("luasnip.util.log").new("choice")
 
 ---@class LuaSnip.ChoiceNode.ItemNode: LuaSnip.Node
----@field choice ...
----@field parent? LuaSnip.ChoiceNode
----@field next_choice? LuaSnip.ChoiceNode.ItemNode
+---@field choice LuaSnip.ChoiceNode
+---@field next_choice LuaSnip.ChoiceNode.ItemNode
 ---@field prev_choice LuaSnip.ChoiceNode.ItemNode
 
 ---@class LuaSnip.ChoiceNode: LuaSnip.Node
 ---@field choices LuaSnip.ChoiceNode.ItemNode[]
----@field parent? LuaSnip.ChoiceNode
 ---@field active_choice LuaSnip.ChoiceNode.ItemNode The active choice item node
 ---@field restore_cursor boolean Whether the cursor should be restored when
 ---  changing the current choice.
+---@field absolute_position integer[] (note: init when a choice is set)
+---@field absolute_insert_position integer[] (note: init when a choice is set)
 local ChoiceNode = Node:new()
 
 function ChoiceNode:init_nodes()
@@ -74,7 +74,8 @@ end
 --- `nodes` have to contain an `insertNode` (e.g. `i(1)`). Using an `insertNode`
 --- or `textNode` directly as a choice is also fine, the latter is special-cased
 --- to have a jump-point at the beginning of its text.
----@param pos integer Jump-index of the node.
+---
+---@param pos integer? Jump-index of the node.
 ---  (See [Basics-Jump-Index](../../../DOC.md#jump-index))
 ---
 ---@param choices (LuaSnip.Node|LuaSnip.Node[])[] A list of nodes that can be
@@ -83,13 +84,13 @@ end
 ---  Jumpable nodes that generally need a jump-index don't need one when used as
 ---  a choice since they inherit the choiceNode's jump-index anyway.
 ---
----@param opts? LuaSnip.Opts.ChoiceNode Additional optional arguments.
+---@param node_opts? LuaSnip.Opts.ChoiceNode Additional optional arguments.
 ---@return LuaSnip.ChoiceNode
-function ChoiceNode.C(pos, choices, opts)
-	opts = opts or {}
-	if opts.restore_cursor == nil then
+function ChoiceNode.C(pos, choices, node_opts)
+	node_opts = node_opts or {}
+	if node_opts.restore_cursor == nil then
 		-- disable by default, can affect performance.
-		opts.restore_cursor = false
+		node_opts.restore_cursor = false
 	end
 
 	-- allow passing table of nodes in choices, will be turned into a
@@ -109,8 +110,9 @@ function ChoiceNode.C(pos, choices, opts)
 		mark = nil,
 		dependents = {},
 		-- default to true.
-		restore_cursor = opts.restore_cursor,
-	}, opts)
+		restore_cursor = node_opts.restore_cursor,
+	}, node_opts)
+	---@cast c LuaSnip.ChoiceNode
 	c:init_nodes()
 	return c
 end
@@ -147,6 +149,8 @@ function ChoiceNode:make_args_absolute()
 
 	for _, choice in ipairs(self.choices) do
 		-- relative to choiceNode!!
+		---(allowed: this arg only exists for some node types)
+		---@diagnostic disable-next-line: redundant-parameter
 		choice:make_args_absolute(self.absolute_insert_position)
 	end
 
@@ -266,6 +270,7 @@ end
 
 function ChoiceNode:setup_choice_jumps() end
 
+---@return LuaSnip.Node?
 function ChoiceNode:find_node(predicate, opts)
 	if self.active_choice then
 		if predicate(self.active_choice) then
@@ -362,7 +367,12 @@ function ChoiceNode:copy()
 	local o = vim.deepcopy(self)
 	for i, node in ipairs(self.choices) do
 		if node.type == types.snippetNode or node.type == types.choiceNode then
-			o.choices[i] = node:copy()
+			---@diagnostic disable-next-line: cast-type-mismatch (not smart enough..)
+			---@cast node LuaSnip.SnippetNode|LuaSnip.ChoiceNode
+			local nodecopy = node:copy()
+			---@diagnostic disable-next-line: cast-type-mismatch (not smart enough..)
+			---@cast nodecopy LuaSnip.ChoiceNode.ItemNode
+			o.choices[i] = nodecopy
 		else
 			setmetatable(o.choices[i], getmetatable(node))
 		end
--- a/lua/luasnip/nodes/choiceNode.lua
+++ b/lua/luasnip/nodes/choiceNode.lua
@@ -11,9 +11,17 @@ local feedkeys = require("luasnip.util.feedkeys")
 local log = require("luasnip.util.log").new("choice")
 
 ---@class LuaSnip.ChoiceNode.ItemNode: LuaSnip.Node
+---@field choice ...
+---@field parent? LuaSnip.ChoiceNode
+---@field next_choice? LuaSnip.ChoiceNode.ItemNode
+---@field prev_choice LuaSnip.ChoiceNode.ItemNode
 
 ---@class LuaSnip.ChoiceNode: LuaSnip.Node
 ---@field choices LuaSnip.ChoiceNode.ItemNode[]
+---@field parent? LuaSnip.ChoiceNode
+---@field active_choice LuaSnip.ChoiceNode.ItemNode The active choice item node
+---@field restore_cursor boolean Whether the cursor should be restored when
+---  changing the current choice.
 local ChoiceNode = Node:new()
 
 function ChoiceNode:init_nodes()
@@ -111,8 +119,8 @@ extend_decorator.register(ChoiceNode.C, { arg_indx = 3 })
 function ChoiceNode:subsnip_init()
 	for _, choice in ipairs(self.choices) do
 		choice.parent = self.parent
-		-- only insertNode needs this.
-		if choice.type == 2 or choice.type == 1 or choice.type == 3 then
+		-- only insertNode needs this. (?)
+		if vim.list_contains({types.textNode, types.insertNode, types.functionNode}, choice.type) then
 			choice.pos = self.pos
 		end
 	end
--- a/lua/luasnip/nodes/dynamicNode.lua
+++ b/lua/luasnip/nodes/dynamicNode.lua
@@ -607,6 +607,7 @@ function DynamicNode:extmarks_valid()
 	return true
 end
 
+---@param opts LuaSnip.Opts.NodeSubtreeDo
 function DynamicNode:subtree_do(opts)
 	opts.pre(self)
 	if opts.static then
--- a/lua/luasnip/nodes/dynamicNode.lua
+++ b/lua/luasnip/nodes/dynamicNode.lua
@@ -15,7 +15,7 @@ local session = require("luasnip.session")
 ---@field old_state? any
 ---@field dynamicNode? LuaSnip.DynamicNode
 
----@alias LuaSnip.DynamicNode.Fn fun(args: (string[])[], parent: LuaSnip.Snippet | LuaSnip.SnippetNode, old_state?: table, ...: table): LuaSnip.SnippetNode
+---@alias LuaSnip.DynamicNode.Fn fun(args: (string[])[], parent: LuaSnip.Snippet | LuaSnip.SnippetNode, old_state?: table, ...: any): LuaSnip.SnippetNode
 
 ---@class LuaSnip.DynamicNode: LuaSnip.Node
 ---@field fn LuaSnip.DynamicNode.Fn
@@ -217,6 +217,8 @@ function DynamicNode:jump_into_snippet(no_move)
 	return self:jump_into(1, no_move, false)
 end
 
+-- FIXME(@bew): This should be on a `ExpandedDynamicNode` class?
+-- To have access to `next`/`prev` (that would go on a `ExpandedNode` 🤔)
 function DynamicNode:update()
 	local args = self:get_args()
 	local str_args = node_util.str_args(args)
--- a/lua/luasnip/nodes/dynamicNode.lua
+++ b/lua/luasnip/nodes/dynamicNode.lua
@@ -263,8 +263,12 @@ function DynamicNode:update()
 
 	-- build new snippet before exiting, markers may be needed for
 	-- construncting.
-	tmp =
-		self.fn(effective_args or {}, self.parent, old_state, unpack(self.user_args))
+	tmp = self.fn(
+		effective_args or {},
+		self.parent,
+		old_state,
+		unpack(self.user_args)
+	)
 	---@cast tmp LuaSnip.SnippetNodeForDynNode
 
 	if self.snip then
--- a/lua/luasnip/nodes/dynamicNode.lua
+++ b/lua/luasnip/nodes/dynamicNode.lua
@@ -1,4 +1,3 @@
-local DynamicNode = require("luasnip.nodes.node").Node:new()
 local util = require("luasnip.util.util")
 local node_util = require("luasnip.nodes.util")
 local Node = require("luasnip.nodes.node").Node
@@ -12,20 +11,90 @@ local log = require("luasnip.util.log").new("dynamicNode")
 local describe = require("luasnip.util.log").describe
 local session = require("luasnip.session")
 
-local function D(pos, fn, args, opts)
-	opts = opts or {}
+---@class LuaSnip.SnippetNodeForDynNode: LuaSnip.SnippetNode
+---@field old_state? any
+---@field dynamicNode? LuaSnip.DynamicNode
 
-	return DynamicNode:new({
+---@alias LuaSnip.DynamicNode.Fn fun(args: (string[])[], parent: LuaSnip.Snippet | LuaSnip.SnippetNode, old_state?: table, ...: table): LuaSnip.SnippetNode
+
+---@class LuaSnip.DynamicNode: LuaSnip.Node
+---@field fn LuaSnip.DynamicNode.Fn
+---@field user_args any[] Additional args that will be passed to `fn`
+---@field args LuaSnip.NodeRef[]
+---@field snip? LuaSnip.SnippetNodeForDynNode
+---@field static_snip? LuaSnip.SnippetNodeForDynNode
+---@field last_static_args? (string[])[]
+---@field snippetstring_args boolean
+local DynamicNode = Node:new()
+
+---@class LuaSnip.Opts.DynamicNode: LuaSnip.Opts.FunctionNode
+---@field snippetstring_args? boolean (FIXME: not documented?)
+
+--- Very similar to functionNode, but returns a snippetNode instead of just text,
+--- which makes them very powerful as parts of the snippet can be changed based on
+--- user input.
+---
+---@param pos integer? Just like all jumpable nodes, its' position in the
+---  jump-list ([Basics-Jump-Index](#jump-index)).
+---
+---@param fn LuaSnip.DynamicNode.Fn
+---  This function is called when the argnodes' text changes.
+---  It should generate and return (wrapped inside a `snippetNode`) nodes, which
+---  will be inserted at the dynamicNode's place.
+---
+---  note: `args`, `parent` and `user_args` are also explained in
+---  [FunctionNode](#functionnode)
+---
+---  - `argnode_text`: The text currently contained in the argnodes
+---    (e.g. `{{line1}, {line1, line2}}`).
+---
+---  - `parent`: The immediate parent of the `functionNode`. It is included here
+---    as it allows easy access to some information that could be useful in
+---    functionNodes (see [Snippets-Data](#data) for some examples).
+---
+---    Many snippets access the surrounding snippet just as `parent`, but if the
+---    `functionNode` is nested within a `snippetNode`, the immediate parent is
+---    a `snippetNode`, not the surrounding snippet (only the surrounding
+---    snippet contains data like `env` or `captures`).
+---
+---  - `old_state`: a user-defined table.
+---    This table may contain anything; its intended usage is to preserve
+---    information from the previously generated `snippetNode`.
+---    If the `dynamicNode` depends on other nodes, it may be
+---    reconstructed, which means all user input (text inserted in
+---    `insertNodes`, changed choices) to the previous `dynamicNode` is lost.
+---
+---    The `old_state` table must be stored in `snippetNode` returned by
+---    the function (`snippetNode.old_state`).
+---    The second example below illustrates the usage of `old_state`.
+---
+---  - `user_args`: The `user_args` passed in `opts`.
+---    Note that there may be multiple `user_args`.
+---    (e.g. `user_args1, ..., user_argsn`)
+---
+---@param argsnode_refs? LuaSnip.NodeRef[]|LuaSnip.NodeRef
+---  [Node References](#node-reference) to the nodes the dynamicNode depends on.
+---  Changing any of these will trigger a re-evaluation of `fn`, and the result will be inserted at the `dynamicNode`'s place.
+---  (`dynamicNode` behaves exactly the same as `functionNode` in this regard)
+---
+---@param node_opts? LuaSnip.Opts.DynamicNode
+---@return LuaSnip.DynamicNode
+local function D(pos, fn, argsnode_refs, node_opts)
+	node_opts = node_opts or {}
+
+	local node = DynamicNode:new({
 		pos = pos,
 		fn = fn,
-		args = node_util.wrap_args(args),
+		args = node_util.wrap_args(argsnode_refs or {}),
 		type = types.dynamicNode,
 		mark = nil,
-		user_args = opts.user_args or {},
-		snippetstring_args = opts.snippetstring_args or false,
+		user_args = node_opts.user_args or {},
+		snippetstring_args = node_opts.snippetstring_args or false,
 		dependents = {},
 		active = false,
-	}, opts)
+	}, node_opts)
+	---@cast node LuaSnip.DynamicNode
+	return node
 end
 extend_decorator.register(D, { arg_indx = 4 })
 
@@ -195,7 +264,8 @@ function DynamicNode:update()
 	-- build new snippet before exiting, markers may be needed for
 	-- construncting.
 	tmp =
-		self.fn(effective_args, self.parent, old_state, unpack(self.user_args))
+		self.fn(effective_args or {}, self.parent, old_state, unpack(self.user_args))
+	---@cast tmp LuaSnip.SnippetNodeForDynNode
 
 	if self.snip then
 		self.snip:exit()
@@ -332,6 +402,8 @@ function DynamicNode:update_static()
 		-- set empty snippet on failure
 		tmp = SnippetNode(nil, {})
 	end
+	---@cast tmp LuaSnip.SnippetNodeForDynNode
+
 	self.last_static_args = str_args
 
 	-- act as if snip is directly inside parent.
@@ -419,6 +491,7 @@ function DynamicNode:update_restore()
 		and vim.deep_equal(str_args, self.last_args)
 	then
 		local tmp = self.snip
+		---@cast tmp -nil
 
 		-- position might (will probably!!) still have changed, so update it
 		-- here too (as opposed to only in update).
@@ -498,7 +571,7 @@ end
 DynamicNode.make_args_absolute = FunctionNode.make_args_absolute
 DynamicNode.set_dependents = FunctionNode.set_dependents
 
-function DynamicNode:resolve_position(position, static)
+function DynamicNode:resolve_position(_position, static)
 	-- position must be 0, there are no other options.
 	if static then
 		return self.static_snip
--- a/lua/luasnip/nodes/dynamicNode.lua
+++ b/lua/luasnip/nodes/dynamicNode.lua
@@ -28,7 +28,8 @@ local session = require("luasnip.session")
 local DynamicNode = Node:new()
 
 ---@class LuaSnip.Opts.DynamicNode: LuaSnip.Opts.FunctionNode
----@field snippetstring_args? boolean (FIXME: not documented?)
+---@field snippetstring_args? boolean Preserve expanded inner snippets as
+---  snippet strings. Helps keep jump-points in self-dependent dynamic nodes.
 
 --- Very similar to functionNode, but returns a snippetNode instead of just text,
 --- which makes them very powerful as parts of the snippet can be changed based on
--- a/lua/luasnip/nodes/functionNode.lua
+++ b/lua/luasnip/nodes/functionNode.lua
@@ -47,7 +47,7 @@ local FunctionNode = Node:new()
 --- local function fn(
 ---   args,     -- text from i(2) in this example i.e. { { "456" } }
 ---   parent,   -- parent snippet or parent node
----   user_args -- user_args from opts.user_args 
+---   user_args -- user_args from opts.user_args
 --- )
 ---    return '[' .. args[1][1] .. user_args .. ']'
 --- end
--- a/lua/luasnip/nodes/functionNode.lua
+++ b/lua/luasnip/nodes/functionNode.lua
@@ -206,7 +206,8 @@ function FunctionNode:indent(_) end
 function FunctionNode:expand_tabs(_) end
 
 function FunctionNode:make_args_absolute(position_so_far)
-	self.args_absolute = node_util.make_args_absolute(self.args, position_so_far)
+	self.args_absolute =
+		node_util.make_args_absolute(self.args, position_so_far)
 end
 
 function FunctionNode:set_dependents()
--- a/lua/luasnip/nodes/functionNode.lua
+++ b/lua/luasnip/nodes/functionNode.lua
@@ -9,7 +9,6 @@ local opt_args = require("luasnip.nodes.optional_arg")
 local snippet_string = require("luasnip.nodes.util.snippet_string")
 
 ---@alias LuaSnip.FunctionNode.Fn fun(args: (string[])[], parent: LuaSnip.Snippet | LuaSnip.SnippetNode, ...: table): string|string[]
--- FIXME: how to document each param of the callback function?
 
 ---@class LuaSnip.FunctionNode: LuaSnip.Node
 ---@field fn LuaSnip.FunctionNode.Fn
@@ -94,9 +93,6 @@ local FunctionNode = Node:new()
 ---
 ---@param node_opts? LuaSnip.Opts.FunctionNode
 ---@return LuaSnip.FunctionNode
---
--- FIXME(@bew): The super flexible NodeRef param & the fact that the NodeRef
--- type is an alias, makes luals make a huge function signature here 👀
 local function F(fn, argsnode_refs, node_opts)
 	node_opts = node_opts or {}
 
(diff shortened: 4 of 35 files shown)


L3MON4D3:
Hey bew, this sounds amazing, I'm happy you got nerd-sniped like that :P

It's pretty late for me right now, but I'll take a look at this first thing tomorrow and just answer the few things I can without looking deeper into your changes :)

> As for where to go with the types design, I think the codebase would benefit from splitting the Snippet class into multiple classes to reflect the various stages of use (barebone for internal use, snippetnode, snippet, expandedsnippet, (..?)).

Oh, I totally agree here, the decision to make all of those the same thing was just because it was easy to incrementally cram more stuff into that one lua-table, but if I were to make a top-down decision right now, those'd have to be separated.

> For the actual type annotations, maybe you'd like to have a public/private interface to avoid exposing the huge internals of the nodes to the users 🤔

Having a split there sounds pretty nice, is it possible to do so via LuaCATS?
If not, it'd be enough to just not put the internal stuff into `DOC.md` :)

> I was also quite surprised that you used Node:new both to create new node types & new node instances, would you welcome some changes around this to segregate the types/functions more by use-cases?

Oh, I think that's a good idea, this is also something I've first written in 1048cdfec1a3cc406f1c5bcb5616e4953fe7238f (or even earlier :eyes:), more than four years ago, and not really questioned since :sweat_smile:

> This PR targets the self-dependent-dNode branch, since this is the branch I'm on normally.
❔ I'm wondering, do you think we could merge it soon or do you still have planned work on it?
I'm asking because otherwise I'd need to rework some parts of this PR to be able to rebase on master 👀

Nah, I'll look into merging it, that band-aid will have to come off anyway, this is a good reason to finally do that.
I'll cut a release before the merge because I'm still a bit unsure everything will work smoothly, but here's hoping :crossed_fingers: :D

Speaking a bit more on changes to luasnips' internals: while self-dependent-dNode already contains a few larger changes, I plan on making more drastic ones, up to a point where we'd (internally) only have `textNode`, `insertNode`, `dynamicNode` and `snippetNode`, which should be enough to represent any tree of nodes and emulate the behaviour of `choiceNode`, `functionNode` and `restoreNode`.
My point with this tangent is that I just wanted to make you aware that I'm eventually planning to rip out/rework large paths of luasnips internals, so best focus on the user-facing side of things, that's unlikely to change much :D (OTOH, you know how long `self-dependent-dNode` has taken to be in a good state, and anything larger in scope will likely be on a similar time-frame xD)

> I'm available to do a day/night pair-programming/review session if you want 🙃

I'd be happy to handle this similarly to the last PR we did, that was very pleasant :D

bew:
Take your time! :pray:

Really happy that you'd be okay with making some structural changes! :sparkles: 

> > For the actual type annotations, maybe you'd like to have a public/private interface to avoid exposing the huge internals of the nodes to the users 🤔
> 
> Having a split there sounds pretty nice, is it possible to do so via LuaCATS? If not, it'd be enough to just not put the internal stuff into `DOC.md` :)

Hmm actually maybe not with LuaCATS alone :eyes:
But it could be possible with a ~hidden `_` or `_private` attr containing all private state for example.
It's true that we can selectively generate stuff in DOC.md but I'm mostly thinking of in-editor luals documentation & eventual completion suggestions..

> Nah, I'll look into merging it, that band-aid will have to come off anyway, this is a good reason to finally do that. I'll cut a release before the merge because I'm still a bit unsure everything will work smoothly, but here's hoping 🤞 :D

Oh yeah :rocket:

> I'm eventually planning to rework large paths of luasnips internals, so best focus on the user-facing side of things, that's unlikely to change much :D

I personally like having types when refactoring, because the LSP is smarter and can immediately show you where things are used and how without having to hold everything in your head, and can show the impact of changes with diagnostics..
And it makes the codebase more approachable so to me it's always a net positive even when changes are planned :thinking: 

---
For the review, I think the most important ones are in `node.lua` & `snippet.lua`
And I'll probably split out small snippets of refactors / docs that don't depend on _everything else_ in separate PRs

L3MON4D3:
> Take your time! 🙏

Always :P

> But it could be possible with a ~hidden _ or _private attr containing all private state for example.

Puuuh, that does not sound appealing, nodes have like >20 fields/functions that are not supposed to be used by users.
Oh, maybe we could annotate a different class at places where the nodes are passed to users?
e.g. (from the dynamicNode update-function):
```lua
tmp = self.fn(
    effective_args,
    self.parent ---[[@as LuaSnip.PublicAPISnippetNode]],
    old_state,
    unpack(self.user_args)
)

```
and then `LuaSnip.PublicAPISnippetNode` only has the public subset of fields the real snippetNode has. We'd have a fair bit of duplication, but users aren't exposed to the internals :)

> I personally like having types when refactoring, because the LSP is smarter and can immediately show you where things are used and how without having to hold everything in your head, and can show the impact of changes with diagnostics..

Ah, that's true, I was thinking more of reimplementing the internals from scratch than an incremental change. There are lots and lots of things we don't really need anymore/have better ways of doing (for example `absolute_insert_position` was my first attempt to get dynamicNodes to reference nodes outside its snippetNode, but `key` is just way better for that).

> For the review, I think the most important ones are in node.lua & snippet.lua

:+1:

Taking a look now :eyes:

lua/luasnip/extras/init.lua
```diff
@@ -116,11 +116,18 @@ end
...
 	match = match,
-	-- repeat a node.
-	rep = function(node_indx)
+	--- Repeat a node, by inserting the text of the passed node.
+	---
+	--- ```lua
+	--- s("extras4", { i(1), t { "", "" }, extras.rep(1) })
+	--- ```
+	---
+	---@param node_ref LuaSnip.NodeRef a single [Node Reference](#node-reference).
```
L3MON4D3: Ah, you can put the relative path to the `DOC.md` in here, so the link also works when viewing the file ;)
(`[Node Reference](../../../DOC.md#node-reference)`, +- a `..` :D)

lua/luasnip/nodes/dynamicNode.lua
```diff
@@ -12,20 +11,90 @@ local log = require("luasnip.util.log").new("dynamicNode")
...
+---@field user_args any[] Additional args that will be passed to `fn`
+---@field args LuaSnip.NodeRef[]
+---@field snip? LuaSnip.SnippetNodeForDynNode
+---@field static_snip? LuaSnip.SnippetNodeForDynNode
+---@field last_static_args? (string[])[]
+---@field snippetstring_args boolean
+local DynamicNode = Node:new()
+
+---@class LuaSnip.Opts.DynamicNode: LuaSnip.Opts.FunctionNode
+---@field snippetstring_args? boolean (FIXME: not documented?)
```
L3MON4D3: Ah, `snippetstring_args` is a recent addition in `self-dependent-dNode`: it's a boolean which, when set, makes the dynamicNode pass `snippetStrings` instead of `string[]` as the first argument (snippetStrings are text copied from the buffer, with snippets preserved, so you can do stuff like `:gsub` on a nested snippet)

A bit of a problem with that the signature of `LuaSnip.DynamicNode.Fn` changes based on that boolean.. Can we properly represent that or can we only do something like `args:(string[]|LuaSnip.SnippetString)`?

lua/luasnip/nodes/dynamicNode.lua
```diff
@@ -12,20 +11,90 @@ local log = require("luasnip.util.log").new("dynamicNode")
...
 
-local function D(pos, fn, args, opts)
-	opts = opts or {}
-
-	return DynamicNode:new({
+---@class LuaSnip.SnippetNodeForDynNode: LuaSnip.SnippetNode
+---@field old_state? any
+---@field dynamicNode? LuaSnip.DynamicNode
+
+---@alias LuaSnip.DynamicNode.Fn fun(args: (string[])[], parent: LuaSnip.Snippet | LuaSnip.SnippetNode, old_state?: table, ...: table): LuaSnip.SnippetNode
```
L3MON4D3: Should the vararg be `...: any` instead of `...: table`?

lua/luasnip/nodes/dynamicNode.lua
bew: yeah it should be possible using `@overload` I think

lua/luasnip/nodes/dynamicNode.lua
bew: very true, I didn't remember how to type varargs at the time :grimacing:

lua/luasnip/nodes/dynamicNode.lua
L3MON4D3: Oh that sounds like a good substitute :+1:
I don't know if it'd even be possible for `luals` to track which dynamicNodes have `snippetstring_args` set and which don't, so overload may be as precise as we can get here

bew: yeah, the `@overload` would only help user code, to have the immediately passed callback function have the correct type, but the internal code would have zero knowledge of which node has what (like at runtime).
And the types can still _break_ (show the `(opt1|opt2)` in callback) if their code is a bit generic functions calling the node functions, and don't want to deal with overloads in their config...

lua/luasnip/nodes/dynamicNode.lua
```diff
@@ -194,8 +263,13 @@ function DynamicNode:update()
...
 	-- construncting.
-	tmp =
-		self.fn(effective_args, self.parent, old_state, unpack(self.user_args))
+	tmp = self.fn(
+		effective_args or {},
+		self.parent,
+		old_state,
+		unpack(self.user_args)
+	)
+	---@cast tmp LuaSnip.SnippetNodeForDynNode
```
L3MON4D3, line 275: WDYT about using `--[[@as ...]]` instead of the `@cast`?

bew:
> > But it could be possible with a ~hidden _ or _private attr containing all private state for example.
> 
> Puuuh, that does not sound appealing, nodes have like >20 fields/functions that are not supposed to be used by users.

Actually I noticed there is a `@field protected` thing in luals, would be worth trying!
As for function we can simply make them start with an underscore, to do like in Python with its convention of `_foo` is protected and isn't part of the public interface :thinking:

lua/luasnip/nodes/dynamicNode.lua
bew: I think `@cast` is better because it's not a HARD typing replacement, see:
image
For `@as` you can really go wild, but `@cast` is safer and doesn't allow things that don't make sense.

lua/luasnip/nodes/functionNode.lua
```diff
@@ -9,16 +8,108 @@ local key_indexer = require("luasnip.nodes.key_indexer")
 local opt_args = require("luasnip.nodes.optional_arg")
 local snippet_string = require("luasnip.nodes.util.snippet_string")
 
-local function F(fn, args, opts)
-	opts = opts or {}
+---@alias LuaSnip.FunctionNode.Fn fun(args: (string[])[], parent: LuaSnip.Snippet | LuaSnip.SnippetNode, ...: table): string|string[]
+-- FIXME: how to document each param of the callback function?
```
L3MON4D3: Yeah that's a good question :+1:
I think it's not possible to document free-standing functions like that with LuaCATS :/
For reference:
https://github.com/EmmyLuaLs/emmylua-analyzer-rust/issues/779
https://github.com/LuaLS/lua-language-server/issues/1456

I think the current way of documenting these callbacks in the dynamicNode/functionNode signature is... fine, but not very elegant :/

lua/luasnip/nodes/functionNode.lua
```diff
@@ -9,16 +8,108 @@ local key_indexer = require("luasnip.nodes.key_indexer")
...
+---  Changing any of these will trigger a re-evaluation of `fn`, and insertion of
+---  the updated text.
+---  If no node reference is passed, the `functionNode` is evaluated once upon
+---  expansion.
+---
+---@param node_opts? LuaSnip.Opts.FunctionNode
+---@return LuaSnip.FunctionNode
+--
+-- FIXME(@bew): The super flexible NodeRef param & the fact that the NodeRef
+-- type is an alias, makes luals make a huge function signature here 👀
```
L3MON4D3: Settings `hover.expandAlias` in `lua-language-server` to false makes this work pretty well

screen

Honestly, setting that does not seem like a bad idea at all :eyes: The only annoying bit is that it doesn't seem like one can jump to definition from the documentation window :/
(and obviously that not everyone will want to set this only to make this annotation work :D)

lua/luasnip/nodes/insertNode.lua
```diff
@@ -12,10 +11,40 @@ local str_util = require("luasnip.util.str")
...
+--- ```lua
+--- s("trigger", {
+--- 	t({"After expanding, the cursor is here ->"}), i(1),
+--- 	t({"", "After jumping forward once, cursor is here ->"}), i(2),
+--- 	t({"", "After jumping once more, the snippet is exited there ->"}), i(0),
+--- })
+--- ```
+---
+--- FIXME(@bew): how to put the whole InsertNode documentation here, how to deal
+--- with the gifs here?
```
L3MON4D3: I think we can omit the gifs, they can be added via code in `DOC.md` (or just by putting them into a separate section that gives more of an overview of `insertNode`, compared to this detailed documentation)

lua/luasnip/nodes/insertNode.lua
```diff
@@ -389,6 +419,7 @@ function InsertNode:get_snippetstring()
 			)
 			snippetstring:append_snip(
 				snip,
+				-- FIXME(@bew): This param does not exist on append_snip? 👀
```
L3MON4D3: Oh, that's a mistake, get rid of it :+1:

lua/luasnip/nodes/insertNode.lua
```diff
@@ -37,8 +66,9 @@ local function I(pos, static_text, opts)
 			type = types.insertNode,
 			inner_active = false,
 			input_active = false,
-		}, opts)
+		}, node_opts)
 	end
+	---@cast node LuaSnip.InsertNode|LuaSnip.ExitNode
```
L3MON4D3, line 119: Does luals understand what's going on if you put this into the respective branches of the `if`?

lua/luasnip/nodes/multiSnippet.lua
```diff
@@ -10,6 +12,8 @@ function VirtualSnippet:get_docstring()
 end
 function VirtualSnippet:copy()
 	local copy = self.snippet:copy()
+	-- FIXME(@bew): why does the VirtualSnippet have an ID ?
+	--   It's never added to a collection is it?
```
L3MON4D3: Ah, they are: the multiSnippet is passed to `add_snippets`, and there `:retrieve_all()` returns all its virtual snippets.

lua/luasnip/nodes/node.lua
```diff
@@ -10,37 +10,66 @@ local snippet_string = require("luasnip.nodes.util.snippet_string")
...
----@field type number Identifies the type of the snippet.
+---@field active? boolean
+---@field type LuaSnip.NodeType Identifies the type of the snippet.
 ---@field next LuaSnip.Node Link to the next node in jump-order.
 ---@field prev LuaSnip.Node Link to the previous node in jump-order.
+---@field parent LuaSnip.Snippet|LuaSnip.SnippetNode The parent snippet or
+---  snippet node.
+---@field indx ... (FIXME(@bew): what is this? is it an indexer?)
+---  It looks like it's something used like an integer,
+---  but other times `:resolve` is called on it 🤔
```
L3MON4D3: Ah, this is an integer, the `indx` with `:resolve` is from `ParentIndexer` (same field-name, different class)

lua/luasnip/nodes/node.lua
```diff
@@ -10,37 +10,66 @@ local snippet_string = require("luasnip.nodes.util.snippet_string")
...
+---
+---@field visible boolean
+---@field static_text string[] (FIXME(@bew): Where is this initialized?)
```
L3MON4D3: `static_text` for insertNodes and textNodes is initialized in their constructor, for snippet(Node) and restoreNode, `:get_static_text` caches its result in there (but never invalidates it :grimacing:)

L3MON4D3: So, the `Node:get_static_text` is a fallback used only when the node actually has a `static_text`, those that don't have their own `:get_static_text` that does not access `self.static_text`

lua/luasnip/nodes/insertNode.lua
bew: probably yes, you'd prefer to do this?

lua/luasnip/nodes/node.lua
```diff
@@ -223,7 +252,11 @@ function Node:init_insert_positions(position_so_far)
 	self.absolute_insert_position = vim.deepcopy(position_so_far)
 end
 
+---@param event LuaSnip.EventType
 function Node:event(event)
+	-- FIXME(@bew): wrong type for node_callbacks ?
+	-- We index witha EventType(integer), but the field definition uses string
+	-- keys "enter"/"leave" 🤔
```
L3MON4D3: Oh yeah you're totally right, it's not `"enter","leave"`, it should be `events.enter,events.leave` (so, numbers is correct)

lua/luasnip/nodes/optional_arg.lua
```diff
@@ -1,5 +1,7 @@
 local M = {}
 
+-- FIXME: This is not documented!
+
```
L3MON4D3: Right, another new thing from `self-dependent-dNode`, I'll write some docs for that soon!

lua/luasnip/nodes/snippet.lua
```diff
@@ -66,20 +70,65 @@ local stored_mt = {
 	end,
 }
 
----@class LuaSnip.Snippet: LuaSnip.Addable, LuaSnip.ExpandedSnippet
+---@class LuaSnip.BareInternalSnippet: LuaSnip.Node
```
L3MON4D3, line 64: I don't quite get at what point a snippet is a `BareInternalSnippet`.. In my mind there is `Addable` for things that can be put into `add_snippets`, `Expandable` for stuff that has `:matches` and `condition` and the like, and then there is `ExpandedSnippet` for when the snippet is expanded.

As far as I remember `snippet`, `dependents_dict`, `child_snippets`, `static_text`, and `indentstr` are only used by snippets once they are expanded, maybe `ExpandedSnippet` is a better place for them? (And maybe it'd be a good idea to initialize them in `trigger_expand`?)

lua/luasnip/nodes/snippet.lua
```diff
@@ -66,20 +70,65 @@ local stored_mt = {
...
+---@field priority? integer
+---@field snippetType? "snippets"|"autosnippets"
+---@field filetype? string
+---@field wordTrig boolean
+---@field hidden boolean
+---@field regTrig boolean
+---@field docTrig? string
+---@field trig_matcher LuaSnip.SnipContext.TrigMatcher
+---@field resolveExpandParams LuaSnip.ResolveExpandParamsFn
+---@field show_condition fun(line_to_cursor: string): boolean
```
L3MON4D3: Also `condition`, no?

lua/luasnip/nodes/snippet.lua
```diff
@@ -178,34 +234,48 @@ local function init_snippetNode_opts(opts)
...
+	local in_node = {} ---@diagnostic disable-line: missing-fields
 
-	-- return sn(t("")) for so-far-undefined keys.
-	in_node.stored = setmetatable(opts.stored or {}, stored_mt)
+	-- The metatable will return `sn(t(""))` for so-far-undefined keys.
+	in_node.stored = setmetatable({}, stored_mt)
 
 	-- wrap non-snippetNode in snippetNode.
-	for key, nodes in pairs(in_node.stored) do
+	for key, nodes in pairs(opts.stored or {}) do
```
L3MON4D3, line 267: Very good, way more clear :+1:

lua/luasnip/nodes/dynamicNode.lua
L3MON4D3: I see, that's definitely an advantage, let's keep using `@cast` in that case :+1:

lua/luasnip/nodes/insertNode.lua
L3MON4D3: I mostly asked out of curiosity, may you had tried it and it didn't work :D
I'm fine with this too, I don't think it makes a non-negligible difference

lua/luasnip/nodes/snippet.lua
bew: Of course...
I remember thinking about them both, and then I only did one  :sweat_smile:

lua/luasnip/nodes/snippet.lua
```diff
@@ -385,20 +464,177 @@ local function _S(snip, nodes, opts)
...
+---@return LuaSnip.Snippet
 local function S(context, nodes, opts)
 	opts = opts or {}
 
 	local snip = init_snippet_context(node_util.wrap_context(context), opts)
-	snip = vim.tbl_extend("error", snip, init_snippet_opts(opts))
+	local snip = vim.tbl_extend("error", snip, init_snippet_opts(opts))
 
-	snip = _S(snip, nodes, opts)
+	local snip = _S(snip, nodes, opts)
```
L3MON4D3: Why the re-definitions? I think we should have different names if they are necessary

bew: IIRC the redefinition was to ensure the final type was automatically the return type from `_S`, without luals trying to do some merges or whatnot with the earlier assignments :thinking:

lua/luasnip/nodes/snippet.lua
```diff
@@ -636,6 +913,12 @@ function Snippet:insert_into_jumplist(
 	table.insert(sibling_snippets, own_indx, self)
 end
 
+-- IDEA(THINKING, @bew): Most methods in Snippet should really be on a BareInternalSnippet class
+-- But this method should be on an actual Snippet class
+-- (so it can be called on a full Snippet, but not a BareInternalSnippet)
+-- This came to my mind because this function uses `self.stored` &
+-- `self.merge_child_ext_opts` that _only_ exist on full Snippet but not on
+-- BareInternalSnippet.
```
L3MON4D3, line 965: :+1:
Splitting Snippet's responsibilities is a really good idea, it definitely has the weirdest lifecycle out of all nodes.

lua/luasnip/nodes/snippet.lua
```diff
@@ -875,6 +1144,7 @@ function Snippet:del_marks()
 	end
 end
 
+-- FIXME: `info` is _never_ used anywhere?
```
L3MON4D3: Oh, this is very hidden: `is_interactive` is overridden for a few ast-parsed nodes, and _that_ version uses `info` (or, more precisely, `env`)

lua/luasnip/nodes/snippet.lua
L3MON4D3: Ahh alright, that makes sense.
I think the first one could be `snip_with_opts` instead of `snip`

lua/luasnip/nodes/snippet.lua
```diff
@@ -1184,6 +1461,12 @@ function Snippet:text_only()
 	return true
 end
 
+--- Trigger event with args
+---@param event LuaSnip.EventType
+---@param event_args? table
+---@return any
+-- FIXME(@bew): This should only be on SnippetNode & Snippet 🤔
+--   (to access callbacks)
```
L3MON4D3, line 1515: You mean this should only be defined on the types used at runtime? I agree :D

lua/luasnip/nodes/util.lua
```diff
@@ -851,6 +863,8 @@ local function collect_dependents(node, which, static)
 	return tbl_util.set_to_list(dependents_set)
 end
 
+---@param args (string[])?
```
L3MON4D3: `args` should be `(string|SnippetString)[]` I think

lua/luasnip/nodes/util/snippet_string.lua
```diff
@@ -1,7 +1,30 @@
...
+---  A kind-of extmark to the text in this buffer.
+---
+---  It moves with inserted text, and has a gravity to control into which
+---  direction it shifts. pos is 1-based and refers to one character in the
+---  string
+---
+---  If the edge is in the middle of multiple characters (for example rgrav=true,
+---  and chars at pos and pos+1 are replaced), the mark is removed.
+---
+---  FIXME(@bew): How is it different than extmarks? (-> why not use them?)
```
L3MON4D3: Completely right, this is an extmark for simple lua-strings (not neovim buffers) :D

lua/luasnip/nodes/util/snippet_string.lua
```diff
@@ -130,6 +154,7 @@ function SnippetString:put(pos)
 	end
 end
 
+-- FIXME(@bew): what is the return type here?
```
L3MON4D3: Pretty sure this should just be another `SnippetString`

lua/luasnip/session/snippet_collection/source.lua
```diff
@@ -1,7 +1,14 @@
+---@class LuaSnip._Source
```
L3MON4D3: How about `LuaSnip.Source`? I don't see the reason for the `_`

lua/luasnip/util/ext_opts.lua
```diff
@@ -78,18 +80,22 @@ end
...
 
 	return ext_opts
 end
 
+---@param ext_opts? LuaSnip.NodeExtOpts
+---@return LuaSnip.NodeExtOpts
 local function complete(ext_opts)
-	_complete_ext_opts(ext_opts)
-	ext_opts.base_prio = 0
+	local ext_opts = _complete_ext_opts(ext_opts)
```
L3MON4D3, line 97: Good catch :+1:

lua/luasnip/util/table.lua
```diff
@@ -1,7 +1,7 @@
 ---Convert set of values to a list of those values.
 ---@generic T
----@param tbl T|T[]|{[T]: boolean}
----@return {[T]: boolean}
+---@param tbl T[]|{[T]: boolean}
```
L3MON4D3: Shouldn't this just be `{[T]:boolean}`?

lua/luasnip/util/events.lua
```diff
@@ -1,18 +1,27 @@
...
-		elseif event_id == 4 then
-			return "PreExpand"
-		else
-			return node_names[node_type]
-				.. (event_id == 1 and "Enter" or "Leave")
-		end
-	end,
 }
+
+local M = setmetatable({}, { __index = EventType })
```
L3MON4D3, line 11: I don't get why we need the metatable here, could you explain?

bew: So that callers of this `events` module can do `events.change_choice` to get a named value from the enum

The thing is, if we try to go back by removing `M` & just have `EventType` everywhere, I'm pretty sure that adding the `to_string` member will just break the `@enum` (the function shouldn't be part of the enum!)

lua/luasnip/util/table.lua
bew: oh right, I missed that the value is not used, only the keys, so for a list it doesn't make much sense

lua/luasnip/util/events.lua
L3MON4D3: Ahh I see, that makes sense :+1:
Would you put a comment about that in there?

lua/luasnip/util/types.lua
```diff
@@ -8,27 +9,57 @@ return {
...
+---@type string[]
+M.names = {}
+---@type string[]
+M.names_pascal_case = {}
+
+for _, ref in ipairs(refs) do
+	table.insert(M.node_types, ref.value)
+	table.insert(M.names, ref.name)
+	table.insert(M.names_pascal_case, ref.pascal_name)
+end
```
L3MON4D3, line 66: This is a nice improvement :+1:

lua/luasnip/nodes/choiceNode.lua
```diff
@@ -111,8 +121,13 @@ extend_decorator.register(ChoiceNode.C, { arg_indx = 3 })
...
 	for _, choice in ipairs(self.choices) do
 		choice.parent = self.parent
-		-- only insertNode needs this.
-		if choice.type == 2 or choice.type == 1 or choice.type == 3 then
+		-- only insertNode needs this. (?)
+		if
+			vim.list_contains(
+				{ types.textNode, types.insertNode, types.functionNode },
+				choice.type
+			)
```
L3MON4D3: I just ran the tests and apparently `list_contains` does not exist in nvim 0.7.0 :|
Could you put a quick implementation into `util/table.lua`?

lua/luasnip/nodes/insertNode.lua
bew: > or just by putting them into a separate section that gives more of an overview of insertNode, compared to this detailed documentation

I like this idea, to split the reference-level docs from the guide-level docs.
In the meantime I'll do a sort of hybrid but'll add a note for this somewhere

L3MON4D3: Yes, well put :+1:

lua/luasnip/nodes/functionNode.lua
bew: Thanks I added it to my config :+1: (and removed this comment)

👉 Could you maybe add this tip into a kind of `editor support` section of the DOC ?
(near the other one about completion :thinking: )

bew:
I rebased on `master` since you merged #1137 (wooo :rocket:)

How do you want to handle me fixing things incrementally?
I usually like to leave the commenter resolve each of his comment (so you can check they were fixed properly), and I can just put a `done` comment to notify that this one can be checked.
But I can work differently :thinking:

lua/luasnip/nodes/dynamicNode.lua
bew: done in 344ecda1874f7a8b1ffd3fb071cae7daf59c29fe

lua/luasnip/nodes/insertNode.lua
bew: done in 344ecda1874f7a8b1ffd3fb071cae7daf59c29fe

lua/luasnip/nodes/multiSnippet.lua
bew: done in 344ecda1874f7a8b1ffd3fb071cae7daf59c29fe

lua/luasnip/nodes/functionNode.lua
L3MON4D3: Mhmm, sounds reasonable :+1:

lua/luasnip/extras/init.lua
bew: Done in 02cabf1378f3d49146e4add6ddfaa8d0324ea2fb

lua/luasnip/nodes/node.lua
bew: Done in 02cabf1378f3d49146e4add6ddfaa8d0324ea2fb

lua/luasnip/util/events.lua
bew: Done in 02cabf1378f3d49146e4add6ddfaa8d0324ea2fb

lua/luasnip/nodes/choiceNode.lua
bew: Actually I used `vim.tbl_contains` which you used some place else for the same kind of usecase.
Done in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/nodes/util.lua
bew: No? I can only find calls with list of string :thinking: 
(the `get_args` local fn in node always return list of string)

lua/luasnip/nodes/snippet.lua
bew: good catch, I tried to find usage with luals, but ripgrep found more :grimacing:

lua/luasnip/util/table.lua
bew: Done in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/session/snippet_collection/source.lua
bew: Done in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/nodes/util/snippet_string.lua
bew: Done in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/nodes/optional_arg.lua
bew: Fixed in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/nodes/snippet.lua
bew: Done in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/nodes/snippet.lua
bew: Done in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/nodes/util/snippet_string.lua
bew: Updated in 6204bbd3e99526f766bd029773c5f19145b25ef8

lua/luasnip/nodes/node.lua
bew: How should that field be documented?
Or I just remove my FIXME & let you fill in the missing internal docs later

lua/luasnip/nodes/node.lua
bew: Thanks, fixed in 164183832079b8bc51cab0edfa5359a705d73370

lua/luasnip/nodes/insertNode.lua
bew: I added the rest of the insert node docs in 164183832079b8bc51cab0edfa5359a705d73370
The DOC.md generator will need to be updated but I 'included' the gifs with a line like `<demo-gif:SomeID>` :thinking:

lua/luasnip/nodes/optional_arg.lua
L3MON4D3: Clean :+1:

lua/luasnip/util/events.lua
L3MON4D3: Perfect :+1:

lua/luasnip/nodes/choiceNode.lua
L3MON4D3: For such a small table any difference is surely negligible :+1:

lua/luasnip/nodes/util.lua
L3MON4D3: Ahh, that is a bit wacky: the local `get_args` calls either `argnode_text` or `get_static_snippetstring` (function-name passed as a string :grimacing:), and they both can return `snippetsStrings` for e.g. `insertNode`.

lua/luasnip/nodes/snippet.lua
L3MON4D3: Yeah, there are a few.. unorthodox things like that in the codebase :D

lua/luasnip/nodes/node.lua
L3MON4D3: Make it a `FIXME(@L3MON4D3)` :D

lua/luasnip/nodes/insertNode.lua
L3MON4D3: I think including this in `luals-mdgen` should be straightforward, and the syntax looks fine (it's even being highlighted for me, what is it for in regular markdown?)

lua/luasnip/util/events.lua
L3MON4D3: Ah, and also great that you fixed this for `types` as well 😊

lua/luasnip/nodes/util.lua
```diff
@@ -45,6 +45,7 @@ end
 ---@param args LuaSnip.NodeRef[]
 ---@param parent_insert_position integer[]
 ---@param target LuaSnip.NormalizedNodeRef[] Target for the normalized node refs
+-- FIXME(@bew): Why is the `target` param updated instead of returning a new table?
```
L3MON4D3: Oh, yeah that's stupid :D
Feel free to change that :)

lua/luasnip/nodes/node.lua
```diff
@@ -27,12 +27,10 @@ local describe = require("luasnip.util.log").describe
...
 ---@field parent LuaSnip.Snippet|LuaSnip.SnippetNode The parent snippet or
 ---  snippet node.
----@field indx ... (FIXME(@bew): what is this? is it an indexer?)
----  It looks like it's something used like an integer,
----  but other times `:resolve` is called on it 🤔
+---@field indx integer Index of the node in the snippet or snippet node.
 ---
 ---@field visible boolean
----@field static_text string[] (FIXME(@bew): Where is this initialized?)
+---@field static_text string[] (FIXME(@bew): What is this for?)
```
L3MON4D3: Funny diff :D
It's used in `put_intial`, it's the text that will actually appear in the buffer when the snippet is expanded, and it is updated to represent the last text that the user typed into this node (via `:store()`)

lua/luasnip/nodes/insertNode.lua
bew: > (it's even being highlighted for me, what is it for in regular markdown?)

FYI: In markdown it's an explicit link

lua/luasnip/nodes/util.lua
bew: Done in 85b9618bf87fd3a37019c0567460b801ae8895f2

lua/luasnip/nodes/util.lua
bew: Of course you're right, I actually saw that but missed a level & thought the conversion to strings[] was at that level as well, but no ><
Fixed in 85b9618bf87fd3a37019c0567460b801ae8895f2 :innocent:

lua/luasnip/nodes/node.lua
bew: Done in 85b9618bf87fd3a37019c0567460b801ae8895f2 :innocent:

lua/luasnip/nodes/snippet.lua
```diff
@@ -630,6 +631,11 @@ end
 
 ---@class LuaSnip.Opts.Snippet: LuaSnip.Opts.SnippetNode
 ---@field stored? {[string]: LuaSnip.Node} Snippet-level state for restore node.
+---
+---@field show_condition? LuaSnip.SnipContext.ShowConditionFn Same as
+---  `show_condition` in snippet context. (here for backward compat)
+---@field condition? LuaSnip.SnipContext.Condition Same as `condition` in
+---  snippet context. (here for backward compat)
```
bew, line 658: not sure about why these are possible to specify in the snippet opts, is it for backward compat?

lua/luasnip/nodes/snippet.lua
bew: I had to make something for the return type of `_S`, which is used in different places for defining different types.
And it's made to have the common fields for all other classes that are derived from it.

It's used as the common base for `Snippet` & `SnippetNode`:

- `---@class LuaSnip.SnippetNode: LuaSnip.BareInternalSnippet, LuaSnip.NormalizedSnippetNodeOpts`
- `---@class LuaSnip.Snippet: LuaSnip.BareInternalSnippet, LuaSnip.NormalizedSnippetContext, LuaSnip.NormalizedSnippetOpts, LuaSnip.Addable`

Those two don't share all the same fields nor the same parent classes, so we need a kind of base class for all their common fields :thinking:

---

> As far as I remember snippet, dependents_dict, child_snippets, static_text, and indentstr are only used by snippets once they are expanded, maybe ExpandedSnippet is a better place for them? (And maybe it'd be a good idea to initialize them in trigger_expand?)

Interesting, I'll have to explore that!

lua/luasnip/nodes/insertNode.lua
L3MON4D3: Ahh that's pretty cool :+1:
Check out https://github.com/L3MON4D3/luals-mdgen/commit/f4f134b8322f62ae14c42aefb4aec04510658906, it implements almost exactly what you suggested (only `<media:SomeID>` to make it a bit more generic).

These media-links are ignored if their target is not specified, so we no longer need the whole `<panvimdoc-ignore>` thing, we can just not set `media_mapping` when generating the markdown for the vimdoc :)

lua/luasnip/nodes/util.lua
L3MON4D3: Clean :+1:

lua/luasnip/nodes/util.lua
L3MON4D3: Ahh yeah I see :D
The fix looks good :+1:

lua/luasnip/nodes/node.lua
L3MON4D3: Perfect :)

lua/luasnip/nodes/snippet.lua
L3MON4D3: Yeah exactly, they used to be in that table, and until now I don't think there's been a big reason to actually remove them. Best put something about them being deprecated in there? Unfortunately `@deprecated` only works for functions :cry:

lua/luasnip/nodes/snippet.lua
L3MON4D3: > I had to make something for the return type of _S, which is used in different places for defining different types.
And it's made to have the common fields for all other classes that are derived from it.

> It's used as the common base for Snippet & SnippetNode

Ahh I see.. this seems fine :+1:
So far I'd thought of snippet as a specialization of snippetNode, but it's possible that, rigourously, there are some conflicts there :sweat_smile:

lua/luasnip/nodes/snippet.lua
```diff
@@ -125,6 +128,8 @@ local Snippet = node_mod.Node:new()
 ---@class LuaSnip.NormalizedSnippetOpts: LuaSnip.NormalizedSnippetNodeOpts
 ---@field stored {[string]: LuaSnip.SnippetNode}
 
+-- FIXME(@bew): What is this for? (not documented..)
+-- FIXME(@bew): Should be moved to its own file? (like the other indexers)
```
L3MON4D3, line 136: I actually don't know if it is even used anywhere.. it was the first idea for referencing a node that is not a sibling (so no relative jump-index possible).
Checking the code, I can see no references to it, best ignore it for now/rip it out in a separate commit :)

lua/luasnip/util/feedkeys.lua
```diff
@@ -115,10 +121,12 @@ local function cursor_set_keys(pos, before)
 		.. ","
 		-- -1 works for multibyte because of rounding, apparently.
 		.. pos[2]
-		.. "})"
-		.. "<cr><cmd>:silent! foldopen!<cr>"
+		.. "})<cr>"
+		.. "<cmd>:silent! foldopen!<cr>"
```
L3MON4D3, line 125: Pretty :+1:

lua/luasnip/util/mark.lua
```diff
@@ -1,17 +1,30 @@
 local session = require("luasnip.session")
 local util = require("luasnip.util.util")
 
----@class LuaSnip.Mark
+---@class LuaSnip.Pos0 A [row, col] position in a buffer, 0-indexed.
+---@field [1] integer
+---@field [2] integer
```
L3MON4D3, line 6: I think 0|1 for the first column index is good and important, how about another 0|1 for first row index?
I.e. `Pos00` and `RawPos00`?

lua/luasnip/util/mark.lua
```diff
@@ -94,18 +121,26 @@ function Mark:pos_begin_raw()
...
+---
+--- note: opts as first arg because positions are pretty likely to stay the same.
+---
+---@param opts vim.api.keyset.set_extmark
+---@param pos_begin LuaSnip.RawPos0
+---@param pos_end LuaSnip.RawPos0
 function Mark:update(opts, pos_begin, pos_end)
 	-- if one is changed, the other is likely as well.
 	if not pos_begin then
+		-- FIXME(@bew): old_pos_begin & old_pos_end don't exist??
```
L3MON4D3: I'm pretty sure this entire function is not used anywhere, neither "go to references", nor `grep ':update\('` show any calls xD

lua/luasnip/util/mark.lua
```diff
@@ -181,6 +226,9 @@ function Mark:set_rgrav(which, rgrav)
 	self:set_opts(self.opts)
 end
 
+--- Returns the raw (byte) position of the wanted side
+---@param which -1|1
```
L3MON4D3, line 199: I think this type of side-parameter is used in other functions too, could be a good candidate for an enum

bew: Implemented in #1420

lua/luasnip/util/mark.lua
bew: Removed in #1420

L3MON4D3:
> I think we can defer work on DOC.md 🤔

Yeah for sure, there's no rush :+1:

PR #1408, opened 2025-12-18
add quickstart section to README

eduardoarandah:
Using luasnips functions directly feel like a job for a compiler 😅

image

We are all super busy and lazy, that's why I've added a quickstart so luasnips can be used **just as easy as textmate**!

For example, adding `[==========[` makes it super easy for the eyes to locate the "template" 

https://github.com/user-attachments/assets/6c208566-3941-4df2-a6cc-8558ad860c0b

I also added two commands:

- `EditLuasnipSnippets` to quickly jump to your luasnip snippets
- `PasteAsLuasnipSnippet` to paste your clipboard content as a snippet (escaping { } signs as {{ }})

See them in action:

https://github.com/user-attachments/assets/d9a399b3-c497-47d5-98f3-cf872f227cd6

### DIFF
--- a/doc/luasnip.txt
+++ b/doc/luasnip.txt
@@ -1,4 +1,4 @@
-*luasnip.txt*         For NeoVim 0.7-0.11        Last change: 2025 November 03
+*luasnip.txt*         For NeoVim 0.7-0.11        Last change: 2025 December 18
 
 ==============================================================================
 Table of Contents                                  *luasnip-table-of-contents*


eduardoarandah:
Yeah, that’s the idea, one can only have opinions after having experience which is hard as a newbie. 

May I suggest adding your opinionated examples instead? 😄 for simple stuff. Text and placeholder. 

My experience with luasnips was that at the beginning all I wanted was simple text with simple placeholders. I found only complex examples so I started using snipmate format. 

A thousand snippets after, I found that I could get that level of simplicity with luasnips, but it’s now too late 😄 my snippets folder is chaos 

and yeah. I will submit my examples to the wiki

PR #1410, opened 2025-12-20
Update README with BSD/GNU Makefile distinction

tywkeene:
Added a little call-out in the readme for users regarding GNU make installation.

This has bitten me a few times and it took a moment to remember what to do. I have no opinions on whatever GNU/BSD makefile choices are made in the project, just that it's documented so I can reference it later.

Thanks.


L3MON4D3:
Ah, yes, very good, that's a good place to mention this :+1:
Thank you!

PR #1418, opened 2026-01-10
util: Add types & docs for SnippetString

bew:
Extracted from #1396 

Depends on #1420

I still have 4 warnings in that file, that I'm not sure how to fix 👀

### DIFF
--- a/lua/luasnip/nodes/util/snippet_string.lua
+++ b/lua/luasnip/nodes/util/snippet_string.lua
@@ -43,8 +43,8 @@ function SnippetString.new(initial_str, metadata)
 	return setmetatable(o, SnippetString_mt)
 end
 
---- Returns whether the given obj is a SnippetString instance.
----@param o table
+--- Returns whether the given value is a SnippetString instance.
+---@param o any
 ---@return boolean
 function SnippetString.isinstance(o)
 	return getmetatable(o) == SnippetString_mt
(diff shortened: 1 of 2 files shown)


lua/luasnip/nodes/util/snippet_string.lua
```diff
@@ -23,13 +43,18 @@ function M.new(initial_str, metadata)
 	return setmetatable(o, SnippetString_mt)
 end
 
-function M.isinstance(o)
+--- Returns whether the given obj is a SnippetString instance.
+---@param o table
```
L3MON4D3: Oh, this can also be string, `any` should be fine.

lua/luasnip/nodes/util/snippet_string.lua
```diff
@@ -39,6 +64,10 @@ end
 -- * each component in the snippet_string (including nested) to the text-index
 --   of its first character.
 -- * the string of each nested snippetString.
+---@param self LuaSnip.SnippetString
+---@param map table
```
L3MON4D3, line 68: `map` is not a good type because it maps:
* SnippetString+integer to offset in the snippetString
* snippet to its string
* SnippetString to their string

I guess it'd be cleaner to have separate tables for each of these, I'll make a note to do this.
Until then table is fine :+1:

lua/luasnip/nodes/util/snippet_string.lua
bew: done in bb1c84d

PR #1420, opened 2026-01-10
Rework pos types, add types for mark utils

bew:
Extracted from #1396 (and reworked a little)

- Rework position types
- Add types for all Mark utils
- Remove unused `Mark:update` method ([ref](https://github.com/L3MON4D3/LuaSnip/pull/1396#discussion_r2678564111))

### DIFF
--- a/lua/luasnip/extras/_treesitter.lua
+++ b/lua/luasnip/extras/_treesitter.lua
@@ -291,10 +291,10 @@ function TSParser.new(bufnr, parser, source)
 	return o
 end
 
----@param pos { [1]: number, [2]: number }?
+---@param pos LuaSnip.RawPos00
 ---@return TSNode?
 function TSParser:get_node_at_pos(pos)
-	pos = vim.F.if_nil(pos, util.get_cursor_0ind())
+	local pos = pos or util.get_cursor_0ind()
 	local row, col = pos[1], pos[2]
 	assert(
 		row >= 0 and col >= 0,
@@ -308,7 +308,7 @@ function TSParser:get_node_at_pos(pos)
 end
 
 ---Get the root for the smallest tree containing `pos`.
----@param pos { [1]: number, [2]: number }
+---@param pos LuaSnip.RawPos00
 ---@return TSNode?
 function TSParser:root_at(pos)
 	local tree = self.parser:tree_for_range(
@@ -323,7 +323,7 @@ function TSParser:root_at(pos)
 end
 
 ---@param match_opts LuaSnip.extra.EffectiveMatchTSNodeOpts
----@param pos { [1]: number, [2]: number }
+---@param pos LuaSnip.RawPos00
 ---@return LuaSnip.extra.NamedTSMatch?, TSNode?
 function TSParser:match_at(match_opts, pos)
 	-- Since we want to find a match to the left of pos, and if we accept there
--- a/lua/luasnip/extras/_treesitter.lua
+++ b/lua/luasnip/extras/_treesitter.lua
@@ -28,7 +28,7 @@ local function inspect_node(node)
 end
 
 ---@param bufnr number
----@param region LuaSnip.MatchRegion
+---@param region LuaSnip.Region00InLine
 ---@return LanguageTree, string
 local function reparse_buffer_after_removing_match(bufnr, region)
 	local lang = get_lang(bufnr)
@@ -56,11 +56,11 @@ end
 ---@class LuaSnip.extra.FixBufferContext
 ---@field ori_bufnr number
 ---@field ori_text string
----@field region LuaSnip.MatchRegion
+---@field region LuaSnip.Region00InLine
 local FixBufferContext = {}
 
 ---@param ori_bufnr number
----@param region LuaSnip.MatchRegion
+---@param region LuaSnip.Region00InLine
 ---@return LuaSnip.extra.FixBufferContext
 function FixBufferContext.new(ori_bufnr, region, region_content)
 	local o = {
--- a/lua/luasnip/util/mark.lua
+++ b/lua/luasnip/util/mark.lua
@@ -50,8 +50,8 @@ local function bytecol_to_utfcol(pos)
 end
 
 --- Returns the (utf-32) begin/end positions of the mark.
----@return LuaSnip.Pos00 _ begin position, (0,0)-indexed
----@return LuaSnip.Pos00 _ end position, (0,0)-indexed
+---@return LuaSnip.Pos00 begin Begin position, (0,0)-indexed
+---@return LuaSnip.Pos00 end End position, (0,0)-indexed
 function Mark:pos_begin_end()
 	local mark_info = vim.api.nvim_buf_get_extmark_by_id(
 		0,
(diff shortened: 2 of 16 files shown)


lua/luasnip/extras/_treesitter.lua
```diff
@@ -291,10 +291,10 @@ function TSParser.new(bufnr, parser, source)
 	return o
 end
 
----@param pos { [1]: number, [2]: number }?
+---@param pos LuaSnip.RawPos00
 ---@return TSNode?
 function TSParser:get_node_at_pos(pos)
-	pos = vim.F.if_nil(pos, util.get_cursor_0ind())
+	local pos = pos or util.get_cursor_0ind()
```
L3MON4D3, line 320: I'm fine with this either way, but is there a difference when redeclaring `pos` like this?

lua/luasnip/util/mark.lua
```diff
@@ -31,12 +41,17 @@ local function mark(pos_begin, pos_end, opts)
...
+---@return LuaSnip.Pos00
 local function bytecol_to_utfcol(pos)
 	local line = vim.api.nvim_buf_get_lines(0, pos[1], pos[1] + 1, false)
 	-- line[1]: get_lines returns table.
 	return { pos[1], util.str_utf32index(line[1] or "", pos[2]) }
 end
 
+--- Returns the (utf-32) begin/end positions of the mark.
+---@return LuaSnip.Pos00 _ begin position, (0,0)-indexed
+---@return LuaSnip.Pos00 _ end position, (0,0)-indexed
```
L3MON4D3: I think I asked something equivalent before, but I don't recall the result and can't find it:
What do you think about replacing `_` with `begin` and `end` respectively?

Image

vs

Image

I see that it's a tiny bit repetetive, but imo not too bad.

lua/luasnip/extras/_treesitter.lua
bew: IIRC in this case the redefinition of the var as a local helps the typing to be correctly detected as non-nil for the rest of the function

lua/luasnip/util/mark.lua
bew: I don't recall either, but here you go in fca4b28

lua/luasnip/util/util.lua
```diff
@@ -1,19 +1,21 @@
 local session = require("luasnip.session")
 local vimversion = require("luasnip.util.vimversion")
 
+---@return LuaSnip.RawPos00
 local function get_cursor_0ind()
 	local c = vim.api.nvim_win_get_cursor(0)
```
bew, line 6: note: I'm honestly not sure about using raw pos here, but vim docs are not super clear afaik about when the result is byte-level and when it's not..
At the beginning I had non-raw pos here but iirc at some point I had a type conflict so I switched all to raw pos 🤔

lua/luasnip/util/mark.lua
```diff
@@ -181,32 +192,38 @@ function Mark:set_rgrav(which, rgrav)
...
+--- Invalidate this mark object only, leave the underlying extmark alone.
 function Mark:invalidate()
 	self.id = nil
 end
 
+--- Delete the underlying extmark if any.
 function Mark:clear()
 	if self.id then
 		vim.api.nvim_buf_del_extmark(0, session.ns_id, self.id)
+		-- FIXME(@bew): Should also invalidate the Mark obj? (self.id = nil)
```
L3MON4D3: Mhmmm probably, yeah :+1:
If we try to access a mark after it has been deleted there's a problem further up, but it seems proper anyway

bew: Done in 0f29e82

lua/luasnip/util/util.lua
```diff
@@ -199,6 +204,13 @@ local function wrap_nodes(nodes)
 	end
 end
 
+--- Returns whether the given positions are equal
+---
+---@param p1 LuaSnip.RawPos00|LuaSnip.Pos00
+---@param p2 LuaSnip.RawPos00|LuaSnip.Pos00
+---@return boolean
+--
+-- FIXME(@bew): I don't know how to enforce the use of the same type of pos..
```
L3MON4D3, line 213: I think it should be doable via [`@overload`](https://luals.github.io/wiki/annotations/#overload), but I'm not getting the correct results with this simple example

```lua
---@param a number
---@param b number
---@return boolean
---@overload fun(table,table) : boolean
local function test(a, b) return a == b end

test(1,{})
```

lua/luasnip/util/util.lua
```diff
@@ -305,6 +317,7 @@ local function get_snippet_filetypes()
 	return deduplicate(redirect_filetypes(fts))
 end
 
+-- FIXME(@L3MON4D3): these 2 functions are not used anywhere?
```
L3MON4D3: Very correct, feel free to junk them :+1:

lua/luasnip/util/util.lua
```diff
@@ -413,19 +436,37 @@ local function default_tbl_get(default, t, ...)
...
+		row_offset == 0 and pos[2] - base_pos[2] or pos[2],
+	}
 end
 
--- compute offset of `pos` into multiline string starting at `base_pos`.
--- This is different from pos_sub because here the column-offset starts at zero
--- when `pos` is on a line different from `base_pos`.
--- Assumption: `pos` occurs after `base_pos`.
+-- FIXME(@L3MON4D3): Is this documentation correct? It looks the same as
+-- `pos_offset` above 🤔
```
L3MON4D3: Oh, no it is incorrect, good catch :+1:

`pos_from_offset` computes the end-position of text with `offset[1]` lines and `offset[2]` columns in the last line inserted at `base_pos`. So, something of an inverse to `pos_offset`.

lua/luasnip/extras/_treesitter.lua
L3MON4D3: Good to know, TY! :)

lua/luasnip/util/mark.lua
L3MON4D3: Perfect :)

lua/luasnip/util/util.lua
L3MON4D3: Ha, I wasn't sure either and double-checked it xD
It's indeed correct, `nvim_win_get_cursor` on both of `ÿÿ` gives results more than one byte apart :+1:

lua/luasnip/util/mark.lua
L3MON4D3: Wonderful :)

lua/luasnip/util/util.lua
bew: junked in ab31a25

lua/luasnip/util/util.lua
bew: could you make a full suggestion to replace the doc of that function?
(I don't have the brain power to think much to rewrite that doc right now 👀)

L3MON4D3: Sure :)
```lua
--- Given an offset computed via `pos_offset`, compute its position when
--- applied to `base_pos`.
```
We use `pos_offset` to store the cursor-position relative to the node it is currently in, and then `pos_from_offset` to recover the cursor at a new node-position `base_pos` (`store_cursor_node_relative`, `restore_cursor_pos_relative`)

bew: Should be fixed in f079e72

PR #1445, opened 2026-06-21
fix: replaced deprecated vim.diagnostic.disable()

EliTyson:
# Replace Deprecated vim.diagnostic.disable()

vim.diagnostic.disable() was deprecated in Neovim 0.10
<https://neovim.io/doc/user/deprecated/#vim.diagnostic.disable()>

This causes an error when using the "Snippet List" feature: require("luasnip.extras.snippet_list").open().

Replacing vim.diagnostic.disable() with vim.diagnostic.enable(false) resolves the problem.


EliTyson:
Is a version checking necessary here?
```lua		
if vim.fn.has("nvim-0.10") == 1 then
	vim.diagnostic.enable(false, { bufnr = buf })
else
	vim.diagnostic.disable(buf)
end
```

